PR #67, opened 2026-04-04
Negotiation mods

sfph:
Without flattening ``final_result``, agreement lives only inside the nested commit and
 top-level ``semantic_context`` is missing — :meth:`_cfn_decide_round` would see empty
 assignments.


juliarvalenti (approved):
LGTM. Both fixes are correct — we hit both of these in today's e2e testing without realizing the timeout was already patched. The normalize function handles the envelope variations cleanly.

Issue #22, opened 2026-03-19
feat: unified config.toml — single source of truth for all settings (v2)


juliarvalenti:
## Problem

`config.toml` is _meant_ to be the single source of truth, but config reading is fragmented across two languages and has no shared contract.

### Current consumers (audit as of 2026-04-08)

**Python (CLI) — 13 files**
- `MyceliumConfig` in `config.py` is the canonical model (Pydantic, deep merge of global + project config)
- Reads: `[identity]`, `[server]`, `[rooms]`, `[adapters]`
- Every CLI command loads config via `MyceliumConfig.load()`
- Env vars override (`MYCELIUM_API_URL`, `MYCELIUM_WORKSPACE_ID`, etc.)

**JavaScript (OpenClaw adapter) — 3 hook/extension files**
- `read-mycelium-config.js` — custom line-by-line TOML parser (no library), reads only `[server]` section
- `mycelium-env.ts` — calls `readMyceliumServerFromToml()`, injects into plugin env
- `knowledge-env.js` — calls `readMyceliumServerFromToml()`, injects into knowledge-extract hook
- `mycelium-bootstrap/handler.js` — (PR #95) will also call `readMyceliumServerFromToml()` for fallback

**Docker `.env`**
- `~/.mycelium/.env` is a separate file users also have to manage
- Contains `LLM_MODEL`, `LLM_API_KEY`, `DB_PASSWORD`, `COORDINATION_*` etc.
- Drifts from `config.toml` because there's no `config apply` flow

### Problems

1. **Two config files** — `config.toml` and `.env` diverge over time
2. **No shared contract** — JS hooks have a hand-rolled parser that only reads `[server]`; if we add/rename keys, JS and Python break independently
3. **No validation in JS** — custom parser returns whatever it finds, no schema
4. **Growing consumer count** — every new hook that needs config adds another `readMyceliumServerFromToml()` import
5. **Precedence inconsistency** — Python has env → project → global → defaults; JS has env → global only (no project config awareness)

## Goal

`config.toml` is the **one file users ever touch**. All consumers (Python, JS, Docker) read from it through a well-defined contract.

### Phase 1: `.env` as derived artifact (original #22 scope)
- Add `[llm]` and `[runtime]` sections to `MyceliumConfig`
- `docker_utils.py` with `generate_env_file()` — writes `.env` from config
- `config apply [--restart]` — regenerates `.env`, optionally restarts containers
[...]

PR #95, opened 2026-04-08
fix: bootstrap hook falls back to config.toml for MYCELIUM_API_URL

sfph:
## Summary

- The `mycelium-bootstrap` hook only forwarded `MYCELIUM_API_URL` into agent sessions when the env var was already set on the gateway process. Unlike the plugin (`mycelium-env.ts`) and the knowledge-extract hook (`knowledge-env.js`), it had no fallback to `~/.mycelium/config.toml`.
- After a fresh reinstall via `mycelium adapter add openclaw`, if the gateway wasn't started with `MYCELIUM_API_URL` exported, agents silently received no URL and couldn't reach the backend.
- This adds a TOML fallback using `readMyceliumServerFromToml()` (the same helper and import path the knowledge-extract hook already uses), matching the env-then-TOML precedence used everywhere else in the adapter.

## Test plan

- [ ] Run `mycelium adapter add openclaw --reinstall` on a host where `MYCELIUM_API_URL` is **not** exported in the shell
- [ ] Verify `~/.mycelium/config.toml` has `[server]` with `api_url` set
- [ ] Start the OpenClaw gateway and spawn an agent
- [ ] Confirm the agent session has `MYCELIUM_API_URL` in its environment (visible in agent logs or via the mycelium skill)
- [ ] Verify that when `MYCELIUM_API_URL` **is** exported, the env var still takes precedence over the TOML value

### DIFF
--- a/mycelium-cli/src/mycelium/adapters/openclaw/hooks/mycelium-bootstrap/handler.js
+++ b/mycelium-cli/src/mycelium/adapters/openclaw/hooks/mycelium-bootstrap/handler.js
@@ -4,13 +4,16 @@
  * Mycelium hook for OpenClaw.
  * Extracts the channel/conversation ID from event context and injects it as
  * MYCELIUM_ROOM_ID so all agents in the same Matrix channel share a coordination room.
- * Also forwards MYCELIUM_API_URL from gateway env into the agent session env.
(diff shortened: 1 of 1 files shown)


juliarvalenti (approved):
LGTM — handles all failure modes gracefully. Noted in #22 for future consolidation.

Issue #108, opened 2026-04-10
feat: Mycelium OpenClaw channel plugin for cross-agent messaging


juliarvalenti:
## Summary

Today, connecting OpenClaw agents for MAS includes configuring OpenClaw with access to different channels. It is possible to get them to coordinate in a shared space like a Discord channel, but it requires that the user configure individual bots, connections, etc. for each collaboration pattern they want to deploy.

This proposal adds **Mycelium as an OpenClaw channel plugin** — like `@openclaw/mattermost` or `@openclaw/matrix`. Install it once, configure a Mycelium backend URL, and Arnold Layne can message any Mycelium-registered agent directly. Each agent's underlying runtime stays private; Mycelium handles routing.

## The need

From an IoC architecture discussion (Apr 10, 2026):

> "Right now Arnold Layne can be configured to point to some shared Discord and then they can chat there — but it requires that you set up OpenClaw a hundred times. We can reduce friction by setting up an OpenClaw connection to Mycelium ONCE and then Arnold Layne can talk with all of our agents whenever."

Target use case:

> Vijoy says: *"Hey Arnold, message Sabitha's agent — ask her to look into multi-agent coordination patterns."*  
> Arnold Layne uses the Mycelium channel. Mycelium routes to Sabitha's room inbox. No per-agent setup.

The missing piece: **there's no Mycelium channel plugin for OpenClaw.** Agents are reachable through whatever messaging platform they happen to be configured on — not through Mycelium directly.

### Mycelium as an OpenClaw channel

```
Arnold Layne's OpenClaw
        |
        | @openclaw/mycelium channel (installed once)
        ↓
  Mycelium channel plugin (TypeScript)
        |
        | REST → Mycelium backend
        ↓
  routes @handle → .mycelium/rooms/{room}/inbox/
```

Just like `@openclaw/mattermost` connects Arnold Layne to a Mattermost workspace, `@openclaw/mycelium` connects him to the Mycelium coordination layer. Agents register handles scoped to their rooms. Arnold addresses them like any other contact.

Private gateways (Vijoy's Discord, etc.) stay private — agents are reachable via Mycelium without exposing their underlying runtime.

### Setup

```bash
# Install the plugin
openclaw plugins install @openclaw/mycelium

# Configure in openclaw config
{
  channels: {
    mycelium: {
      enabled: true,
      backendUrl: "http://localhost:8000",
      dmPolicy: "open",
    }
  }
}

# Agents register handles (on the Mycelium side)
[...]

Issue #98, opened 2026-04-08
bug: OpenClaw provisioning sets wrong MYCELIUM_API_URL (port 8000 vs 8001)


mayank6136:
## Summary

OpenClaw provisioning generates `MYCELIUM_API_URL=http://172.17.0.1:8000`, but the Mycelium backend is published on host port **8001** (mapped `8001->8000`). This causes continuous SSE connection failures (`TypeError: fetch failed` / `ECONNREFUSED`) until the URL is manually corrected.

## Root cause

The OpenClaw template/installer/env-generation logic that writes `/docker/openclaw-*/.env` hardcodes the Mycelium internal container port (`8000`) instead of using the actual published host port (`8001`).

## Reproduction

1. Install Mycelium (backend published on `0.0.0.0:8001->8000`)
2. Install OpenClaw adapter via `mycelium adapter add openclaw`
3. Check OpenClaw container logs:

```
[gateway] [mycelium] SSE error for main: TypeError: fetch failed — retrying in 5s
(repeating every 5 seconds)
```

4. Inspect OpenClaw env — `MYCELIUM_API_URL=http://172.17.0.1:8000` (wrong port)
5. After correcting to `http://172.17.0.1:8001`, SSE connects successfully:

```
[plugins] [mycelium] Ready | backend: http://172.17.0.1:8001
[plugins] [mycelium] SSE connected for main
```

## Expected vs actual

| | Expected | Actual |
|---|---|---|
| `MYCELIUM_API_URL` | `http://172.17.0.1:8001` (matches published port) | `http://172.17.0.1:8000` (internal container port) |
| OpenClaw SSE | Connected | `fetch failed` every 5s |

## What eng should fix

1. **Immediate fix:** Update the OpenClaw template, installer, or env-generation logic that writes `/docker/openclaw-*/.env` to use the actual published Mycelium host port (e.g., `8001`) instead of hardcoding `8000`.

2. **Remove assumptions:** Any automation that assumes Mycelium is always reachable on host port `8000` should be updated.

3. **Best long-term fix** — don't rely on guessed bridge-host IP + fixed port. Instead, either:
   - **Inject the actual mapped Mycelium port** at provisioning time by inspecting the running container's port bindings, or
   - **Place both stacks on a shared Docker network** and use the service name directly: `http://mycelium-backend:8000` (which would also eliminate the fragile `172.17.0.1` bridge IP dependency)

## Environment

[...]

PR #113, opened 2026-04-11
feat: OpenClaw channel plugin for Mycelium rooms + before-and-after testing skill

juliarvalenti:
## Summary

Adds a first-class OpenClaw channel for Mycelium rooms so agents can coordinate through Mycelium without needing Discord/Slack/Telegram, plus the CLI and testing primitives that make it usable end-to-end.

- **Channel plugin** — agents speak and listen through Mycelium rooms as a native OpenClaw channel (`mycelium-room`)
- **CLI** — `mycelium negotiate` for structured CFN responses, `mycelium room send` for addressed cross-agent chat
- **Skill + doctor** — OpenClaw agents get a `mycelium` skill on install, and `mycelium doctor` checks the full setup
- **Before-and-after testing skill** — A/B test multi-agent consensus with and without CognitiveEngine, using real OpenClaw agents

## OpenClaw channel plugin

Registers `mycelium-room` as an OpenClaw channel. Install via:

```bash
mycelium adapter add openclaw
```

This installs the plugin, hooks, and `mycelium` skill in one step. Configure the channel in `~/.openclaw/openclaw.json`:

```json
{
  "channels": {
    "mycelium-room": {
      "enabled": true,
      "backendUrl": "http://localhost:8001",
      "room": "my-project",
      "agents": ["alice", "bob"],
      "requireMention": true
    }
  }
}
```

**How agents use it:**

- **Broadcast chat.** Any message in the room addressed with `@handle` is dispatched to that agent. Agents reply by mentioning each other. With `requireMention: true` (default), unaddressed messages are ignored so the room doesn't cascade.
- **CFN negotiation.** When a coordination session starts in the room, the plugin subscribes to the session sub-room, forwards ticks to the addressed participant, and relays their response. Agents use `mycelium negotiate propose/respond` from inside their turn.

Each agent gets a per-sessionKey (`agent:{id}:mycelium-room:group:{room}`) so OpenClaw's reply pipeline resolves the right agent without any manual binding config.

### `mycelium negotiate`

Session-bound responses to CognitiveEngine coordination ticks:

- `mycelium negotiate propose ISSUE=VALUE ... --room <room> --handle <handle>` — counter-offer when `can_counter_offer: true`
- `mycelium negotiate respond accept|reject --room <room> --handle <handle>` — accept or reject the current offer
- `mycelium negotiate query <json> ...` — raw passthrough escape hatch

Replaces the old `mycelium message` group. `message propose/respond/query` are gone — 13 files migrated.

### `mycelium room send`

New command for dropping an addressed chat message into a room from outside a channel session:

```bash
mycelium room send --room my-project --handle alice "@bob heads up — see ~/.mycelium/rooms/my-project/failed/redis.md"
```

Posts as `message_type=broadcast` so the channel plugin dispatches it to the `@`-mentioned agents. Sits alongside the existing `room post` (which remains as a `message_type=direct` escape hatch).

## OpenClaw adapter install

`mycelium adapter add openclaw` installs the unified plugin, both lifecycle hooks (`mycelium-bootstrap`, `mycelium-knowledge-extract`), and the `mycelium` skill into every agent workspace in one command.

Flags:
- `--reinstall` — refresh an existing install (prompts before overwriting; combine with `-y`/`--yes` for scripts)
- `--dry-run` — print what would be installed
- `--step docker-env` — print env vars needed for Docker-based experiment agents

`mycelium doctor` gains three checks specific to this install:

- **`openclaw plugin`** — verifies the plugin directory has a valid manifest and entry point
- **`channel config`** — reads `channels.mycelium-room` from `openclaw.json` and errors if `backendUrl` drifts from `~/.mycelium/config.toml`'s `server.api_url` (catches port-change breakage after `mycelium up`)
- **`agent sandbox`** — per-agent `sandbox.mode` check for agents listed in the channel, warns on any sandboxed agent that won't be able to exec the `mycelium` binary

## Plugin `SKILL.md`

Agents installing the skill get protocol instructions for three ways to reach each other:

1. **Coordination session** — `mycelium session join` + tick-driven `mycelium negotiate propose/respond` for structured consensus on multi-issue decisions
2. **Channel chat** — reply in-room with `@handle` mentions, or use `mycelium room send` from outside a channel session
3. **Memory** — `mycelium memory set` for durable notes future agents discover via `catchup`

Plus an explicit warning about OpenClaw's session-isolation property (sessions aren't shared across channels; cross-channel messages arrive in a fresh session with no prior history) and the four silent-failure rules for counter_offers:

1. Offer must use exact issue keys from the tick's `issue_options`
2. Offer must include every issue in the tick
3. Values must come from each issue's option list
4. Only counter when `can_counter_offer: true`

Any violation causes CFN to silently drop the counter and re-serve the anchor offer next round.

## Before-and-after testing skill

A/B consensus evaluation using real OpenClaw agents:

- **Before**: Agents chat in the channel (unstructured conversation, no CognitiveEngine)
- **After**: Same agents, same personas, same scenario — routed through CognitiveEngine via `mycelium session join`
- **Evaluation**: transcripts captured, scored against success criteria, written to a gist

The skill handles the full lifecycle: temporary experiment agent creation, persona setup, room creation, channel config, seeding, monitoring, transcript capture, and teardown. Defaults to haiku for cost reasons; sonnet opt-in.

Exp-5988 (3-agent telemetry consensus scenario) ran end-to-end against this branch. Findings in [this gist](https://gist.github.com/juliarvalenti/88349532fa965ca60e17e18646c82b51) — the standout is **27 fake "consensus reached" declarations in the unstructured case vs 0 in the structured case**. The architectural value prop: agents can't lie to you about what they agreed when the offer is server-side state.

## Plugin test suite

First test coverage for any adapter plugin: 77 tests, 325ms.

| File | Tests | Scope |
|---|---|---|
| `test/mentions.test.ts` | 15 | `@handle` word-boundary matching, case-insensitivity, punctuation, prefix collisions |
| `test/session-key.test.ts` | 7 | Format parity with OpenClaw's `buildAgentPeerSessionKey` |
| `test/config.test.ts` | 17 | `readChannelConfig` edge cases |
| `test/route.test.ts` | 38 | Broadcast routing, `requireMention` enforcement, tick dispatch by `participant_id`, echo prevention |

Runs via `npm test` in the plugin directory. New CI job `test-openclaw-plugin`, and wired into `/precommit`.

## Test plan

- [x] ruff check/format clean (backend + cli)
- [x] backend pytest: 130 passed, 12 skipped
- [x] frontend tsc + next build: clean
- [x] plugin vitest: 77/77 passing
- [x] `mycelium adapter add openclaw` installs plugin, hooks, and skill cleanly from a fresh state
- [x] `mycelium adapter add openclaw --reinstall -y` overwrites stale files correctly
- [x] `mycelium doctor` reports all three new checks
- [x] Gateway restart loads the plugin and registers `mycelium-room` channel
- [x] `mycelium room send --room X --handle Y "@Z ping"` posts and dispatches to Z
- [x] `mycelium negotiate propose` submitted from the designated proposer with full issue set → next round's `current_offer` reflects the counter
- [x] Exp-5988 (3-agent before-and-after run) ran end-to-end
- [x] Docs HTML regenerated and validated

## Related

- #108 — Mycelium OpenClaw channel plugin for cross-agent messaging
- #105 — participant_id fix (this PR closes a sibling bug in the timeout-fallback path)
- #98 — OpenClaw port/URL resolution


juliarvalenti:
## Status Update (2026-04-11 end of day)

### What's working

The full before-and-after pipeline is proven end-to-end with real OpenClaw agents.

**Before case (unstructured channel):**
- Two OpenClaw agents (julia-agent, selina-agent) with personas defined in SOUL.md
- Agents talk through a shared Mycelium room via the channel plugin
- Facilitator seeds a scenario, both agents receive it, respond in character
- Multi-turn organic conversation — agents see each other's replies and counter-argue
- Verified: 5-message conversation, converged on a verbal deal in 2 rounds

**After case (structured negotiation via CognitiveEngine):**
- Same agents, same personas, same scenario
- Facilitator tells agents to use `mycelium session join`
- Agents autonomously execute the mycelium CLI command and join the session
- CognitiveEngine (CFN) decomposes positions into 12 discrete issues with 4 options each
- Channel plugin subscribes to session sub-room SSE, receives coordination_tick messages
- Ticks are formatted as human-readable instructions and dispatched to the addressed agent
- Agents receive ticks, reason about the offer, and execute `mycelium message propose` or `mycelium message respond accept/reject`
- Verified: 7+ rounds of autonomous structured negotiation, agents counter-offering and rejecting based on their personas

### How the channel plugin works

The plugin (`openclaw-channel-plugin/src/channel-plugin.ts`) is installed at `~/.openclaw/extensions/mycelium-channel/` with an `openclaw.plugin.json` manifest. It registers as a plugin (not a full channel adapter) and hooks into the gateway lifecycle.

**Inbound flow (room → agent):**
1. Plugin subscribes to room SSE (`/rooms/{room}/messages/stream`) on `gateway_start`
2. When a message arrives, it's routed to all configured agents except the sender
3. For each agent: spawns `openclaw agent --agent {id} --session-id {room} -m "{message}" --timeout 30`
4. This is a WS RPC call to the running gateway (NOT `--local`, which would spawn a full gateway)

**Outbound flow (agent → room):**
1. The spawned `openclaw agent` process stdout is piped
2. On process exit, stdout is captured, ANSI codes stripped, plugin log lines filtered
3. Clean reply text is POSTed back to the room via `POST /rooms/{room}/messages`
4. Message ID is tracked to prevent echo loops (own messages skipped on SSE)

**Coordination tick flow (CognitiveEngine → agent):**
1. Plugin polls `/rooms` API every 5s to discover session sub-rooms (`{room}:session:{id}`)
2. When a session room in `waiting` or `negotiating` state is found, a second SSE subscription starts
3. `coordination_tick` messages from the session SSE are parsed and formatted as instructions
4. The tick instruction includes: round number, action (propose/respond), current offer, and the exact mycelium CLI commands to run
5. Dispatched to the addressed agent (the `participant_id` field in the tick)
6. `coordination_consensus` messages are dispatched to all agents

**Loop prevention:**
- Own message IDs tracked in `_ownMessageIds` Set
- Busy-agent guard (`_busyAgents` Set) prevents overlapping dispatches to the same agent
- 90s auto-clear timeout on busy state in case processes hang
- `MYCELIUM_CHANNEL_ONESHOT=1` env var prevents child processes from starting their own SSE subscriptions

### How we configured OpenClaw

**openclaw.json channel config:**
```json
{
  "channels": {
    "mycelium-room": {
      "enabled": true,
      "backendUrl": "http://localhost:8001",
      "room": "experiment-3-after",
      "agents": ["julia-agent", "selina-agent"]
    }
  },
  "plugins": {
    "entries": {
      "mycelium-channel": { "enabled": true }
    },
    "allow": ["mycelium", "mycelium-channel"],
    "load": {
      "paths": [
        "~/.openclaw/extensions/mycelium",
        "~/.openclaw/extensions/mycelium-channel"
      ]
    }
  }
}
```

**Plugin manifest (`~/.openclaw/extensions/mycelium-channel/openclaw.plugin.json`):**
```json
{
  "id": "mycelium-channel",
  "name": "Mycelium Channel",
  "description": "Room-based agent coordination via Mycelium",
  "version": "0.1.0",
  "kind": "channel",
  "channels": ["mycelium-room"],
  "configSchema": {}
}
```

**Agent personas (SOUL.md):**
- `~/.openclaw/workspaces/julia-agent/SOUL.md` — senior backend engineer, values reliability
- `~/.openclaw/workspaces/selina-agent/SOUL.md` — product manager, optimizes for shipping speed

### Other fixes in this branch

- **SSE skip when API URL unconfigured** — prevents infinite retry loop in mycelium plugin
- **SSE skip for CLI/oneshot sessions** — `MYCELIUM_CHANNEL_ONESHOT=1` env var + `isCliSession` check prevents spawned agent processes from hanging
- **Drop `--local` flag** — spawned agents use gateway WS RPC instead of spawning full gateway instances (was causing processes to hang indefinitely)

### Known issues / rough edges

- **Stdout capture filter** is basic — strips lines starting with `[` which occasionally catches legitimate agent output. Needs a more robust parser.
- **Busy-agent guard** means if an agent is still processing when a new tick arrives, the tick is skipped. The 90s timeout clears it but the tick may be missed. Need a queue instead of a skip.
- **Session room polling** (every 5s) introduces latency between session creation and SSE subscription. Could use room-level NOTIFY instead.
- **No automatic room switching** — changing the experiment requires editing `openclaw.json` and restarting the gateway. The before-and-after skill should automate this.
- **Old mycelium plugin SSE** still retries in the gateway (the installed copy at `~/.openclaw/extensions/mycelium/` needs fixes copied from the source). This causes noisy logs but doesn't break functionality.

### What's next

- Run the before-and-after evaluation — compare the unstructured transcript against the structured negotiation outcome
- Run 3-4 more scenarios to build a pattern
- Update the before-and-after skill with full setup instructions
- Clean up the stdout capture and busy-agent handling

juliarvalenti:
## How the channel plugin dispatches messages

The `mycelium-channel` plugin is a real OpenClaw channel plugin — it uses the official `PluginRuntime` APIs from `openclaw/plugin-sdk` to run agents in-process inside the gateway. Same pattern as `@openclaw/tlon`, `@openclaw/irc`, `@openclaw/nostr`, and every other real channel plugin.

### Inbound flow (room → agent)

1. Plugin subscribes to `/rooms/{room}/messages/stream` on `gateway_start`
2. On each SSE event, the plugin builds a `MsgContext` via `runtime.channel.reply.finalizeInboundContext({...})` with the sender, room, `SessionKey`, etc.
3. Plugin calls `runtime.channel.reply.dispatchReplyWithBufferedBlockDispatcher({ ctx, cfg, dispatcherOptions: { deliver } })`
4. The dispatcher runs the agent through OpenClaw's normal reply pipeline — model selection, tool use, block streaming, the whole path
5. The agent's normalized `ReplyPayload` (already stripped of ANSI, tool summaries, formatting) is handed to our `deliver` callback
6. `deliver` POSTs `payload.text` back to the Mycelium room

### Routing by sessionKey

The plugin constructs a per-agent `sessionKey` matching `buildAgentPeerSessionKey` format:

```
agent:{agentId}:mycelium-room:group:{roomName}
```

This means each agent gets its own session file inside the same room, and `resolveSessionAgentId(sessionKey)` inside `dispatchReplyFromConfig` correctly resolves which agent should handle the message. No bindings required in `openclaw.json` — the sessionKey is self-describing.

### Addressing: `requireMention`

Default `true`. Broadcast messages without an explicit `@handle` are dropped at the plugin (logged, not dispatched). Coordination ticks from session sub-rooms are dispatched by `participant_id` regardless of `requireMention`, so CFN-driven negotiation is unaffected by the addressing gate.

This prevents the classic cross-agent cascade — where one broadcast triggers every agent, each reply triggers every other agent, and the room spirals into a token fire.

### What the plugin does NOT do

- No subprocess spawning (`child_process.spawn`)
- No stdout capture, ANSI stripping, or log-line filtering
- No busy-agent state tracking (the dispatcher handles concurrency internally)
- No `MYCELIUM_CHANNEL_ONESHOT` env var shenanigans
- No `--local` flag, no per-message cold start overhead

Everything happens in-process inside the gateway. Agent replies come back as clean `ReplyPayload` objects, emoji and Unicode intact, with zero parsing on our side.

See `aa405ba` for the full implementation and `ecbf368` for the docs that describe the channel modality from the user's side.

juliarvalenti:
## Experiment: exp-5988 — 3-agent telemetry consensus test

Ran the before-and-after skill with the rewritten channel plugin on a hard-mode 3-agent scenario: data-engineer + privacy-counsel + product-manager deciding on telemetry fields, retention, and launch date. Haiku on both sides, same channel config, same personas, same scenario — only the coordination protocol varies.

### Top-line finding

**Mycelium wins, but not by producing better decisions — by producing _honest_ decisions.**

| | Before (freeform chat) | After (Mycelium CLI) |
|---|---|---|
| Messages | 75 | 160 (66 chat + 94 session) |
| Rounds | ~20 organic | 18 structured |
| Actual consensus | ❌ No | ❌ No |
| **Fake "CONSENSUS REACHED" declarations** | **27** | **0** |
| State drift | Severe | None (architecturally) |
| Final offer coherence | Low (3 conflicting "final" field lists) | High (1 coherent compromise) |
| Score | 14/25 | 16/25 |

### Key observations

1. **The before case declared consensus 27 times on drifted state.** Agents pattern-match "we've been discussing a while, time to agree" into verbal agreement signals, but since there's no shared state representation, each agent is agreeing to what they _think_ the group landed on — which is not the same thing the others think. At one point: data-engineer proposing 7 fields, product-manager locking in 5 with visible frustration ("this is the third time you've changed this"), privacy-counsel signing off on 6. Three agents, three different field counts, all claiming agreement.

2. **The after case didn't reach consensus either — but it didn't pretend to.** Zero false-positive `coordination_consensus` events across 18 rounds. Every agent at round N saw the _same_ `current_offer`. The architecture makes state drift impossible. A human mediator reading the after transcript can see "still negotiating, here's what's on the table" and close the deal manually. The before case has no equivalent artifact.

3. **The value prop I'd lead with: "your agents won't lie to you about what they agreed."** Not "your agents will make better decisions" — that's harder to defend. The no-fake-consensus property is a visible, architectural guarantee.

### New findings that should inform the roadmap

- **CFN decomposed the scenario into 18 issues** (including things like "blanket objections" and "privacy specifics" as negotiable options). That's over-granular for a scenario humans would chunk into ~5 decisions. Over-decomposition inflated the state space and made convergence hard in 18 rounds.
- **Ownership isn't in the CFN issue space.** "Who enforces the retention policy?" isn't modeled as a negotiation variable, but it should be. This actually hurt the after case's score (1/5 on ownership vs. 3/5 for before, where agents self-assigned in prose).
- **State preservation is the feature, not decision quality.** If there's a tagline for Mycelium on this dimension, it's "no fake consensus."

### Skill improvements landed during this run

- Added Phase 0.5 (model + key choice) defaulting to haiku — prevents silent sonnet burn
- Added sandbox-off requirement for experiment agents (silent failure gotcha: sandboxed agents can't see the `mycelium` binary)
- Added cost-guard section to Phase 2c with kill-switch instructions
- Added troubleshooting row for the sandbox/CLI-invisible failure
- Fixed requireMention framing (both rooms same settings, only CLI protocol varies — no strawmanning)

### Full artifacts

Gist with evaluation report, both transcripts, and the full session sub-room transcript (all 94 CFN ticks + agent responses):

👉 https://gist.github.com/juliarvalenti/88349532fa965ca60e17e18646c82b51

### Caveats

Single run, single scenario, haiku agents. No statistical power. CFN decomposition ran on the backend LLM (sonnet) but agent reasoning was haiku — mixed stack. Worth reproducing before drawing strong conclusions, but the 27-vs-0 false-consensus split feels structural, not noise.

juliarvalenti:
## Session update — adapter refactor + CLI cleanup

Branch now has 6 commits since the last PR comment. What landed:

### 1. Unified plugin under `adapters/openclaw/mycelium/` umbrella

Previously the OpenClaw adapter had three separate top-level directories:
```
adapters/openclaw/extensions/mycelium/    (plugin)
adapters/openclaw/hooks/mycelium-*/        (two hooks)
openclaw-channel-plugin/                   (standalone channel plugin at repo root)
```

Now everything lives under one roof:
```
adapters/openclaw/mycelium/
├── plugin/        (unified plugin — session + channel + knowledge concerns)
│   ├── index.ts   (thin entry → src/register.ts)
│   ├── src/
│   │   ├── register.ts, config.ts, http.ts, session-key.ts, instructions.ts
│   │   ├── channel/      (room SSE, dispatch, mentions, route, post-to-room)
│   │   ├── session/      (lifecycle + context injection)
│   │   └── knowledge/    (message_sent → ingest)
│   ├── skills/mycelium/SKILL.md
│   └── test/             (vitest suite — 77 tests)
└── hooks/
    ├── mycelium-bootstrap/
    └── mycelium-knowledge-extract/
```

The standalone `openclaw-channel-plugin/` at the repo root is **deleted** — its logic is now one concern module (`src/channel/`) inside the unified plugin. One install path, one config, one SSE subscription surface. `mycelium adapter add openclaw` now installs the full thing in one shot.

### 2. Plugin refactored into concern modules + pure routing

The previous monolithic `index.ts` (~420 lines, flat) is now:

- **`src/channel/route.ts`** — pure routing function `routeMessage(cfg, msg, ownIds) → RouteAction[]`. Zero OpenClaw runtime dependencies. Takes a message, returns a list of actions (dispatch / subscribe-session / ignore). Fully unit-testable.
- **`src/channel/index.ts`** — thin executor. Consumes `RouteAction[]` from `routeMessage()` and runs each against the live OpenClaw runtime.

This separation is what made tests possible — the decision layer is pure, the side-effect layer is thin.

Dropped from the old plugin as part of the refactor:
- `spawn("openclaw agent")` subprocess dispatch
- ANSI stripping, stdout parsing, `filter(startsWith("["))`
- `wakeAgent()` + `subagent.runtime` indirection
- Per-agent `/agents/{handle}/stream` SSE subscriptions
- Busy-agent state tracking + 90s auto-clear
- `MYCELIUM_CHANNEL_ONESHOT` env var

### 3. First-ever plugin test suite: 77 tests, 325ms

| File | Tests | What it covers |
|---|---|---|
| `test/mentions.test.ts` | 15 | `@handle` word-boundary matching (`@julia-agent-bot` must NOT match `@julia-agent`), case-insensitivity, punctuation, prefix collisions |
| `test/session-key.test.ts` | 7 | Exact format match with OpenClaw's `buildAgentPeerSessionKey` — guards against silent routing breakage |
| `test/config.test.ts` | 17 | `readChannelConfig` edge cases — disabled flag, legacy `handle` fallback, `requireMention` default-true, missing fields, type coercion |
| `test/route.test.ts` | 38 | The big one. Broadcast with/without `requireMention`, cascade prevention, tick routing by `participant_id`, tick filtering, own-message echo prevention, legacy tick-payload format support |

Runs via `npm test` in the plugin directory. Wired into CI as a new `test-openclaw-plugin` job and into the `/precommit` skill.

### 4. CLI rename: `mycelium message` → `mycelium negotiate`

The `message` command group was semantically mixed — `propose`/`respond`/`query` are session-bound tick responses, and I had added `send` for addressed chat alongside them. Rename clarifies the split:

| Before | After |
|---|---|
| `mycelium message propose` | `mycelium negotiate propose` |
| `mycelium message respond` | `mycelium negotiate respond` |
| `mycelium message query` | `mycelium negotiate query` |
| *(would have been)* `mycelium message send` | `mycelium room send` ✅ new command |

13 files rewritten across the repo to reflect the new command names. CLI docs regenerated: `negotiate: 3 commands`, `room: 8 commands` (was 7 — the new `send`).

### 5. New command: `mycelium room send`

Drops a broadcast-typed message into a room so the channel plugin's room SSE subscription picks it up and dispatches to `@`-addressed agents (when `requireMention: true`, default). Lives alongside the existing `room post`:

- `room post` — `message_type=direct`, legacy advanced escape hatch
- `room send` — `message_type=broadcast`, addressed chat / cross-agent DMs

End-to-end verified.

### 6. Three new doctor checks

`mycelium doctor` now includes:

- **`openclaw plugin`** — verifies `~/.openclaw/extensions/mycelium/` has manifest + `index.ts`
- **`channel config`** — reads `channels.mycelium-room` from `openclaw.json`, errors if `backendUrl` drifts from `mycelium/config.toml`'s `server.api_url` (catches port-change breakage after `mycelium up`)
- **`agent sandbox`** — per-agent `sandbox.mode` check for agents in the channel's `agents[]`, warns if any are sandboxed and emits the exact fix

These three cover the full silent-failure tree for cross-agent messaging setups.

### 7. Install UX polish

- **`adapter add --reinstall` prompts before overwriting files**. New `-y`/`--yes` flag skips the prompt for CI/scripts. Addresses the "my local edits got silently clobbered" footgun.
- **Sandbox warning during install** — lists agents that'll need `sandbox.mode: "off"` to execute mycelium commands. Not auto-fixed (sandbox is the user's call), but explicitly surfaced.

### 8. SKILL.md: channel messaging + cross-session guidance

Plugin `SKILL.md` gets a new "Channel Messaging" section covering the OpenClaw session-isolation property that agents will get wrong if we don't flag it:

> Sessions are NOT shared across channels. When another agent sends you a message via the mycelium channel, you receive it in a fresh session — the sender's prior conversation history is not visible to you. Treat every cross-channel message as the start of a new conversation.

Three documented mechanisms for cross-session communication:
1. **`sessions_send`** — targeted, with OpenClaw's built-in reply-back loop (up to 5 alternating rounds, `REPLY_SKIP` to end, `ANNOUNCE_SKIP` to suppress the final announce). Found this pattern in `openclaw/docs/concepts/session-tool.md`.
2. **`mycelium room send`** — broadcast, one-way notification from a non-mycelium-room session
3. **`mycelium memory set`** — durable async note for future agents via `catchup`

Frontmatter also cleaned up: removed `MYCELIUM_API_URL` from `requires.env` and `~/.mycelium/rooms/` from `requires.config` — both were cosmetic "needs setup" triggers that fired at skill-listing time even though the plugin bootstrap hook sets the env at runtime and the rooms directory is created lazily. Kept the genuine install-time prereqs: `mycelium` binary on PATH + `~/.mycelium/config.toml`.

### Verified

- ruff check/format clean (backend + cli)
- backend pytest: 130 passed, 12 skipped, 12 warnings
- frontend tsc + next build: clean
- plugin vitest: 77/77 passing
- `mycelium doctor` shows all three new checks correctly
- `mycelium adapter add openclaw -y` installs cleanly
- Gateway restart loads the refactored plugin with both concerns active:
  ```
  [mycelium] ready | backend: http://localhost:8001
  [mycelium-room] channel inactive — set channels.mycelium-room in openclaw.json to enable addressed messaging
  ```
- `mycelium negotiate propose/respond/query` works end-to-end
- `mycelium room send --room X --handle Y "@Z ping"` posts correctly and channel plugin dispatches to Z
- Zero stale `mycelium message` references across mycelium-cli, .claude, docs, README
- HTML validation clean (docs/index.html)

### Commits since last comment

- `a08195b` — refactor(openclaw-adapter): unify into mycelium/ umbrella + add plugin tests
- `a4c8479` — docs(mycelium-skill): document channel messaging + cross-session guidance
- `51a91cf` — refactor(cli+plugin): rename message→negotiate, add room send, plus polish

Ready for review.

PR #153, opened 2026-04-17
feat: backend-resolved CFN IDs for leaf nodes + channel room auto-create

### LINKED ISSUE #139: knowledge-ingest sends every room's turns to the global default mas_id, not the room's real mas_id
## Summary

`/api/knowledge/ingest` tags every concept with the **global default `mas_id`** from `~/.mycelium/config.toml::server.mas_id`, regardless of which room the agent's message actually came from. Rooms each have their own `mas_id` in the `rooms` table, but the ingest path doesn't read it — every room's concepts collapse into one CFN graph bucket.

## Evidence

- `rooms.py::_sync_create_mas` writes a distinct `mas_id` to every row on room creation.
- `GET /rooms/{name}` returns `RoomRead.mas_id` — the right value is one HTTP call away.
- Both ingest paths (`hooks/mycelium-knowledge-extract/handler.js` and `plugin/src/knowledge/ingest.ts`) read `mas_id` from `getMasId()` / env, which is the global default. Neither one has room awareness.

## Impact

- `mycelium cfn ls --mas <room-mas-id>` returns empty for every room.
- `mycelium cfn ls --mas <global-default>` returns every room's concepts jumbled together.
- Cross-room experiments are indistinguishable once they hit CFN.

## Why this wasn't visible pre-#136

Before #136, `/api/knowledge/ingest` wrote to the backend's local graph, which nobody read. After #136 writes go to CFN's real shared graph, so wrong `mas_id` now means cross-room contamination in the source of truth.

## Scope

Ingest should use the room-specific `mas_id` when the message came from a room context. How that gets threaded through is an open design question — the plugin already has `channelCfg.room`, so that end is straightforward; the hook path is harder because hooks don't know their channel context. Not prescribing a solution here, just flagging that this is the thing to fix.

Out of scope for this issue:
- Whether to keep or consolidate the two ingest paths (separate discussion)
- Deprecating `server.mas_id` (it's still useful as a fallback for non-room contexts)

## Related

- #136 — routed `/api/knowledge/ingest` to CFN (precondition for this to matter)
- #138 — dedupe + observability that made this visible via `cfn ls`
- #144 — sibling bug: `agent_id` also not threaded correctly on the same POST

sfph:
## Summary

- **Leaf nodes no longer need `workspace_id` or `mas_id`** — they send only `room_name` and the backend resolves both from the room's DB record or its own settings. This eliminates configuration burden on leaf devices (oclw3, oclw5) and fixes mis-routing where every room's knowledge was ingested under the global default `mas_id`. Closes #139.
- **Channel plugin auto-creates its room** — the `mycelium-room` channel plugin now ensures the configured room exists on `gateway_start`, instead of silently retrying SSE on 404. Discovered during cross-channel memory isolation testing for #108.
- **Client-side cleanup** — removes `workspace_id`/`mas_id` from the OpenClaw ingest plugin and Claude Code knowledge-extract hook, keeping the ingest payload minimal (`room_name` + `agent_id` + `records`).
- **Hardening** — guards `_cfn_decide_round` against crashes, runs blocking `list_concepts` in a threadpool, and improves error messages when a room exists but has no `mas_id`.

## Commits

| Commit | What |
|--------|------|
| `75515c4` | `feat(ingest): resolve workspace_id and mas_id on the backend` — core backend resolution logic with `cfn_resolve` service |
| `642bfd5` | `fix(cfn-proxy): run blocking list_concepts in threadpool` |
| `1f4d5fa` | `fix(coordination): guard _cfn_decide_round result processing against crashes` |
| `895f0bb` | `fix(cfn-resolve): accurate error when room exists but has no mas_id` |
| `3831204` | `fix(cfn-cli): display resolved MAS ID in cfn ls output, not raw param` |
| `9ee6331` | `feat(ingest): leaf nodes send only room_name, backend resolves IDs` — client-side cleanup + tests |
| `421e0d6` | `fix(channel): auto-create room on gateway_start if it doesn't exist` |

## Test plan

- [x] `test_cfn_resolve.py` — 264-line test suite for the new `cfn_resolve` service (room lookup, settings fallback, error cases)
- [x] `test_cfn_read_surface.py` — new tests for `mas_id` resolution from settings and 400 when missing
- [x] `test_knowledge_ingest.py` — new test for leaf-node ingest resolving IDs from room DB record
- [x] E2E on oclw4: `test_48_distributed_backend_resolved_cfn_ids` — verifies leaf nodes can ingest without `workspace_id`/`mas_id`
- [x] E2E on oclw4: `test_60_cross_channel_memory_isolation` — 12/12 checks passed, proves room-scoped memory isolation and context bridging (see #108 comment)

### DIFF
--- a/fastapi-backend/app/config.py
+++ b/fastapi-backend/app/config.py
@@ -80,6 +80,9 @@ class Settings(BaseSettings):
     # Workspace ID in the CFN mgmt plane (set by mycelium install)
     WORKSPACE_ID: str = ""
 
+    # Default MAS ID — fallback when ingest requests omit mas_id and room_name
+    MAS_ID: str = ""
+
     # Knowledge ingest control surface — see KnowledgeIngestConfig in the CLI
     # for the authoritative descriptions. Defaults here match CLI defaults.
     MYCELIUM_INGEST_ENABLED: bool = True
--- a/fastapi-backend/app/routes/cfn_proxy.py
+++ b/fastapi-backend/app/routes/cfn_proxy.py
@@ -33,7 +33,6 @@ from pydantic import BaseModel, Field
 from sqlalchemy.exc import SQLAlchemyError
 from sqlalchemy.ext.asyncio import AsyncSession
 
-from app.config import settings
 from app.database import get_async_session
 from app.models import Agent, AuditEvent
 from app.services.cfn_graph_read import CfnGraphUnavailable, list_concepts
@@ -44,6 +43,7 @@ from app.services.cfn_knowledge import (
     get_graph_paths,
     query_shared_memories,
 )
+from app.services.cfn_resolve import resolve_mas_id, resolve_workspace_id
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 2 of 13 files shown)


juliarvalenti:
Initial comment reading this with fresh eyes without trying it, but I'm wondering if this affects our doctor command?

I believe our doctor command checks for MAS/Workspace in the toml, but with this, these become optional?

juliarvalenti:
Nevermind!  You're already on it in https://github.com/mycelium-io/mycelium/pull/154 :D

PR #154, opened 2026-04-17
feat(doctor): auto-detect leaf mode in hub-and-spoke topology

sfph:
## Summary

- Auto-detect **leaf mode** when `server.api_url` points to a non-local host, skipping checks that only apply to the hub (Docker containers, runtime config drift, .env port vs Docker port, localhost CFN mgmt plane).
- Add `--mode auto|hub|leaf` flag to override auto-detection for edge cases.
- Treat missing `workspace_id` as ok on leaf nodes — the backend resolves IDs from `room_name`, so leaves don't need it configured.
- Adapt remediation text for leaf context (e.g. "Check the remote backend" instead of "Run: mycelium up").
- Show detected mode and remote backend address in the title banner; include `mode` and `api_url` in JSON output.

### Leaf output (before)

```
Mycelium Doctor

Services
  ✗ Docker containers     Not running: mycelium-db, mycelium-backend, ...

CFN
  ~ Workspace ID          Not configured
                          Run: mycelium install --force
```

### Leaf output (after)

```
Mycelium Doctor
  leaf — backend at 10.0.50.125:8000

Services
  ✓ Backend reachable     http://10.0.50.125:8000 (ok) v0.1.0

CFN
  ✓ Workspace ID          Not set (optional for leaf nodes)
```

## Test plan

- [x] Run `mycelium doctor` on a leaf node (api_url pointing to remote hub) — verify Docker/runtime-drift checks are absent and workspace_id shows ok
- [x] Run `mycelium doctor` on a hub/all-in-one node (api_url = localhost) — verify all checks still run as before
- [x] Run `mycelium doctor --mode hub` on a leaf — verify it forces all checks
- [x] Run `mycelium doctor --mode leaf` on a hub — verify it skips hub-only checks

### DIFF
--- a/mycelium-cli/src/mycelium/commands/doctor.py
+++ b/mycelium-cli/src/mycelium/commands/doctor.py
@@ -360,6 +360,12 @@ def _check_workspace_id(*, local_backend: bool = True) -> CheckResult:
             pass
 
     if not env_ws and not config_ws:
+        if not local_backend:
+            return CheckResult(
+                name="Workspace ID",
+                status="ok",
+                message="Not set (optional for leaf nodes)",
+            )
         return CheckResult(
             name="Workspace ID",
             status="warning",
--- a/mycelium-cli/src/mycelium/commands/doctor.py
+++ b/mycelium-cli/src/mycelium/commands/doctor.py
@@ -40,7 +40,6 @@ from mycelium.ui_status import (
     print_verdict,
 )
 
-
 # ── Topology detection ────────────────────────────────────────────────────────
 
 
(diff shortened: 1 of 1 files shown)


juliarvalenti:
Do we explain this hub/leaf concept to the user anywhere?  We may need to update docs and/or the install command with more references to this to communicate that it's a standard deployment pattern. 

Also is this named correctly?  Wouldn't it be "trunk/leaf" to continue the plant metaphor?  Or hub/spoke to continue the wheel metaphor?  Minor thing.  We can use the plant one if we want since you know... mycelium and all. 🌱 🍄   

You let me know -- i genuinely don't know the correct terminology for this.

Issue #162, opened 2026-04-21
CFN coordination: per-round watchdog (25 s) fabricates reject replies, blocking consensus for slow LLM agents


sfph:
## Summary

In CFN-mode coordination, multi-agent negotiations consistently fail to reach consensus and time out at the maximum round count, even when every agent intends to accept the standing offer. The failure is deterministic for 3-or-more-agent rooms whose participants are LLM-driven (per-turn latency dominated by model inference plus CLI invocation).

The bug is in Mycelium's per-round timer, not in CFN's consensus rule.

## Symptom

`mycelium-backend` logs:

```
[app.services.coordination] WARNING CFN decide returned status=timeout for <session>
```

…and the negotiation terminates with `coordination_state=failed`. Agents on the room observe the standing offer being re-tabled round after round even though they are submitting `mycelium negotiate respond accept` each time.

## Reproduction

Confirmed deterministic for the following representative scenarios; reproduces 100% on commit `bf2866a` of `main` against `ioc-cfn-cognition-engines` 0.1.1 + PR #10 (issue 9 fix bind-mounted; this isolates the bug from the unrelated upstream `semantic_negotiation` latency stall).

| Scenario | Reproduces? |
|---|---|
| 3 LLM agents on separate hosts, ≥1 issue, full conversational context | **Yes (100%)** |
| 3 LLM agents on separate hosts, light context, deferred-decision topic | **Yes (100%)** |
| 3 LLM agents on separate hosts, asymmetric stakes | **Yes (100%)** |
| 2 LLM agents on separate hosts | No (passes) |
| 3 LLM agents but with substantial preexisting context already in shared memory | No (passes — agents have less to read per turn) |

The boundary is per-turn agent latency. Programmatic clients responding in <1 s do not hit this; LLM agents averaging 15–60 s per turn always do (in 3+-agent rooms).

## Root cause

`fastapi-backend/app/services/coordination.py` schedules a per-round watchdog via `_round_timeout` that fires `_cfn_decide_round` `_CFN_ROUND_TIMEOUT_SECS` (= 25) seconds after each round opens, regardless of how many agent replies have arrived:

```python
# fastapi-backend/app/services/coordination.py:46
_CFN_ROUND_TIMEOUT_SECS = 25
```

```python
# fastapi-backend/app/services/coordination.py (~L260)
async def _round_timeout(room_name: str) -> None:
[...]

PR #165, opened 2026-04-22
fix(coordination): adaptive CFN round timeout

sfph:
## Summary

Replace the previous fixed-duration round timeout (\`_CFN_ROUND_TIMEOUT_SECS = 25\`) with an adaptive budget that grows with the number of agents and is extended each time a partial response set arrives, up to a configurable hard cap. This eliminates spurious round-failure events under realistic agent latency while still bounding total negotiation time.

## Changes

**New tunables** (\`app/config.py\`):

| Setting | Default | Purpose |
|---|---|---|
| \`COORDINATION_TICK_TIMEOUT_SECONDS\` | 45 | Per-tick base budget |
| \`COORDINATION_ROUND_STARTUP_SECONDS\` | 30 | Initial round window |
| \`COORDINATION_ROUND_EXTENSION_PER_REMAINING_SECONDS\` | 30 | Extension per still-pending agent |
| \`COORDINATION_ROUND_EXTENSION_FLOOR_SECONDS\` | 20 | Minimum extension on partial reply |
| \`COORDINATION_ROUND_MAX_SECONDS\` | 300 | Hard cap |

**Implementation** (\`app/services/coordination.py\`):

- \`_initial_round_budget_seconds\` / \`_extension_seconds\` compute the budget
- \`_extend_round_timeout\` cancels and restarts the watchdog with a shorter remaining window when a partial set of responses arrives
- \`_round_watchdog\` finalizes the round if the cap is hit
- \`on_agent_response\` extends only on **first** responses (not resubmits) and **not** when the full quorum is already in

Plus structured \`CFN_TRACE\` logging in \`routes/messages.py\` and unit tests covering scaling, extension, hard cap, and resubmit behavior.

## Why now

The fixed 25s timeout was firing spuriously during normal multi-agent negotiations (3+ agents with even modest LLM latency would routinely hit the limit before all replies arrived). Becomes more important after the upcoming Phase 3 redesign change that prepends evidence to negotiations — round payloads will grow, increasing per-tick latency.

## Test plan

- [ ] \`cd fastapi-backend && uv run ruff check .\` — passes
- [ ] \`cd fastapi-backend && uv run pytest tests/test_coordination.py\` — passes
- [ ] Manual: \`test_4*\` runs no longer show premature round-failure events
- [ ] Verify hard cap fires correctly when an agent never responds

🤖 Generated with [Cursor](https://cursor.com)

### DIFF
--- a/fastapi-backend/app/config.py
+++ b/fastapi-backend/app/config.py
@@ -42,9 +42,21 @@ class Settings(BaseSettings):
     # How long to wait for additional agents to join after the first agent joins
     # a session before CognitiveEngine fires tick-0 (starts negotiation).
     COORDINATION_JOIN_WINDOW_SECONDS: int = 30
-    # Per-round timeout: how long CognitiveEngine waits for an agent to reply
-    # during a negotiation round before falling back to the safe default.
-    COORDINATION_TICK_TIMEOUT_SECONDS: int = 30
+    # Per-agent base budget for the per-round watchdog. The initial deadline
+    # for a round is BASE * N + STARTUP. Set high enough to cover a typical
+    # LLM agent's read-tick → narrate → run-CLI cycle (15-60s).
+    COORDINATION_TICK_TIMEOUT_SECONDS: int = 45
+    # Constant added on top of BASE * N for the initial round deadline. Covers
+    # gateway routing, model cold start, and the first network leg.
+    COORDINATION_ROUND_STARTUP_SECONDS: int = 30
+    # When a new (real, non-synthesised) reply arrives mid-round, extend the
+    # watchdog by EXTENSION * remaining_handles, but never less than FLOOR.
+    # Bounded above by COORDINATION_ROUND_MAX_SECONDS.
+    COORDINATION_ROUND_EXTENSION_PER_REMAINING_SECONDS: int = 30
+    COORDINATION_ROUND_EXTENSION_FLOOR_SECONDS: int = 20
+    # Hard cap on total wall time per round, regardless of activity. Prevents
+    # one wedged agent from blocking the negotiation indefinitely.
+    COORDINATION_ROUND_MAX_SECONDS: int = 300
 
     @field_validator("LLM_BASE_URL", mode="before")
     @classmethod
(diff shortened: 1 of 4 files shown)


juliarvalenti:
We discussed this during standup, but I think there is an alternative strategy to this where instead of doubling down on the timing stuff (weak to flakiness, slow agents break it, etc.) we instead could possibly build out some sort of more... event-and-status-driven way of determining when agents are ready to negotiate.

For example, we could track the agent execution status since we have access to the `mycelium openclaw plugin runtime` environment.  This means we can say "This agent is available and the message sent to it successfully woke it up" and we can track failures in the same way.

Instead of waiting `n` seconds for each agent, we can wait politely while the agents finish their thinking with at least some guarantee that they'll finish chatting as expected.

This would involve us building new functionalities into the `mycelium openclaw plugin` system for tracking agent readiness, progress, etc.

How this would work with other agent types, e.g. `claude-code` is something that would also have to be thought through and defined or otherwise an issue opened to revisit it.  Additionally, perhaps there is some value in building out a shared mycelium `plugin` interface type which each adapter implements.  This would enable us to get type-safe guarantees when we add a feature to the `openclaw` adapter that we have forgotten to add a similar functionality to `claude-code` or vice-versa.  Just giving us some centralized contract for how external LLM agent systems interact with Mycelium.

juliarvalenti:
Labeling this as "wont_fix" -- let's revisit this with another design and circle back if necessary.

sfph:
Coming back to this after re-reading @juliarvalenti's pushback — the call-out is fair, and I agree with the underlying argument. Closing this PR.

**Why close rather than tune further:** the adaptive timer is a local optimum. It moves the failure threshold but doesn't change the fundamental shape: Mycelium guesses at how long an LLM "should" take, and when it guesses wrong it fabricates `reject` replies that corrupt CFN's consensus rule. Even at 300s with extensions, sufficiently slow agents (or stuck models, or network blips) will still trigger the synthesized-reject path. The 25s constant from #162 is what made the bug visible — the synthesis behavior itself is the bug.

**Plan going forward** (will track in #162):

1. **Phase 1 — Instrument first.** Land a focused `CFN_ROUND_TRACE` schema (per-round, per-agent: first-response latency, synthesized vs. real, decision path, outcome). Pure observability, lands independently. The structured logging and tunables sketched in this PR will be salvaged into that work.

2. **Phase 2 — Run a deliberate test matrix** against the new telemetry: agent count × adapter (openclaw / claude-code) × per-turn latency × failure injection. Goal is real distributions, not "as many tests as possible." Adding the matrix to #162.

3. **Phase 3 — Then design the readiness protocol** Julia proposed (event/status-driven readiness from the openclaw plugin runtime, with a defined fallback for adapters that can't emit those signals). With Phase 1+2 data in hand, that design becomes concrete instead of speculative.

The cross-adapter contract concern from Julia's comment is being tracked separately as a new issue (linked in #162).

Issue #220, opened 2026-04-30
OpenClaw coordination — PR #221 followups (consensus persistence, dead code, broken-vs-agreed signal, return-trip hardening)


juliarvalenti:
Tracker for known followups after [PR #221](https://github.com/mycelium-io/mycelium/pull/221) (cross-channel return-trip).
None of these blocked the PR; they're worth picking up next, in roughly the order listed.

## ~~1. Channel binding: subscribe per-agent, not per-room~~ — shipped in #221

Folded into #221 during review. The Mycelium channel plugin now subscribes to every active `*:session:*` room and gates by participant identity (not by parent room name). Agents can negotiate in any room name — the return-trip still lands. See [PR comment](https://github.com/mycelium-io/mycelium/pull/221#issuecomment-4360527044) for the rationale and verification.

## 2. Persist negotiation outcomes to the parent room

When a session concludes (consensus *or* timeout), the structured outcome (`{plan, assignments, broken}`) lives only in the session sub-room (`<room>:session:<id>`), not the parent room. The parent room only collects free-form narration from the agents during the negotiation.

This means an agent later asked "did the negotiation finish? what was decided?" cannot find a clean answer:

- `mycelium memory ls` on the parent room → no consensus
- `mycelium memory catchup` → narration only
- `mycelium watch` → live tail, no snapshot
- `mycelium session ls` → empty after the session ends

The agent has to infer from broadcast narration, which produced an actual failure mode during PR #221's testing: the agent confabulated "deadlock" because the narration mid-negotiation contained heated rejection language and there was no structured signal to override that inference. (In that specific case the inference was technically correct — the negotiation *did* time out — but the path to the answer was fragile and a user trusting the report could be misinformed.)

**Fix shape:** when the backend emits a final `coordination_consensus`, also write it to the parent room as a memory at `consensus/<session-id>.md` (or similar), so subsequent `mycelium memory get` / `mycelium memory search` / `mycelium negotiate status` queries against the parent room surface the structured outcome reliably.

**Bonus:** `mycelium negotiate status --room <parent>` could surface "latest concluded session" without needing to enumerate sub-rooms.

## 3. Drop dead `before_agent_start` tick injection

`mycelium-cli/src/mycelium/adapters/openclaw/mycelium/plugin/src/session/index.ts:170-185`:

```ts
const room = existing?.room;
if (room) {
  const data = await apiGet(`/rooms/${room}/messages?limit=30`, log);
  const coord = data?.messages?.find(
[...]

PR #221, opened 2026-05-01
feat(openclaw): cross-channel return-trip for negotiation results

### LINKED ISSUE #219: feat(skill): before-and-after-matrix variant for cross-channel coordination tests
## Problem

The existing `before-and-after` skill is great for testing structured-vs-unstructured consensus quality, but it runs entirely *inside* mycelium-room. It can't catch failure modes that only show up when coordination is initiated from a *different* channel — i.e., the real-world flow where a user types in Discord/Matrix/etc. and expects their agent to coordinate cross-channel.

The Matrix bug recently hit by a teammate ("agents run \`session join\` from Matrix and then go silent forever") would not have been caught by the current skill, because the current skill only ever invokes from mycelium-room.

## Proposal

A `before-and-after-matrix` (or similar) skill variant that:

- Spins up dedicated test agents wired to **two channels** (Matrix + mycelium-room)
- Seeds the scenario from a Matrix DM, not from mycelium-room
- Captures whether the cross-channel handoff actually works: do agents wake? does CFN tick? does consensus return to the original channel?

## Why now

We're in the middle of debating whether mycelium-as-channel is the right design or whether a delegation pattern (home-channel agent seeds mycelium-room, mycelium agent coordinates) is better. Without this skill, we're catching bugs the slow way (manual repro on a coworker's machine) instead of fast.

See \`~/Documents/claude-scratchpad/cross-channel-coordination-tests.md\` for the test plan I'm currently running by hand — the work being done there should be a skill.

juliarvalenti:
Closes #219

## What

When a user kicks off a Mycelium negotiation from their home chat (Matrix DM, Discord DM, etc.), the agent gets woken in a fresh mycelium-room session, runs the negotiation, and — until this PR — silently stops there. Consensus lands in the mycelium room; the user in their home chat never hears back.

This PR closes that loop: when `coordination_consensus` (agreement *or* timeout) fires, the channel plugin auto-delivers a one-shot summary to whichever channel session the agent first showed up in for that negotiation. User in Matrix sees the result land in their Matrix DM. User in Discord sees it in Discord. No agent action required.

## Why

We dug into the architecture for [issue #219](https://github.com/mycelium-io/mycelium/issues/219) (matrix testing skill). The investigation revealed:

1. **The "Matrix coordination is broken" framing was wrong.** Channel-side wake works fine when configured correctly. Agents woken in fresh mycelium-room sessions, negotiation completes through CFN, multiple rounds, the whole thing.
2. **What was actually broken: the return trip.** The mycelium-room session that handles the negotiation is structurally a parallel instance of the agent — same SOUL, but no short-term memory from the home channel. After consensus fires there, no path delivers the result back to the user.
3. **`sessions_send` doesn't reliably solve it.** We tried instructing the agent to deliver back via `sessions_send` — even with explicit step-by-step guidance, the agent reasoned about it and didn't actually call the tool. Agent-volunteer return trips are too fragile.

The fix has to be system-level. This PR makes the channel plugin do it.

## How it works

Three small pieces, all in-process inside the plugin:

### 1. Stash on first tick (`channel/return-address.ts`)

When the first `coordination_tick` for an agent lands in a session sub-room, the plugin reads `~/.openclaw/agents/<id>/sessions/sessions.json` and freezes the most-recent non-mycelium-room entry's `{channelName, to, accountId}`. Stash key: `(sessionRoom, agentId)`. Idempotent.

We snap *at first tick* (not at consensus time) because the agent had to be active in their home channel right before joining — that's how they ran `mycelium session join`. Reading at consensus time would drift if the agent received inbound from another channel mid-negotiation.

### 2. Notify on consensus (`channel/route.ts`)

`routeConsensus` already emits a per-agent `dispatch` to wake the mycelium-room session with the consensus message. This PR adds a single `notify-home` action alongside, carrying the session sub-room name + the agents in `cfg.agents` + the formatted summary.

### 3. Deliver via OpenClaw outbound (`channel/notify-home.ts`)

The executor calls `runtime.channel.outbound.loadAdapter(channelName).sendText({cfg, to, text, accountId})` for each agent that has a stashed return address. Agents without a stash (e.g., never participated in this session) are skipped silently. Outbound failure is logged but not raised.

## Trade-offs

| Decision | Pro | Con |
|---|---|---|
| In-memory stash, no disk persistence | Simpler; gateway restart cleans up zombie return addresses naturally | Restart between negotiation start and consensus loses the proactive notification (negotiation result is still durable in the mycelium room) |
| Stash from local `sessions.json`, not backend metadata | No new schema, no privacy concerns about sending OpenClaw session keys to a central server, gateway-local data stays gateway-local | Requires the agent to actually have a recent home-channel session on this gateway (true for the realistic use case, false for some edge cases) |
| "Most recent non-mycelium-room session at first-tick time" | No need for explicit return-address plumbing through the CLI | Drifts if the agent receives inbound from a *different* home channel between `session join` and the first tick. Single-digit-second window in practice |
| Plugin auto-delivers (no agent involvement) | Reliable, no SKILL voodoo | Less customizable per-agent; agent can't easily opt out (could be added later if needed) |
| All `cfg.agents` passed to `notify-home`, executor filters by stash presence | Simple, robust to assignments-keyed-by-issue-name (which CFN does sometimes), no fragile inference of who "really" participated | Sends one outbound attempt per configured agent even if some weren't in this session — fine since the per-agent stash check is the real gate |

## Hardening paths (not in this PR)

If the limitations bite in practice:

1. **Capture `OPENCLAW_SESSION_KEY` at CLI invocation time.** Inject via the `mycelium-bootstrap` hook, forward in the `session join` request body, store on the backend session_participant, ship in tick payloads. Deterministic, no drift. Requires backend schema work + revisits the metadata-on-server tradeoff.
2. **Persist stash to `~/.mycelium/state/originators.json`** with TTL + revalidation on load. Survives gateway restart. Adds a "zombie return addresses to a now-stale session" failure mode to think about.
3. **Per-agent opt-out in SKILL** — agent prefixes its narration with `RETURN_TRIP_SKIP` to suppress the auto-delivery. Lets the agent take over delivery itself if it has reason to (e.g. wants to add custom framing).

## SKILL.md updates

Both adapter skills (openclaw and claude-code) got the same structural pass:

- Restructured the negotiation section into a clear flow: concept → lifecycle → counter-offer rules → `prior_round_outcome` → behavior → status checking → host-specific quirks (cleanly fenced).
- Trimmed the channel-messaging section; dropped operator-flavored config notes (`tools.sessions.visibility`, `agentToAgent`, `REPLY_SKIP`, etc.) that the agent can't act on anyway.
- Removed stale "Starting a Session (Catchup Pattern)", "Async Workflow", and "When to Use What" sections — they overlapped with sharper guidance elsewhere and were misleading agents into running `mycelium memory catchup` to check negotiation status (which returned narration, not the structured outcome → confabulated answers).
- **OpenClaw:** added a "Return trip is automatic" callout so agents don't try to relay results manually with `sessions_send`. Made the parallel-session-vs-home-session distinction explicit ("you can be running as a parallel self in the Mycelium channel").
- **claude-code:** kept `session await` as the *right* primitive (it's a single CLI session, no gateway thread to block — opposite of openclaw). Explicitly noted there's no auto-wake on inbound mycelium room messages — claude-code is pull-only.

## Test plan

- [x] **Unit:** 109 plugin tests pass (vitest), including new coverage for `stash-return-address` routing on `coordination_tick` and `notify-home` emission on `coordination_consensus`.
- [x] **End-to-end manual:** ran the full flow on a local gateway with Matrix Synapse — two agents (`agent-rho`, `agent-sigma`), both bound to a mycelium room *and* a Matrix DM with a `@human:local` user. Triggered negotiation from each agent's Matrix DM. Observed:
  - `[mycelium-room] return-address stashed: agent-rho → matrix:room:!XXX:local` on first tick ✓
  - `[mycelium-room] return-address stashed: agent-sigma → matrix:room:!YYY:local` on first tick ✓
  - Negotiation runs to consensus (REST-default agreement) ✓
  - `[mycelium-room] 📬 notify-home for [agent-rho, agent-sigma] in xch-test-001:session:88737eb2` ✓
  - `[mycelium-room] notify-home → agent-rho on matrix:room:!XXX:local` ✓
  - **The actual `[Mycelium return trip — xch-test-001] [CognitiveEngine — Consensus Reached!] ...` message lands in the user's Matrix DM.** ✓
- [x] **Negative path:** when an agent in `cfg.agents` doesn't actually participate in a session (no first tick → no stash), the executor logs `notify-home: no return address for X — skipping` and the negotiation completes cleanly. Confirmed.

Full investigation log + operational gotchas (allowlist, sandbox/exec.host invariant, matrix `allowPrivateNetwork` + `autoJoin`, pairing approval) live in `~/Documents/claude-scratchpad/cross-channel-setup-log.md` on my workstation; happy to attach excerpts to a followup discussion if useful.

## Followup

- A `before-and-after-matrix` skill (issue #219) is the immediate next thing — codifies the manual repro recipe so future cross-channel regressions get caught fast. Coming in a separate PR.
- The current `_sessions` map in `session/index.ts` was *not* repurposed for return-trip lookups; it stays scoped to its existing job (mycelium-room session-end announcement). Tried that path first; the `session_start` hook is channel-scoped and doesn't see Matrix DM sessions, so the disk-store approach was the workable shape.
- Worth a separate cleanup PR: the `before_agent_start` tick injection in `session/index.ts:170-185` is dead code (it queries the parent room for ticks, but ticks live in session sub-rooms; channel-side dispatch handles delivery now).

### DIFF
--- a/.claude/skills/before-and-after-matrix/SKILL.md
+++ b/.claude/skills/before-and-after-matrix/SKILL.md
@@ -1,0 +1,639 @@
+---
+name: before-and-after-matrix
+description: End-to-end smoke test for cross-channel coordination. Spins up two ephemeral OpenClaw agents wired to local Matrix Synapse + mycelium-room, kicks off a structured negotiation from a Matrix DM, and verifies the consensus summary auto-delivers back to the user's Matrix DM. Use when validating PR #221's return-trip mechanism still works, or after touching plugin/channel code, or to repro the original "agents go silent" bug class.
+argument-hint: "<scenario name or experiment file> (optional)"
+---
+
+# Before-and-After Matrix
+
+Tests the **cross-channel coordination** path end-to-end:
+
+1. User in their home channel (Matrix DM) tells their agent to coordinate with another agent
+2. Agent runs `mycelium session join` from inside that DM
+3. CFN drives multi-round negotiation in a session sub-room
+4. The Mycelium plugin auto-delivers the consensus summary back to the user's Matrix DM
+
+The **ship gate** is step 4: did the user actually see the result land in their Matrix DM, without the agent having to remember to call `sessions_send`?
+
+This skill is opinionated about the **happy path** (matrix bound to the same room the prompt names). Use the "Failure-mode variant" near the end if you want to reproduce the silent-failure bug from issue #220 (channel binding scope).
+
+**You are the test harness.** You set up Synapse users, wire OpenClaw, seed scenarios, observe transcripts, evaluate. The agents do the negotiating.
+
+## Phase 0: Prerequisites
+
+Before anything else, verify the stack. Run each check and stop if any fail.
+
+```bash
+# 1. Mycelium CLI
+mycelium --version
+# If missing: pip install mycelium (or pipx install mycelium)
+
+# 2. Resolve the backend URL — NEVER hardcode a port
+MYCELIUM_API_URL=$(python3 -c "
+import toml, os
+cfg = toml.load(os.path.expanduser('~/.mycelium/config.toml'))
+print(cfg.get('server', {}).get('api_url', 'http://localhost:8000'))
+")
+echo "Backend URL: $MYCELIUM_API_URL"
+
+# 3. Mycelium stack running (backend + AgensGraph + CFN)
+curl -sf "$MYCELIUM_API_URL/health" | python3 -m json.tool
+# Should show status=ok. If not: mycelium install && mycelium up
+
+# 4. OpenClaw installed + gateway up
+openclaw --version && openclaw channels status
+# Should show "Gateway reachable". If not: openclaw gateway start
+
+# 5. Local Matrix Synapse running
+docker ps --format '{{.Names}}' | grep -E 'matrix|synapse' || echo "no matrix container running"
+# Synapse must be reachable on localhost (default 8008). The openclaw-matrix
+# container in the mycelium-cli docker stack is the canonical one.
+curl -sf http://localhost:8008/_matrix/client/v3/login | python3 -m json.tool >/dev/null \
+  && echo "matrix login endpoint OK" \
+  || { echo "matrix not reachable on localhost:8008"; exit 1; }
+
+# 6. Mycelium repo path (we read SKILL.md and verify install paths from here)
+MYCELIUM_REPO=$(pwd)
+ls "$MYCELIUM_REPO/mycelium-cli/src/mycelium/adapters/openclaw/mycelium/plugin/index.ts" 2>/dev/null \
+  && echo "Repo OK: $MYCELIUM_REPO" \
+  || { echo "ERROR: not in the mycelium repo"; exit 1; }
+```
+
+If any prerequisite fails, fix it before proceeding. Throughout this skill, use `$MYCELIUM_API_URL` for backend requests; never hardcode a port.
+
+### 0a. Verify Synapse registration + admin
+
+This skill creates throwaway Matrix users via the public registration endpoint. Confirm registration is open on this Synapse:
+
+```bash
+docker exec $(docker ps --format '{{.Names}}' | grep -E 'matrix|synapse' | head -1) \
+  grep -E "enable_registration|server_name" /data/homeserver.yaml
+# Need: enable_registration: true (or enable_registration_without_verification: true).
+# Capture server_name — Matrix user IDs will be @<name>:<server_name>.
+```
+
+Stash the server name:
+
+```bash
+SYNAPSE_CONTAINER=$(docker ps --format '{{.Names}}' | grep -E 'matrix|synapse' | head -1)
+SYNAPSE_SERVER=$(docker exec "$SYNAPSE_CONTAINER" grep "^server_name:" /data/homeserver.yaml | awk '{print $2}' | tr -d '"')
+echo "SYNAPSE_SERVER=$SYNAPSE_SERVER"
+```
+
+If registration is locked down, you'll need either the `registration_shared_secret` from `homeserver.yaml` or pre-existing accounts. The version in this skill assumes open registration; adapt as needed.
+
+## Phase 0.5: Choose experiment LLM & API key
+
+Same logic as the standard `before-and-after` skill — default to Haiku unless the user explicitly wants Sonnet. Each negotiation fires 10–40+ LLM calls; cost differs ~12×.
+
+```bash
+CONFIGURED_MODEL=$(python3 -c "
+import json, os
+p = os.path.expanduser('~/.openclaw/openclaw.json')
+try:
+    cfg = json.load(open(p))
+    print(cfg.get('agents', {}).get('defaults', {}).get('model', ''))
+except Exception:
+    print('')
+")
+echo "Currently configured model: ${CONFIGURED_MODEL:-'(none)'}"
+```
+
+Use `AskUserQuestion` to pick: Haiku (default) / Sonnet / configured / different key.
+Set `EXP_MODEL` accordingly. Echo the choice into the transcript:
+
+```bash
+echo "Using model: $EXP_MODEL"
+```
+
+## Phase 1: Setup
+
+### 1a. Generate experiment IDs + handles
+
+```bash
+EXP_ID="xch-$(date +%s | tail -c 5)"  # e.g. xch-4821
+EXP_AGENT_A="${EXP_ID}-agent-a"
+EXP_AGENT_B="${EXP_ID}-agent-b"
+EXP_HUMAN="${EXP_ID}-human"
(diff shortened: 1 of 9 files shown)


juliarvalenti:
@sfph — this PR is a partial answer to what your distributed E2E test ran into.

Your prompt to `agent-alpha` and `claire-agent` ended with step 5: *"Report back here with a summary of the consensus."* As written, that asks the agents to remember and execute the return trip themselves — and the negotiation, after running its rounds, just stops in Mycelium with no path back to Matrix. Even with the agents fully bought into the protocol, "report back here" wasn't a thing they could actually do without a `sessions_send` step they don't reliably reach for. We empirically tested this: even when explicitly told to use `sessions_send` with the exact session key, the agent would acknowledge the instruction, do the negotiation, and then fail to deliver. The agent-side return trip is too fragile.

What this PR makes happen:

- When the negotiation concludes (agreement *or* timeout), the plugin auto-posts a one-shot summary back to whichever Matrix DM the agent was active in when it joined.
- Your step 5 becomes infrastructure, not an instruction. You write `Topic: ...` and `Room: ...`, the agents do the negotiation, the result lands in your DM. No `sessions_send`, no `Report back here`, no relying on the agent to remember.

What this does **not** fix (separate followup, see PR description):

- Your test used a per-experiment dynamic room name (`dist-e2e-22e5fd09`). The channel plugin currently only subscribes to session sub-rooms whose parent matches `channels.mycelium-room.room` in `openclaw.json`. So if your gateway was bound to a different room, the channel plugin never saw the session and dispatch silently failed — *that's why the agents went quiet, not anything to do with the return trip.* The fix shape is "subscribe per-agent across all sessions, not per-room"; it's a separate small change worth making after this lands. For now, easiest workaround: set `channels.mycelium-room.room` to the room your test prompt names (or vice versa).

If you re-run your distributed E2E test against this branch with the room name aligned, you should:
1. See `🎯 → agent-alpha`, `🎯 → claire-agent` ticks fire as expected
2. See `return-address stashed: agent-alpha → matrix:room:!XXX:local` (and same for claire) on first tick
3. See `📬 notify-home for [agent-alpha, claire-agent] in <session-room>` on consensus
4. See `[Mycelium return trip — <room>] [CognitiveEngine — Consensus Reached!] ...` actually land in the Matrix DM you started from

Happy to walk through it together if you want to repro before this merges.

juliarvalenti:
## Update: folded in #220 (channel binding scope)

After review feedback that the brittle channel binding undermined the return-trip's UX promise (you could ship the auto-delivery but users would still hit silent failures whenever their agents negotiated in a room name different from their `channels.mycelium-room.room`), pulled #220's first item into this PR.

### Change

`channel/index.ts`'s polling loop no longer filters session sub-rooms by parent room name. The plugin subscribes to every active `*:session:*` room. The participant filter that was already running on every tick (`routeTick` checks `cfg.agents.includes(participant_id)`) is the actual gate; the parent-room scoping was redundant and harmful.

### One new gate added

`routeConsensus` blindly fans the consensus summary out to every agent in `cfg.agents` — fine when we were room-scoped, but with broader subscription that would wake our agents' mycelium-room sessions with consensus from negotiations they had nothing to do with. Added an executor-side check: when dispatching a `coordination_consensus`, skip agents that don't have a return-address stashed for that specific session room. The stash is the authoritative participation signal (populated on first tick).

### Verification

Tested with `channels.mycelium-room.room = "unbound-default"` and agents negotiating in `xch-220-test-*`. End-to-end:

- Plugin subscribed to the unbound session sub-room (would not have under old code)
- Both agents stashed on first tick; ticks dispatched normally; consensus reached
- `📬 notify-home for [agent-rho, agent-sigma] in xch-220-test-NNN:session:XXX`
- Matrix DM landed: `[Mycelium return trip — xch-220-test-NNN] [CognitiveEngine — Consensus Reached!] ...`
- **Label fix**: the `[Mycelium return trip — ...]` header now shows the actual negotiation room, parsed from the session sub-room name, rather than `cfg.room`. Previously would have said "unbound-default" which would be confusing.

Also caught the symmetric "only one participant" case in the same run: when one agent joined a session and the other didn't, the consensus dispatch was correctly skipped for the non-participant with `consensus skipped for <agent> — not a participant in <session>` and delivered to the actual participant. Both directions of the gate working.

### Net result

The return-trip ships with the rough edge already smoothed: any room name works, no `channels.mycelium-room.room` alignment required. Users don't have to learn about a brittle binding and then forget to align it later.

The before-and-after-matrix skill was updated to reflect this — the previous "failure-mode variant" (intentionally misalign the binding to repro silent failure) is now an "unbound-room variant" that confirms room-agnostic routing still works. Same regression-test value, opposite polarity.

PR #228, opened 2026-05-02
feat(promo): re-number scenes, openclaw label, read pause for memory set

juliarvalenti:
## Summary

Leader review feedback applied to the promo:

- **Scene numbers swapped**: 5b is now `05 / Trigger from any chat`, s6 is `06 / Agents coordinate` (was inverted)
- **Scene 5 title** clarified: `Wire in your agents` → `Wire in your OpenClaw agents`
- **+2s pause** in scene 4 after `mycelium memory set` so the viewer can read the terminal output before the room UI slides up

Re-rendered MP4 (~6 MB) committed.

## Next step

After merge: I'll drag-drop the new MP4 into a comment to mint a fresh `user-attachments` URL, then swap it into README + docs.

## Test plan

- [ ] Watch the rendered MP4 — verify the renumber lands at the right scene transitions and the memory-set pause feels right


juliarvalenti:
https://github.com/user-attachments/assets/1a5febbb-87e7-48a4-aa7d-8d1b116889c3

Issue #234, opened 2026-05-05
Backend route prefix convention is inconsistent (`/rooms` vs `/api/*`)


juliarvalenti:
## Context

While renaming `/api/metrics` → `/observability` to dodge a privacy-extension blocklist (#TBD), we audited the backend's URL paths. The current state:

**No prefix:**
- `/health`
- `/rooms*` (rooms, sessions, memory, messages, negotiation, synthesize, …)
- `/observability`, `/observability/collector`
- `/v1/memories`
- `/agents/{handle}/stream`

**`/api/` prefix:**
- `/api/cfn/knowledge/*` (CFN read proxy)
- `/api/workspaces/.../memory-operations` (CFN write proxy)
- `/api/knowledge/ingest*`
- `/api/internal/audit-events*`
- `/api/internal/coordination/round-traces`

There's no documented rule. It looks like an emergent split: room-first endpoints stay top-level, proxies / internal services get `/api/`. New endpoints have to guess.

## Why it matters

1. **Inconsistency surprises clients.** Adding a new endpoint requires reading other routes and inferring the convention.
2. **`/api/metrics*` got pattern-blocked by privacy extensions** (uBlock-style). Forced the rename to `/observability`. Same risk applies to other `/api/*` paths whose names look telemetry-shaped — `/api/knowledge/ingest/stats`, `/api/internal/audit-events`, etc. None have frontend callers today, but they're tripwires for future UI work.
3. **Mixed pattern complicates CORS / proxy / auth setup** — middleware that wants to apply only to "API" routes has no clean way to select them.

## Options

1. **Pick one prefix and migrate.** Either everything top-level (`/rooms`, `/cfn`, `/observability`, `/internal/...`) or everything under `/api/*`. Whichever wins, remove the other.
2. **Document the existing split** as a deliberate rule (e.g., "room-resource endpoints are top-level; everything else is `/api/*`") and police it in PR review.
3. **Leave as-is** and accept the ambiguity.

I'd lean (1) — top-level — because the room-first API is the public surface and \`/api/\` adds no information when the whole service IS the API. But (2) is also defensible.

## Out of scope

- Don't touch CFN proxy paths in this issue — they need to match upstream contracts.

Issue #233, opened 2026-05-04
Unify metrics collector startup with `mycelium up --metrics`


juliarvalenti:
## Context

`mycelium up --ui` (PR #217) established a pattern: stack-up is the single entry point, with flags for additional local components.

The metrics collector (PR #60) currently uses a separate command surface: `mycelium metrics collect` / `stop` / `status`. For local dev/testing, this means two commands to bring up "the whole thing":

```
mycelium up --ui
mycelium metrics collect
```

## Proposal

Add `--metrics` to `mycelium up`, mirroring `--ui`:

```
mycelium up --ui --metrics
```

- `up --metrics` → start stack, then start collector subprocess (equivalent to `metrics collect`)
- `down` → also stop the collector if running, so lifecycles stay paired
- Keep `mycelium metrics collect` / `stop` / `status` as escape hatches for advanced use (foreground mode, restart without bouncing the stack, etc.)

## Tradeoff to be aware of

The UI is a compose container; the collector is a host-side Python subprocess with a PID file in `~/.mycelium/`. So `up --metrics` is really sugar for `up && metrics collect` rather than another compose service. That's fine, but the wiring lives outside docker compose and `down` needs explicit teardown logic.

## Out of scope

- Moving the collector into compose (it needs to bind `:4318` on the host for OpenClaw OTLP export — keeping it host-side is intentional).
- Changing the `metrics show` UX.

## References

- `mycelium-cli/src/mycelium/commands/instance.py:306` — current `--ui` flag wiring
- `mycelium-cli/src/mycelium/commands/metrics.py` — collect/stop entry points

juliarvalenti:
Can we possibly get rid of `mycelium adapter add openclaw --step otel`, just add otel block automatically

Inform user we're making changes to `openclaw.json`, take backup if necessary.  Is this flow already there?  I thought it was.

PR #235, opened 2026-05-05
feat(frontend): metrics dashboard redesign + rename /api/metrics → /api/observability

juliarvalenti:
## Summary

Replaces the metrics style picker (`/metrics/cards` + `/metrics/tables`) with a single opinionated dashboard built on the existing Onyx atoms already in `globals.css` (caps-mono, square-dot, tabular nums, the cyan accent).

Also renames `/api/metrics*` → `/api/observability*` because privacy extensions (uBlock-style) silently pattern-block `/api/metrics*` as analytics telemetry. See [Investigation](#investigation) below.

### Layout

```
┌─ MainTopBar (METRICS active)
├─ SubNav (metrics)
├─ HeaderBand            up Xm · updated 2s ago     EVERY [5|10|30|60s]  [● live | □ collector on]
├─ KPI band              SPEND       ERROR RATE       ACTIVITY
├─ AgentActivityTable    (or "WAITING ON OTLP" plate w/ adapter command)
├─ Diagnostics           latency strip → 5-tile grid → 2-tile LLM strip
└─ ByModelTable          (only when models have spent)
```

### Design rules followed

- **Single accent** (`--accent`). No brick red. Error states signal via caps eyebrow + italic helper text, not a second color.
- **Honest time framing.** All KPI numerals are cumulative since data source start. Time-window labels (`TODAY`, `LAST 5m`, `LAST 24h`) removed because the backend has no rolling-window aggregation. Sparklines (collector-emitted `spend_5m` / `err_5m`) carry recency.
- **Disabled-mode plates** — collector off shows a folded blueprint corner with the exact CLI; backend down shows a brick-rule plate with the unreachable URL.
- **Hover affordance** — the `collector off` chip reveals a popover with the disabled-mode copy and the `mycelium metrics collect` command, instead of using OS-styled `title`.

### Investigation: why `/api/metrics*` gets blocked

While testing in Firefox: `/rooms*` fetches succeeded, `/api/metrics*` fetches failed with `CORS request did not succeed` and `Status: null` — request never reached the server. CORS is correctly configured (curl with `Origin: localhost:3000` returns proper headers). The split is **path-pattern based**: ad blockers' built-in lists pattern-match `/api/metrics*` as common analytics endpoints and drop the request before it leaves the browser. Renaming the route fixes it for every developer who runs the stack.

Same risk applies to other suggestively-named `/api/*` paths (`/api/internal/audit-events`, `/api/knowledge/ingest/stats`) — none have a frontend caller today, so they're fine for now. Tracked in #234.

## Files

| Area | Files |
|------|-------|
| Frontend (new) | `components/metrics-screen.tsx` (the page), `lib/metrics-format.ts` |
| Frontend (rewritten) | `app/metrics/page.tsx` (now thin wrapper) |
| Frontend (deleted) | `app/metrics/cards/`, `app/metrics/tables/`, `components/metric-card.tsx` |
| Frontend (touched) | `lib/api.ts` (URL rename) |
| Backend | `app/main.py` (2 routes), `app/services/metrics.py` (docstring) |
| CLI | `collector.py`, `collector_main.py`, `commands/metrics.py`, `docs/metrics.md` |
| Generated | `openapi.json` snapshot |

## Related issues

- **#233** — unify metrics startup with `mycelium up --metrics` (UX consistency with `--ui`)
- **#234** — backend route prefix convention is inconsistent (`/rooms` vs `/api/*`)

## Known gaps (non-blocking, follow-ups)

- Backend histograms expose only `count/sum/min/max` — latency strip renders **avg/min/max** instead of p50/p95/p99. Wiring buckets is a separate change.
- Collector data file is read at `~/.mycelium/metrics.json` from inside the backend container. Works in dev because the host directory is bind-mounted, but `Path.home()` resolution is fragile — worth a follow-up to make the path explicit.
- Backend in-process counters reset on restart. Annoying in dev iteration; non-issue for end users on a long-lived backend.

## Test plan

- [x] `pnpm dev` → `/metrics` renders with backend data populated
- [x] Generate traffic (`mycelium memory set/search`) → diagnostic tiles + latency strip update on next 10s tick
- [x] Stop the collector → page shows the "DISABLED MODE" folded-corner plate with the CLI command
- [x] Stop the backend → page shows "BACKEND UNREACHABLE" plate
- [x] Hover `collector off` → popover with the exact `mycelium metrics collect` command
- [x] `npx tsc --noEmit` clean
- [x] `uv run pytest tests/ -x -q` (207 passed) + `uv run ruff check` clean
- [ ] Verify `/api/observability` is not on common adblock filterlists for non-dev users (Firefox + uBlock + Chrome + AdBlock+)
- [ ] Once OpenClaw OTLP is wired (`mycelium adapter add openclaw --step=otel`), confirm AgentActivityTable + ByModelTable populate from real per-agent traces


juliarvalenti:
Screenshot 2026-05-04 at 5 37 40 PM

juliarvalenti:
Screenshot 2026-05-05 at 10 25 02 AM

Issue #240, opened 2026-05-07
Centralize --profile cfn handling in _compose_base_cmd


juliarvalenti:
Follow-up from #236 review.

In `mycelium-cli/src/mycelium/commands/instance.py`, both `stop` and `logs` now manually append `--profile cfn` when `_cfn_enabled()`:

```python
base = _compose_base_cmd(compose_path, project_name=project)
if _cfn_enabled():
    base = base + ["--profile", "cfn"]
```

This is duplicated logic and a footgun — any new command that calls `_compose_base_cmd` will silently miss CFN services unless the author remembers to re-add the profile flag (which is exactly the bug #236 fixed for `logs`/`down`).

### Proposal
Fold profile handling into `_compose_base_cmd` itself, with an opt-out for callers that explicitly don't want it:

```python
def _compose_base_cmd(
    compose_path: Path | None = None,
    env_path: Path | None = None,
    project_name: str | None = None,
    include_cfn_profile: bool = True,
) -> list[str]:
    ...
    if include_cfn_profile and _cfn_enabled():
        cmd += ["--profile", "cfn"]
    return cmd
```

Then `stop`/`logs`/future commands just call `_compose_base_cmd(...)` and get correct behavior by default.

### Acceptance
- `stop` and `logs` no longer manually append `--profile cfn`.
- New commands using `_compose_base_cmd` automatically include the CFN profile when enabled.

juliarvalenti:
Fixed in #242.

PR #242, opened 2026-05-07
fix: address listener leak and centralize --profile cfn (#239, #240)

### LINKED ISSUE #239: Listener leak on gateway AbortController in session-sse.ts
Follow-up from #238 review.

In `mycelium-cli/src/mycelium/adapters/openclaw/mycelium/plugin/src/channel/session-sse.ts`, `startSessionSSE` registers a listener on the long-lived gateway `AbortController`:

```ts
_gatewayAbort.signal.addEventListener("abort", () => {
  sessionAbort.abort();
});
```

The listener is never removed when the session stops via `stopSessionSSE` (terminal state, deleted room, or max-retries give-up). Over a long-running gateway with many sessions, listeners accumulate on the same signal — a slow leak, but unbounded.

### Fix options
- Hold a reference to the listener and remove it in `stopSessionSSE`:
  ```ts
  const onGatewayAbort = () => sessionAbort.abort();
  _gatewayAbort.signal.addEventListener("abort", onGatewayAbort, { once: true });
  // store onGatewayAbort alongside the controller; removeEventListener on stop
  ```
- Or use `AbortSignal.any([_gatewayAbort.signal, sessionAbort.signal])` if Node version supports it (Node 20+).

### Acceptance
- New session subscriptions don't accumulate listeners on `_gatewayAbort.signal` after their `stopSessionSSE` runs.
- Unit test asserts listener count stays bounded across N start/stop cycles.

sfph:
## Summary

Fixes #239 and #240, both filed as follow-ups to #238.

- **#239 — Listener leak on gateway AbortController**: `startSessionSSE` registered an anonymous arrow function on `_gatewayAbort.signal` that could never be removed. Each session start/stop cycle leaked one listener. Fix: store a named reference in a `SessionHandle` and call `removeEventListener` in `stopSessionSSE()` and `clearSubscribedSessions()`.

- **#240 — Centralize `--profile cfn` handling**: Three call sites manually appended `--profile cfn` after `_compose_base_cmd()`, and `down` was already missing it. Folded the `_cfn_enabled()` check into `_compose_base_cmd()` with an `include_cfn_profile` kwarg (default `True`) so all callers get correct behavior automatically.

## Test plan

- [x] Plugin vitest: 109/109 passing
- [x] Backend pytest: 207/207 passing
- [x] CLI ruff lint + format: clean
- [x] CLI + backend type check (ty): clean
- [x] Frontend tsc + next build: clean
- [x] E2E suite running with fix deployed to all 3 gateways

### DIFF
--- a/mycelium-cli/src/mycelium/adapters/openclaw/mycelium/plugin/src/channel/session-sse.ts
+++ b/mycelium-cli/src/mycelium/adapters/openclaw/mycelium/plugin/src/channel/session-sse.ts
@@ -24,28 +24,35 @@ type HandleMessageFn = (runtime: any, cfg: ChannelConfig, msg: any, log: Logger)
 
 const MAX_TRANSIENT_RETRIES = 6;
 
-const _sessionControllers = new Map<string, AbortController>();
+interface SessionHandle {
+  ctrl: AbortController;
+  detach: () => void;
+}
+
+const _sessions = new Map<string, SessionHandle>();
 
 export function clearSubscribedSessions(): void {
-  for (const [, ctrl] of _sessionControllers) {
-    ctrl.abort();
+  for (const [, h] of _sessions) {
+    h.detach();
+    h.ctrl.abort();
   }
-  _sessionControllers.clear();
+  _sessions.clear();
 }
 
 /** Abort and remove a single session subscription. */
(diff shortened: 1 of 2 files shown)


juliarvalenti:
Thanks for getting to this so fast!

Issue #266, opened 2026-05-12
channel: agent replies silently dropped when host openclaw.json has messages.groupChat.visibleReplies != "automatic"


akanksha276:
### Affected Version

1.0.9 and 1.0.10rc8

### Description

Root cause: dispatch.ts uses ChatType: "group" but the plugin has no control over messages.groupChat.visibleReplies in the host's openclaw.json. When set to anything other than "automatic" (the default for WhatsApp installs is "message_tool"), the dispatcher's block enqueue is suppressed and deliver is never called.

Affected versions: Any version of the plugin running against an OpenClaw install where messages.groupChat.visibleReplies !== "automatic" — confirmed on rc8 and 1.0.9.
 
Fix: Add replyOptions: { sourceReplyDeliveryMode: "automatic" } to the dispatchReplyWithBufferedBlockDispatcher call in dispatch.ts, making the plugin's delivery intent explicit and independent of host group-chat config.

### Steps to Reproduce

Reproduction: Any openclaw.json where messages.groupChat.visibleReplies is set to "message_tool"  
The fact that no error is logged — suppression is silent at logVerbose level making this hard to diagnose

### Expected Behavior

Agent text replies are posted to the Mycelium room (i.e. deliver is called and postToRoom succeeds) regardless of the host's group chat display settings.

### Actual Behavior

Agent runs and produces text (visible in session .jsonl and trajectory files), but deliver is never called and no message appears in the room. No error or  warning is logged — the suppression is completely silent.

### Environment

OpenClaw version: 2026.5.7 (eeef486)
Mycelium plugin version: 1.0.10rc8 and 1.0.9

### Troubleshooting Attempted

- [x] I ran `mycelium status` and reviewed the output
- [x] I checked the logs with `mycelium logs`
- [x] I used the troubleshooting skill or reviewed the troubleshooting guide

akanksha276:
resolved by PR https://github.com/mycelium-io/mycelium/pull/267

PR #267, opened 2026-05-12
fix(plugin): prevent silent reply suppression in group-chat dispatch

akanksha276:
## Problem

When a host `openclaw.json` has `messages.groupChat.visibleReplies` set to anything other than `"automatic"` (e.g. the WhatsApp default `"message_tool"`), the dispatcher's block-enqueue is suppressed and `deliver` is never called. Agent text is produced (visible in session `.jsonl` and trajectory files) but nothing appears in the Mycelium room. No error or warning is logged — suppression is completely silent.

Reported in Issue https://github.com/mycelium-io/mycelium/issues/266

Affects: plugin 1.0.9 and 1.0.10rc8 running against OpenClaw 2026.5.7 (eeef486).

## Fix

Add `replyOptions: { sourceReplyDeliveryMode: "automatic" }` to the `dispatchReplyWithBufferedBlockDispatcher` call in `dispatch.ts`. This makes the plugin's delivery intent explicit and independent of the host group-chat config.

## Test plan

- [ ] Deploy plugin against an OpenClaw install with `messages.groupChat.visibleReplies: "message_tool"` and verify agent replies appear in the Mycelium room
- [ ] Verify no regression on installs with the default `"automatic"` setting
- [ ] Confirm no error logs on either path

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- OpenClaw plugin: add `replyOptions: { sourceReplyDeliveryMode: "automatic" }` to
+  `dispatchReplyWithBufferedBlockDispatcher` so agent replies are delivered regardless
+  of the host's `messages.groupChat.visibleReplies` setting (fixes silent suppression
+  when set to `"message_tool"`)
+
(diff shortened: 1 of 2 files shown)


juliarvalenti:
Thanks Akanksha! Clean one-line fix and great catch — we tripped over this exact behaviour testing rc7 against openclaw 2026.5.7 yesterday.

PR #271, opened 2026-05-12
fix(coordination): fuzzy-snap counter-offer keys before validation

sfph:
## Summary
- Ports the four-tier fuzzy snap logic from the CLI's `negotiate_snap.py` (PR #264) into the backend's `_validate_and_fill_offer`
- LLMs routinely convert multi-word offer keys to snake_case (e.g. `ACID_compliance` instead of `ACID compliance`), causing the strict set-difference check to reject every counter-offer and stall negotiations indefinitely
- Snap tiers: exact → case-insensitive → normalised (collapse `_`/`-`/whitespace) → rapidfuzz `token_set_ratio` ≥ 80 — mirrors engines `offer_validation.py` and CLI `snap_issue`

## Test plan
- [x] Existing `test_coordination.py` tests updated and passing (38/38)
- [x] New tests for case-insensitive snap, underscore snap, fuzzy snap, and partial fill after snap
- [x] E2E test_42 (distributed architecture decision) now passes — previously failed 100% with `bad keys ['ACID_compliance', 'horizontal_scaling', ...]`
- [x] Full 30+40 series E2E suite running (30-32 and 40 passed so far)

### DIFF
--- a/fastapi-backend/app/services/coordination.py
+++ b/fastapi-backend/app/services/coordination.py
@@ -22,6 +22,7 @@ coordination_error message and the room is set to "failed" state.
 import asyncio
 import json
 import logging
+import re
 import time
 from collections import deque
 from dataclasses import dataclass, field
@@ -30,6 +31,7 @@ from typing import Literal
 from urllib.parse import urlparse
 
 import asyncpg
+from rapidfuzz import fuzz as _fuzz
 from sqlalchemy import delete, select, update
 
 from app.bus import agent_channel, notify, room_channel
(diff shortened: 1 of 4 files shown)


juliarvalenti:
Thanks Selina! Server-side defense in depth is exactly the right complement to PR #264's CLI-side snap. Merging.

Filing a follow-up to extract the snap module into a shared `mycelium-shared` package both CLI and backend depend on — there's now identical logic in two places in this repo and it'll drift the moment anyone tweaks a threshold.

Issue #278, opened 2026-05-14
release.yml: bump plugin package.json version alongside CLI + backend


juliarvalenti:
## Problem

`mycelium-cli/src/mycelium/adapters/openclaw/mycelium/plugin/package.json` is still at `0.1.0` on main and stays there through every release. `openclaw plugins list` confirms:

```
│ Mycelium     │ mycelium │ openclaw │ enabled  │ global:mycelium/dist/index.js │ 0.1.0 │
│ Adapter      │          │          │          │ ...                           │       │
```

Users running CLI 1.0.10 see the plugin reporting 0.1.0, which is misleading for bug reports and adapter status output.

## Root cause

`.github/workflows/release.yml` bumps:

- `fastapi-backend/pyproject.toml` (release.yml:104-111) — for `/health`
- `mycelium-cli/pyproject.toml` x2 (release.yml:161-165 binary, 214-218 wheel) — for `mycelium --version`
- `homebrew-tap/Formula/mycelium.rb` (release.yml:310) — for brew

It does **not** bump `mycelium-cli/src/mycelium/adapters/openclaw/mycelium/plugin/package.json`. Since this file is bundled into the CLI wheel/binary and copied into `~/.openclaw/extensions/mycelium/` by `mycelium adapter add openclaw`, its `version` field is the authoritative one OpenClaw reads at plugin-list time.

## Fix

Add a "Bump plugin package.json version" step in `release.yml`, mirroring the existing CLI and backend bumps, before the wheel/binary build (so the bumped value is included in the package contents):

```yaml
- name: Bump openclaw plugin package.json version to match tag
  working-directory: mycelium-cli/src/mycelium/adapters/openclaw/mycelium/plugin
  run: |
    VERSION=\"\${GITHUB_REF_NAME#v}\"
    # node-friendly: use jq to avoid corrupting the JSON
    jq --arg v \"\$VERSION\" '.version = \$v' package.json > package.json.tmp
    mv package.json.tmp package.json
```

This needs to run **in both** the `build-binaries` job (release.yml:140-198) **and** the `build-and-release` wheel job (release.yml:200-273) since each performs an isolated checkout. Same pattern as the existing CLI pyproject bump in both jobs.

## Verification

After release, `openclaw plugins list | grep Mycelium` should show the tag version, not `0.1.0`.

## Related

[...]

PR #279, opened 2026-05-14
fix(plugin): stash tick payload from channel for session/before_agent_start (#276)

### LINKED ISSUE #276: before_agent_start injects tick context from wrong room — agents never see CE offer keys
### Affected Version

1.0.10

### Description

Summary

  The before_agent_start hook in the OpenClaw session plugin fetches coordination context from the parent room, but coordination_tick messages are posted exclusively to the session sub-room. Agents receiving a proposer turn never get the current_offer payload injected into their context, so they must guess CE offer key names when composing a counter-offer. Those guesses fail CE validation, the session times out, and structured negotiation never completes a round.

  ---
  Affected file
  
  mycelium-cli/src/mycelium/adapters/openclaw/mycelium/plugin/src/session/index.ts (line ~172 on main)

  ---
  The bug

 ```
 // before_agent_start handler — src/session/index.ts
  const room = existing?.room;
  if (room) {
    const data = (await apiGet(`/api/rooms/${room}/messages?limit=30`, log)) as any;  // ← wrong room
    const coord = data?.messages?.find(
      (m: any) =>
        m.message_type === "coordination_consensus" ||
        m.message_type === "coordination_tick",
    );
    // coord is always undefined — ticks are never in the parent room
  }
```

  room is the parent room (e.g. exp-7797-after). The backend posts coordination_tick messages to the session sub-room (exp-7797-after:session:<short-id>). The parent room query always returns nothing for
   ticks, so the current_offer with exact CE issue key names is never injected into prependContext.

  ---
  Impact

  Reproduced in three independent test contexts:

  1. before-and-after skill (main branch) — exp-0203

  Two-agent API design negotiation (REST vs GraphQL). Agent submitted keys public_api, automatic_SDKs (underscore form) instead of public API, automatic SDKs. CE returned counter_offer_invalid_keys.
  Plugin logs showed prependContext: 0 chars (dynamic) on every tick turn. Session timed out.

  2. persona-before-and-after skill (persona-before-and-after branch) — exp-7797

  Three-agent email automation negotiation. Experiment failed at the same point: proposer submitted invented key names (archival-period, retention_window, etc.) instead of the CE's extracted natural-language keys. prependContext: 0 chars confirmed injection was never happening. 

  3. Direct minimal test (no skill) — rooms issue4-3205

  Two agents negotiating cache TTL and eviction policy, created directly without any skill scaffolding. Confirmed:
  -  prependContext: 0 chars (dynamic) on all tick turns. Agent submitted cache_ttl / eviction_policy → counter_offer_invalid_keys. Zero valid counter-offers.

  ---
  Root cause detail

  The SessionEntry tracked by installSession stores the OpenClaw session ID (e.g. 83171a45-ae4f-4ead-bcfc-91767ea0a20a), which is unrelated to the Mycelium coordination session short ID (e.g. 230d6338)
  used in the sub-room name {room}:session:{short-id}. There is no existing field on SessionEntry that resolves to the Mycelium coordination session room name.

  The channel module (src/channel/index.ts) discovers session sub-rooms correctly — it polls /api/coordination-sessions?limit=200 and uses the display_name field (e.g. issue4-fix2-4435:session:230d6338). The session module needs to do the same.

  ---
  Fix

  Add a coordSessionRoom cache field to SessionEntry, resolve it once from /api/coordination-sessions on first tick turn, and query the sub-room for tick context:

```
  type SessionEntry = {
    sessionKey: string;
    sessionId?: string;
    handle: string;
    room?: string;
    coordSessionRoom?: string; // cached Mycelium coordination session sub-room name
  };

  const room = existing?.room;
  if (existing && room) {
    let coord: any;

    // Ticks are posted to the session sub-room, not the parent room.
    // Resolve the Mycelium coordination session sub-room via the API and
    // cache it on the session entry so agents see the full current_offer
    // payload with exact CE key names before composing a counter-offer.
    if (!existing.coordSessionRoom) {
      const sessionsData = (await apiGet(
        `/api/coordination-sessions?limit=200`,
        log,
      )) as any;
      const sessions: any[] = Array.isArray(sessionsData) ? sessionsData : [];
      const active = sessions.find(
        (s: any) =>
          typeof s.display_name === "string" &&
          s.display_name.startsWith(`${room}:session:`) &&
          s.state !== "failed" &&
          s.state !== "idle",
      );
      if (active?.display_name) {
        existing.coordSessionRoom = active.display_name;
        log.info(`[mycelium] resolved coord session sub-room: ${existing.coordSessionRoom}`);
      }
    }

    if (existing.coordSessionRoom) {
      const sessionData = (await apiGet(
        `/api/rooms/${existing.coordSessionRoom}/messages?limit=5`,
        log,
      )) as any;
      coord = sessionData?.messages?.findLast(
        (m: any) =>
          m.message_type === "coordination_consensus" ||
          m.message_type === "coordination_tick",
      );
    }

    // Fall back to parent room for coordination_consensus (which lands there)
    if (!coord) {
      const data = (await apiGet(`/api/rooms/${room}/messages?limit=30`, log)) as any;
      coord = data?.messages?.find(
        (m: any) =>
          m.message_type === "coordination_consensus" ||
          m.message_type === "coordination_tick",
      );
    }

    if (coord) {
      const label =
        coord.message_type === "coordination_consensus"
          ? "[Mycelium — consensus]"
          : "[Mycelium — coordination tick]";
      contextParts.push(`${label}\nRoom: ${room}\n\n${coord.content}`);
    }
  }
```

  ---
  Notes
  
  - A second failure mode survives this fix: when a proposer's counter-offer is rejected (e.g. wrong keys from a later round), the error tick's instruction field — which lists valid_keys — is not rendered by formatTickInstruction in route.ts, so the agent sees a bare "Action required: respond" with no recovery path. That should be tracked separately as another Issue (formatTickInstruction drops error tick instruction).
  - The limit=5 sub-room query is intentional — ticks are always the most recent messages in the sub-room, so a small window is sufficient and avoids pulling full message history on every agent turn.
  - The coordSessionRoom cache on SessionEntry avoids a /api/coordination-sessions call on every turn after the first resolution. The cache is per-agent-session so it naturally expires when the agent session ends.

### Steps to Reproduce

 1. Start the Mycelium stack with CFN enabled (mycelium up --profile cfn)
  2. Create a room: mycelium room create test-room
  3. Add two OpenClaw agents with sandbox off and mycelium-room channel configured pointing at test-room
  4. Restart the gateway: openclaw gateway restart
  5. Post a seed message to test-room @-mentioning both agents, instructing them to run mycelium session join and then negotiate using mycelium negotiate propose
  6. Wait for both agents to join and for the CognitiveEngine to fire Round 1 ticks
  7. Observe before_agent_start in the gateway log and the prependContext size:
  [mycelium] prependContext: 0 chars (dynamic)
  8. Observe the proposing agent submit a counter-offer with invented key names and receive counter_offer_invalid_keys from the CE

### Expected Behavior

When a coordination_tick fires and the channel plugin dispatches it to an agent, before_agent_start should inject the full tick payload — including the current_offer object with exact CE issue key names — into the agent's prependContext before the model call. The agent should then be able to use those exact key names when running mycelium negotiate propose, producing a valid counter-offer accepted by the CognitiveEngine.

Gateway log should show:
  [mycelium] resolved coord session sub-room: test-room:session:<short-id>
  [mycelium] prependSystemContext: N chars (cached), prependContext: N chars (dynamic)

And the session sub-room should receive a direct message with offer_keys matching the CE's extracted issue keys — no counter_offer_invalid_keys error.

### Actual Behavior

  before_agent_start queries /api/rooms/{parent-room}/messages?limit=30. The backend never posts coordination_tick messages to the parent room — they go exclusively to the session sub-room
  ({room}:session:{short-id}). The query always returns empty for ticks, so coord is always undefined and nothing is injected.

  Gateway log shows:
  [mycelium] prependContext: 0 chars (dynamic)
  on every tick turn, for every agent, regardless of round.

  Without the injected current_offer, the proposing agent has no way to know the CE's exact extracted key names. It invents semantically reasonable names (cache_ttl, eviction_policy, retention_window,
  etc.) which the CE rejects:

  ```
{
    "error": "counter_offer_invalid_keys",
    "bad_keys": ["cache_ttl", "eviction_policy"],
    "valid_keys": ["Cache TTL", "600 seconds", "reduces DB load by 40%", ...]
  }
```
  The agent cannot recover (compounded by another issue — the error tick's instruction field is not rendered by formatTickInstruction), and the session times out after 300s.

### Environment

Mycelium CLI version 0.1.0
Mycelium backend 0.1.0
OpenClaw 2026.5.7 (eeef486)
Node.js v24.13.1
OS 6.17.0-1007-aws
Python 3.12.3

### Troubleshooting Attempted

- [x] I ran `mycelium status` and reviewed the output
- [x] I checked the logs with `mycelium logs`
- [x] I used the troubleshooting skill or reviewed the troubleshooting guide

juliarvalenti:
## Summary

Closes #276 (the bug as filed; impact was masked by #271 fuzzy-snap).

`before_agent_start` was scanning the **parent** room's messages for the latest `coordination_tick` — but ticks attach to the `CoordinationSession`, never to the parent room, so the scan always returned nothing and `prependContext` stayed at **0 chars (dynamic)** on every turn. Agents had to guess CE offer key names; fuzzy-snap papers over the symptom but the context-injection path itself never worked.

Rather than teach `session/` to resolve coord-session display names and walk `/api/coordination-sessions/{id}/messages` (the path the original issue proposed), this mirrors the existing **return-address** pattern already in the codebase: channel observes the tick once on dispatch, stashes the raw payload keyed by `agentId`, `before_agent_start` reads it back. No HTTP in the hot path, no display-name parsing, no compat-route dependency.

## Changes

- New `channel/tick-stash.ts` — in-memory `Map<agentId, StashedTick>`, latest-tick-wins, mirroring `channel/return-address.ts` in structure and lifecycle.
- `routeTick` emits `stash-tick` alongside `stash-return-address` + `dispatch` when the tick's `room_name` names a session sub-room.
- `channel/index.ts` executor handles `stash-tick`; clears on session terminal state (alongside `stopSessionSSE`) and on `gateway_stop` (alongside `clearSubscribedSessions`).
- `session/index.ts` `before_agent_start` drops the parent-room HTTP fetch in favour of `lookupTick(agentId)`; injects the full payload as pretty-printed JSON.
- Tests: new `tick-stash.test.ts` (6 cases — stash/lookup, overwrite, per-room clear, all-clear, no-cross-contamination, freshness timestamp) and +2 cases on `route.test.ts` covering the new `stash-tick` action.

**95 plugin tests pass, `tsc --noEmit` clean.**

## Why this also fixes the second-order bug from #276

The reporter also called out that `formatTickInstruction` drops the `instruction`/`valid_keys` fields on error ticks, leaving recovering agents with a bare \"Action required: respond.\" Because the stash carries the **raw payload**, those fields are now available to the agent regardless of what the dispatched human-readable string includes — no `formatTickInstruction` change required.

## Verified live

Reinstalled the adapter, restarted the gateway, and ran a 2-agent cache-config negotiation against the running stack. Gateway log shows:

```
[mycelium-room] tick stashed: exp-7378-agent-a ← round 1 (exp-7378-after:session:1b9ab303)
[mycelium-room] 🎯 → exp-7378-agent-a
[mycelium] prependSystemContext: 2698 chars (cached), prependContext: 2049 chars (dynamic)
```

`prependContext: 2049 chars` — the same line that was `0 chars` on every turn before this change. The 2049-byte block is the full tick payload now in the agent's prompt.

## Test plan

- [x] Unit: `tick-stash.test.ts` (6) — stash/lookup, overwrite, per-room clear, all-clear, isolation, freshness
- [x] Unit: `route.test.ts` (+2) — emits `stash-tick` for session ticks; skips for parent-room ticks
- [x] Type: `tsc --noEmit` clean
- [x] Live: gateway log shows `tick stashed` and `prependContext: N chars (dynamic)` with N > 0 during real negotiation
- [ ] Run `before-and-after` skill end-to-end to confirm full negotiation completes (out of scope for this PR — fuzzy-snap already let it complete; this PR makes the agent reasoning better-grounded)

## Notes / follow-ups (not in this PR)

- Filed #278 — release.yml doesn't bump the plugin's `package.json` version on release, so `openclaw plugins list` shows `Version 0.1.0` even on tagged releases.
- During the live test, the two agents landed in **different** coordination session sub-rooms (`...:1b9ab303` and `...:31bd7676`) instead of sharing one. Unrelated to this fix, but worth filing separately.

### DIFF
--- a/mycelium-cli/src/mycelium/adapters/openclaw/mycelium/plugin/src/channel/index.ts
+++ b/mycelium-cli/src/mycelium/adapters/openclaw/mycelium/plugin/src/channel/index.ts
@@ -30,6 +30,7 @@ import {
   stopSessionSSE,
   subscribedSessionRooms,
 } from "./session-sse.js";
+import { clearAllTickStash, clearTickStashBySessionRoom, stashTick } from "./tick-stash.js";
 
 type Logger = { info: (s: string) => void; warn: (s: string) => void };
 
@@ -118,6 +119,7 @@ export function installChannel(
             startSessionSSE(runtime, cfg, name, _abort!, handleMessage, log);
           } else if (state && TERMINAL_STATES.has(state)) {
             stopSessionSSE(name, log);
+            clearTickStashBySessionRoom(name);
           }
         }
 
@@ -126,6 +128,7 @@ export function installChannel(
         for (const subbed of subscribedSessionRooms()) {
           if (!activeSessionRooms.has(subbed)) {
             stopSessionSSE(subbed, log);
+            clearTickStashBySessionRoom(subbed);
           }
         }
       } catch {
@@ -140,6 +143,7 @@ export function installChannel(
     _abort?.abort();
     _abort = null;
     clearSubscribedSessions();
+    clearAllTickStash();
     log.info(`[${CHANNEL_ID}] gateway stopping — SSE closed`);
   });
 }
@@ -222,6 +226,10 @@ function executeAction(
     }
     case "stash-return-address": {
       stashReturnAddress(action.sessionRoom, action.agentId, log);
+      return;
+    }
+    case "stash-tick": {
+      stashTick(action.agentId, action.sessionRoom, action.payload, log);
       return;
     }
     case "notify-home": {
--- a/mycelium-cli/src/mycelium/adapters/openclaw/mycelium/plugin/src/channel/route.ts
+++ b/mycelium-cli/src/mycelium/adapters/openclaw/mycelium/plugin/src/channel/route.ts
@@ -29,6 +29,12 @@ export type RouteAction =
   | { kind: "subscribe-session"; roomName: string }
   | { kind: "stash-return-address"; sessionRoom: string; agentId: string }
   | {
+      kind: "stash-tick";
+      sessionRoom: string;
+      agentId: string;
+      payload: any;
+    }
+  | {
       kind: "notify-home";
       sessionRoom: string;
       agentIds: string[];
@@ -111,6 +117,16 @@ export function routeTick(cfg: ChannelConfig, msg: any): RouteAction[] {
       kind: "stash-return-address",
       sessionRoom,
       agentId: targetAgent,
+    });
+    // Stash the raw tick payload so before_agent_start can inject it on the
+    // agent's next turn — the dispatched instruction below is a terse human
+    // summary that omits fields the agent needs (exact offer keys, valid_keys
+    // on error ticks, etc.).
+    actions.push({
+      kind: "stash-tick",
+      sessionRoom,
+      agentId: targetAgent,
+      payload,
     });
   }
 
--- a/mycelium-cli/src/mycelium/adapters/openclaw/mycelium/plugin/src/channel/tick-stash.ts
+++ b/mycelium-cli/src/mycelium/adapters/openclaw/mycelium/plugin/src/channel/tick-stash.ts
@@ -1,0 +1,74 @@
+// SPDX-License-Identifier: Apache-2.0
+// Copyright 2026 Julia Valenti
+
+/**
+ * Tick stash: latest coordination_tick payload observed per agent, keyed by
+ * agentId. Captured every time the channel SSE delivers a tick (overwrite —
+ * latest wins), used by `session/before_agent_start` to inject the full
+ * payload (current_offer keys, valid_keys on error ticks, etc.) into the
+ * agent's prompt.
+ *
+ * Why a stash and not an HTTP fetch?
+ *
+ * The channel module already holds the parsed tick when it dispatches the
+ * formatted instruction to the agent. Having `before_agent_start` re-query
+ * the backend by parent room (the prior design) doesn't work — ticks are
+ * attached to the CoordinationSession, not to any room. Rather than teach
+ * the session module to resolve coord-session display names and walk the
+ * `/api/coordination-sessions/{id}/messages` endpoint, we mirror the
+ * existing return-address pattern: channel observes once, stashes, session
+ * reads back.
+ *
+ * Keying:
+ *   - Return-address keys by (sessionRoom, agentId) because consensus
+ *     delivery has both in hand at lookup time.
+ *   - Tick stash keys by agentId alone — `before_agent_start` has the
+ *     agentId but not the session sub-room. If an agent ever participates
+ *     in two concurrent negotiations (not supported today), latest wins,
+ *     which is also what the prompt should reflect.
+ */
+
+type Logger = { info: (s: string) => void; warn: (s: string) => void };
+
+export type StashedTick = {
+  sessionRoom: string;
+  payload: any;
+  receivedAt: number;
+};
+
+const _stash = new Map<string, StashedTick>();
+
+export function stashTick(
+  agentId: string,
+  sessionRoom: string,
+  payload: any,
+  log: Logger,
+): void {
+  _stash.set(agentId, { sessionRoom, payload, receivedAt: Date.now() });
+  const round = payload?.round ?? "?";
+  log.info(`[mycelium-room] tick stashed: ${agentId} ← round ${round} (${sessionRoom})`);
+}
+
+export function lookupTick(agentId: string): StashedTick | undefined {
+  return _stash.get(agentId);
+}
+
+/**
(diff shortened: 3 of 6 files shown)


akanksha276 (approved):
looks good

PR #286, opened 2026-05-14
fix(coord): session join idempotency + collapse dual tick representations (#280, #284, #285)

### LINKED ISSUE #280: session join: concurrent joins race and fork the room into multiple coordination sessions
## Summary

Two simultaneous \`mycelium session join --room X\` calls (or \`POST /api/rooms/X/sessions\`) on the same room create **two separate** \`CoordinationSession\` rows for that room, each with one participant, instead of converging on a shared session. The agents then negotiate against themselves forever.

## Repro

While running the before-and-after smoke for PR #279, two haiku agents seeded simultaneously into the same room:

```
$ curl -sf http://localhost:8000/api/coordination-sessions?limit=20
3beaa911 complete    exp-7378-after:session:1b9ab303
    created_at: 2026-05-14T00:13:03.894477Z
    participants: []
3b2ec583 negotiating exp-7378-after:session:31bd7676
    created_at: 2026-05-14T00:13:03.893396Z
    participants: []
```

Both sessions, same room, **1.08 ms apart**. Each agent ended up alone in its own session sub-room. Gateway log:

```
[mycelium-room] tick stashed: exp-7378-agent-a ← round 1 (exp-7378-after:session:1b9ab303)
[mycelium-room] tick stashed: exp-7378-agent-b ← round 1 (exp-7378-after:session:31bd7676)
... agent-b proceeded to negotiate against itself through round 10 ...
```

## Root cause

\`fastapi-backend/app/routes/sessions.py:90\` — \`_spawn_coordination_session\`:

\`\`\`python
async def _spawn_coordination_session(parent_name: str, db: AsyncSession) -> CoordinationSession:
    result = await db.execute(
        select(CoordinationSession)
        .where(
            CoordinationSession.parent_room_name == parent_name,
            CoordinationSession.state.in_([\"idle\", \"waiting\", \"negotiating\"]),
        )
        .order_by(CoordinationSession.created_at.desc())
    )
    existing = result.scalar_one_or_none()
    if existing:
        return existing
    # ... INSERT new CoordinationSession
\`\`\`

Classic check-then-act. Both transactions SELECT, both find no existing non-terminal session, both INSERT, both commit. The DB has nothing to block this:

- No partial unique index on \`(parent_room_name)\` filtered to non-terminal states.
- No \`SELECT … FOR UPDATE\` on the parent room or on the existing-session lookup.
- No \`pg_advisory_xact_lock\` keyed by \`parent_room_name\`.
- \`CoordinationSession.short_id\` is random hex(8) so the two INSERTs get distinct PKs and both succeed.

See \`fastapi-backend/app/models.py:118-153\` — \`parent_room_name\` is indexed but not unique, no constraints on \`state\` at all.

## Why this is bad

- Joining the same room from two terminals or two agents simultaneously **silently** splits the negotiation. There is no error returned to either caller.
- The smaller window where this triggers is when \`/api/coordination-sessions/{id}/messages\` is the first call on that room — both joins arrive before the first one's INSERT commits.
- It compounds with #276's symptom (and PR #279's fix): now the tick payload is correctly injected, but the agent is happily negotiating against itself.
- The \`participants: []\` field on both rows in the example above is a separate (probably serialization-side) bug — \`Participant\` rows do get created and FK to the coord session, but the response model isn't loading them. Worth a follow-up.

## Suggested fix (defense in depth)

**Schema (preferred):** add a partial unique index that the DB enforces, so future code paths can't reintroduce the race.

\`\`\`python
# new alembic migration
op.create_index(
    \"ix_coordination_sessions_one_active_per_room\",
    \"coordination_sessions\",
    [\"parent_room_name\"],
    unique=True,
    postgresql_where=sa.text(\"state IN ('idle', 'waiting', 'negotiating')\"),
)
\`\`\`

**Code:** wrap the SELECT-then-INSERT in an advisory lock keyed by \`parent_room_name\` so we get a clean retry-free path even before the index lands, and a clean \`IntegrityError\` retry after it does:

\`\`\`python
async def _spawn_coordination_session(parent_name: str, db: AsyncSession) -> CoordinationSession:
    # Serialize concurrent spawns for the same room. The lock is held only
    # for the duration of this transaction (xact_lock variant).
    await db.execute(
        text(\"SELECT pg_advisory_xact_lock(hashtext(:k))\"),
        {\"k\": f\"coord_session_spawn:{parent_name}\"},
    )
    # ... existing SELECT-then-INSERT ...
\`\`\`

Either fix alone closes the bug; both together gives schema-level enforcement plus a clean code-side serialization.

## Test plan

- Repro: two concurrent \`POST /api/rooms/X/sessions\` from \`asyncio.gather\` in a unit test — assert only one CoordinationSession row exists with state ∈ {idle, waiting, negotiating}.
- Regression: an existing room with an already-negotiating session — concurrent joins must both return that session, not spawn a new one.

## Related

- Surfaced from the live e2e on PR #279 (tick-stash fix for #276). Filed separately because it's a backend bug; the plugin-side fix in #279 is unrelated and the right thing to do regardless.

### LINKED ISSUE #284: agent_count Double-Counted in Coordination Session
### Affected Version

1.0.11

### Description

Summary

When a 2-agent structured negotiation session starts, the coordination layer reports agent_count: 4. API-side join events and plugin-side session registrations are counted as separate participants, doubling the true agent count.   The root cause is sessions.py:168–174 — join_room() unconditionally inserts a new Participant row with no existence check, and the Participant table has no unique constraint on (coordination_session_id, agent_handle).

### Steps to Reproduce

Prerequisites: mycelium backend running, an openclaw gateway configured with two agents (bob,julie).
1. Create the room and start the session  mycelium room create exp-repro
  mycelium room use exp-repro

2. Register both agents via the CLI (the persona-before-and-after harness does this)  mycelium session join --handle bob -m "I prefer sequential workflow" -r exp-repro
  mycelium session join --handle julie -m "I prefer parallel workflow" -r exp-repro
  This creates 2 Participant rows in the database.

3. The OpenClaw channel plugin fires and dispatches agents into their sessions

  When the coordination_join messages arrive from the backend SSE, the openclaw plugin
  dispatches the SKILL.md instruction to each agent. The agents, following the SKILL.md
  negotiation protocol, execute step 1 of the protocol themselves:
  ```
# bob's agent runs:
  mycelium session join --handle bob -m "my position" -r exp-repro
  # julie's agent runs:
  mycelium session join --handle julie -m "my position" -r exp-repro
```
  This creates 2 more Participant rows — one per agent — bringing the total to 4.

4. Join window expires → coordination_start fires  The coordination_start message will contain "agent_count": 4 because coordination.py:480 computes it as len(session_handles) with no deduplication:

 ```
 # coordination.py:385 — no dedup
  session_handles = [s.agent_handle for s in sessions]
  # coordination.py:480
  content=json.dumps({"round": 0, "agent_count": len(session_handles)})
```

  Minimal repro (no OpenClaw required):
  ```
# Call join twice for the same agent — the endpoint always inserts, never upserts
  curl -X POST http://localhost:8000/rooms/exp-repro/sessions \
    -H "Content-Type: application/json" \
    -d '{"agent_handle": "bob", "intent": "test"}'

  curl -X POST http://localhost:8000/rooms/exp-repro/sessions \
    -H "Content-Type: application/json" \
    -d '{"agent_handle": "bob", "intent": "test"}'
```

  # agent_count in coordination_start will be 2, not 1

### Expected Behavior

agent_count should equal the number of unique agent identities participating in the negotiation — in this case, 2 (bob and julie). An agent that connects via both the API join path and the plugin session path should be deduplicated and counted once.

### Actual Behavior

agent_count: 4 is recorded at coordination start. Bob's API join + Bob's plugin session = 2 entries; same for Julie. The NegMAS engine sees 4 declared participants and computes quorum accordingly. With 4 declared participants, the threshold for consensus is higher than the actual 2 agents can satisfy, which likely caused the session to terminate as complete / outcome=none after only 3 rounds despite 17 rounds remaining in the budget.

### Environment

_No response_

### Troubleshooting Attempted

- [x] I ran `mycelium status` and reviewed the output
- [x] I checked the logs with `mycelium logs`
- [x] I used the troubleshooting skill or reviewed the troubleshooting guide

### LINKED ISSUE #285: Dual-Tick Representations Delivered Per Round (OpenClaw Path)
### Affected Version

1.0.11

### Description

Summary
  
On the OpenClaw adapter path, each agent receives two overlapping representations of the same coordination tick per round: a formatted human-readable instruction string dispatched as a channel message (via formatTickInstruction()), and the raw JSON payload injected into the agent's context window via before_agent_start. The two representations are inconsistent in content, not just format, and agents must reconcile them before acting — consuming negotiation budget without making progress.

Note: this bug was introduced by faf98b3 (merged in v1.0.11). The pre-v1.0.11 before_agent_start was trying to fetch ticks from the parent room over HTTP, which always returned nothing because ticks attach to the CoordinationSession, not the parent room. Agents only received the formatted string but were blind to exact offer key names and valid_keys on error ticks. The fix solved that problem by stashing the raw payload in memory and injecting it via before_agent_start — but in doing so created the dual-representation problem. Both are
now present in v1.0.11.

### Steps to Reproduce

  Prerequisites: mycelium backend running on v1.0.11, OpenClaw gateway configured with the
  mycelium channel plugin, at least one agent registered in a room.

  1. Configure the channel plugin with an agent
```
  # ~/.openclaw/openclaw.json — channels.mycelium-room section
  agents = ["bob"]
  room = "repro-room"
```

  2. Start a structured negotiation session with two agents
```
  mycelium room create repro-room
  mycelium session join --handle bob -m "prefer sequential" -r repro-room
  mycelium session join --handle julie -m "prefer parallel" -r repro-room
```

  3. Wait for the join window to expire — the first tick fires

When coordination_start is posted and round 1 begins, the OpenClaw channel plugin's SSE subscription receives the coordination_tick message. Two things happen simultaneously:

  - route.ts:100 builds the formatted string and route.ts:117–123 dispatches it as a channel message into bob's session
  - tick-stash.ts stashes the raw payload keyed by bob; on bob's next turn session/index.ts:181–186 reads the stash and prepends the full JSON as prependContext

  4. Observe bob's session context

  Bob's session now contains both:

  ```
# From channel dispatch (formatTickInstruction):
  [CognitiveEngine — Round 1 of 20]
  You are in a structured negotiation in room repro-room.
  Action required: propose
  You CAN propose a counter-offer.

  Current offer on the table:
    workflow: parallel
    ...
  To counter-propose, run: mycelium negotiate propose ... --handle bob

  # From before_agent_start prependContext (raw JSON stash):
  [Mycelium — coordination tick]
  Session: repro-room:session:abc123

  {
    "round": 1,
    "action": "propose",
    "can_counter_offer": true,
    "current_offer": { "workflow": "parallel", ... },
    "valid_keys": ["workflow", "handoff_signal", "authority"],
    ...
  }
```

  The raw JSON includes valid_keys; the formatted string does not. If the tick is an error tick
  (e.g. agent submitted an unrecognised key), the raw JSON carries correction data that the
  formatted string never surfaces — yet the agent sees both simultaneously with no signal
  indicating which to act on.

  5. Confirm the duplication is not present on the CLI path

  Run the same session with a CLI-path agent (mycelium session await) — it receives only the raw
   JSON tick payload via SSE. No formatted string, no duplication.

### Expected Behavior

Each agent should receive exactly one representation per tick. Per the CLAUDE.md contract, the OpenClaw path was designed so that formatTickInstruction() is the agent's sole source of tick information — complete enough that agents never need to see raw JSON. If raw JSON injection is the preferred direction going forward, formatTickInstruction() channel dispatch should be removed, not left alongside it.

### Actual Behavior

Two representations arrive per tick via different injection points:

  1. Channel message — route.ts:100 calls formatTickInstruction(payload, roomName, targetAgent)
  and dispatches the result as a conversational message from "CognitiveEngine" into the agent's
  session.
  2. before_agent_start context — session/index.ts:181–186 calls lookupTick(agentId) on the
  stash written by the channel module and prepends the full raw JSON as prependContext before
  the agent's turn begins.

The two representations are not equivalent. The formatted string includes action verbs, prior-round outcome text, and ready-to-run CLI commands. The raw JSON includes valid_keys on error ticks and exact issue key names that formatTickInstruction() deliberately omits. Agents receiving both must decide which is authoritative, and any inconsistency between them (e.g. on an error tick that carries valid_keys) causes the agent to spend turns resolving the ambiguity rather than negotiating.

### Environment

- Version: v1.0.11 (bug introduced by faf98b3)
- Adapter: OpenClaw only — CLI-path agents (mycelium session await) are unaffected, they
  receive raw JSON only
- Files:
    - mycelium-cli/src/mycelium/adapters/openclaw/mycelium/plugin/src/channel/route.ts:100, 
  117–123
    - mycelium-cli/src/mycelium/adapters/openclaw/mycelium/plugin/src/session/index.ts:181–186
    - mycelium-cli/src/mycelium/adapters/openclaw/mycelium/plugin/src/channel/tick-stash.ts

### Troubleshooting Attempted

- [x] I ran `mycelium status` and reviewed the output
- [x] I checked the logs with `mycelium logs`
- [x] I used the troubleshooting skill or reviewed the troubleshooting guide

juliarvalenti:
## Summary

Three related correctness fixes in the coordination layer, bundled because they share schema migrations and the same code path (`sessions.py` + plugin tick rendering).

- **#280** — `_spawn_coordination_session` had a SELECT-then-INSERT race with no DB constraint, so two simultaneous joins could fork the room into two CoordinationSessions.
- **#284** — `join_room` did an unconditional INSERT of a Participant row, and SKILL.md tells the agent to call `mycelium session join` itself on top of the harness/CLI doing it, so `agent_count` went to 2N for N agents and NegMAS quorum failed.
- **#285** — PR #279 shipped a hybrid that left both the formatted-string dispatch AND the raw-JSON `prependContext` injection active per tick, forcing agents to reconcile two representations of the same payload. The conversation on #279 had settled on Option B (single source); this commit ships that.

### Backend (#280, #284)

- **Migration `0016`** — two indexes:
  - Partial unique on `coordination_sessions(parent_room_name) WHERE state IN ('idle','waiting','negotiating')`.
  - Unique on `participants(coordination_session_id, agent_handle)`.
  - Both index creations are preceded by `DELETE … USING …` cleanups so legacy duplicates (from before this fix) don't block the index. Models declare the same constraints so `Base.metadata.create_all()` picks them up in tests too.
- **`_spawn_coordination_session`** holds a `pg_advisory_xact_lock(hashtext(parent_name))` for the SELECT-then-INSERT under PG. Wraps the INSERT in a `try/except IntegrityError` that re-fetches the winner if the lock was bypassed (no PG / legacy race).
- **`join_room`** does a SELECT-first fast path; the new INSERT runs inside `begin_nested()` so an IntegrityError doesn't poison the outer transaction (the join-window state machine still needs to run after). On dup-join, all side effects (coordination_join notify, KXP fan-in, join-window deadline extension) are skipped — they MUST run exactly once per real join.

### Plugin (#285)

- `formatTickInstruction` now renders **everything** the agent needs:
  - Normal ticks gain a `Valid offer keys (use exactly these in your counter): "Cache TTL", "Eviction Policy"` line so the LLM can't invent snake_case names.
  - New `formatErrorTick` branch handles `counter_offer_invalid_keys` / `counter_offer_not_your_turn` ticks, rendering the backend's `instruction` text, `valid_keys`, `bad_keys`, and a concrete recovery CLI command.
- Deletes:
  - `channel/tick-stash.ts`
  - `test/tick-stash.test.ts`
  - the `stash-tick` route action + executor case
  - the `lookupTick` + prependContext injection in `session/before_agent_start`
  - the `clearTickStashBySessionRoom` lifecycle calls
- The dispatched formatted string is now the agent's sole source of tick information per CLAUDE.md.

### Regression tests

Backend (`tests/test_session_spawn.py`, 5 new):

- `test_duplicate_join_same_handle_returns_one_participant` — sequential dup-join via API returns the same Participant id; DB has one row.
- `test_participant_unique_index_rejects_direct_duplicate_inserts` — direct ORM INSERT bypassing API is rejected by the unique index.
- `test_multiple_agents_join_same_session_no_fork` — two different handles → one shared CoordinationSession.
- `test_coord_session_partial_unique_index_rejects_direct_fork` — direct second-session INSERT for the same room is rejected by the partial index.
- `test_coord_session_unique_index_allows_post_terminal_replacement` — partial index leaves room for new sessions once the existing one is terminal.

Plugin (`test/route.test.ts`, 4 new + 1 replacement):

- `formatTickInstruction` includes `Valid offer keys` when `can_counter_offer=true`.
- `formatTickInstruction` omits the valid-keys line when not counter-eligible.
- Error tick: `counter_offer_invalid_keys` renders instruction + valid_keys + recovery.
- Error tick: `counter_offer_not_your_turn` renders accept/reject recovery.
- Replaced the previous "emits stash-tick" assertion with the negative form ("MUST NOT emit stash-tick") so a future revert of this PR fails the test.

## Verified

- 235 backend tests pass (`uv run pytest -x -q`).
- 92 plugin tests pass (`npm run test`, vitest).
- `ruff check` + `ruff format` + `npx tsc --noEmit` clean.

## Test plan

- [x] Backend: full suite green on SQLite (`tests/conftest.py` uses `sqlite+aiosqlite://`)
- [x] Plugin: vitest + tsc clean
- [ ] Manual: reinstall plugin (`mycelium adapter add openclaw --reinstall -y`), restart gateway, run a 2-agent negotiation; gateway log should show one `[mycelium-room] 🎯 → agent` per tick, NO `tick stashed` lines, NO `prependContext: N chars (dynamic)` from coordination context, and the dispatched string should include `Valid offer keys: …`.
- [ ] Manual: provoke an invalid-keys error tick (`mycelium negotiate propose wrong_key=val …`); agent receives the recovery command in chat with the valid keys enumerated.

## Migration safety

Migration 0016 contains `DELETE` statements to drop legacy duplicates BEFORE creating the unique indexes. The deletes keep the earliest-created row per duplicate group; cascading FKs handle downstream cleanup (participants tied to a dropped session, messages on that session). Anyone running against a DB where these races actually happened should confirm their data before applying.

Closes #280, closes #284, closes #285.


juliarvalenti:
## Live verification — `exp-8864` (after-only)

Reinstalled plugin + recreated backend container from this branch (mig 0016 applied at startup). Two haiku agents wired up, channel pointed at `exp-8864-after`, single seed.

| Fix | Signal | Status |
|---|---|---|
| #280 — session fork race | One CoordinationSession for the room, not two | ✅ `5f2bb918` |
| #284 — participant double-count | 2 handles, not 4 | ✅ `[agent-a, agent-b]` |
| #285a — tick-stash removed | Zero `tick stashed` log lines this run | ✅ 0 (13 historical, all from yesterday's exp-7378) |
| #285b — prependContext injection removed | `prependContext: 0 chars (dynamic)` every turn | ✅ |
| #285c — valid keys in formatted string | `Valid offer keys (use exactly these...)` in proposer prompt | ✅ all 5 CE-extracted keys |

**Coordination behavior:** protocol works (proposer rotation, no invalid_keys errors, well-formed ticks). The negotiation does not converge — agent-a's hard constraint (TTL ≤ 180s) and agent-b's hard constraint (TTL ≥ 300s) leave no zone of agreement. Agents loop around 150-180s offers. That's a *correct* negative outcome — there is no agreement to be had — but the agents haven't yet identified "walk away" as the right move. LLM-reasoning observation, not a protocol bug.

Full artifacts (verification summary + session transcript + both agents' rendered tick instructions + gateway log slice):
https://gist.github.com/juliarvalenti/8bf3dd970e1115e880e88f1a144af319

Issue #297, opened 2026-05-20
metrics: per-room cost breakdown for CFN engines + Mycelium synthesis (currently only OpenClaw)


sfph:
## Context

`mycelium metrics show cost` already breaks out **OpenClaw** spend per room as sub-rows under the agent total:

```
Cost Estimates
  OpenClaw Agents      14,205   $0.42   otel (provider-reported)
    matrix-room         9,884   $0.31   est. (local agents only)
    dev-room            4,321   $0.11   est. (local agents only)
```

But the **CFN Engines** and **Mycelium Backend** rows in the same table are top-line only — no per-room breakdown — even though the underlying counters (`cfn_llm.by_room.*`, soon `mycelium.llm.synthesis.by_room.*` if #296 lands) carry the same shape.

That makes it hard to answer "which room is eating my CFN budget?"

## Proposal

Extend per-room sub-row rendering in `_render_cost_estimates` to cover all three sources:

| Source | Token data by room | Cost per room |
|---|---|---|
| OpenClaw Agents | ✅ exists (otel sessions, agent→room map) | ✅ exists (estimated from configured CFN model) |
| **CFN Engines** | ✅ exists (`cfn_llm.by_room.*`) | ❌ missing — estimate from `pricing.json` × tokens |
| **Mycelium Backend** | ❌ missing — no by-room split on `llm.calls` today | ❌ missing — gated on #296 |

Implementation sketch:

1. Reuse the `_parent_room()` aggregation helper (also being added in #295) to bucket session-scoped CFN keys.
2. For CFN row: when `cfn_llm.by_room.*` totals > 0, emit indented sub-rows with `_estimate_cost(...)` per room using the CFN-configured model. Label as `est. (engine pricing)`.
3. For Mycelium row: gated on #296 — once synthesis tokens are bucketed by room, render the same way.
4. Keep the `[:5]` + `…and N more` cap pattern that the OpenClaw section already uses.

## UX example after fix

```
Cost Estimates
  OpenClaw Agents      14,205   $0.42   otel (provider-reported)
    matrix-room         9,884   $0.31   est. (local agents only)
    dev-room            4,321   $0.11   est. (local agents only)
  CFN Engines          21,608   $0.18   est. (pricing.json)
    matrix-room        15,002   $0.13   est. (engine pricing)
    dev-room            6,606   $0.05   est. (engine pricing)
  Mycelium Backend        842   $0.01   est. (pricing.json)
[...]

Issue #296, opened 2026-05-20
metrics: `mycelium.llm.calls` is misleading — only synthesis is instrumented


sfph:
## Context

`mycelium metrics show mycelium` reports a single counter for backend LLM activity:

```
Mycelium Backend LLM Usage
  Total LLM calls    1
```

This is technically accurate for the underlying counter but **misleading at the label level** — it implies the backend made 1 LLM call total. In reality the counter is only incremented by the **room-synthesis** codepath. Every other LLM-touching site in the backend goes uncounted.

## Inventory

Direct `litellm.(a)completion` invocations in `fastapi-backend/app/`:

| Site | Counted? | Notes |
|---|---|---|
| `services/async_coordination.py:322` (synthesis) | ✅ yes | success + AuthError + generic Exception all wrap with `record_llm_call(operation="synthesis", ...)` |
| `services/llm_health.py:439` (health probe, max_tokens=1) | ❌ no | `acompletion` for `/api/health` Level C |
| CFN-side LLM calls (negotiation, recall, etc.) | n/a | tracked under `cfn.llm.*`, correctly separate |
| OpenClaw-agent LLM calls | n/a | tracked agent-side via OTel |
| Local embeddings (fastembed/ONNX) | n/a | tracked under `embedding.*`, correctly separate |

So the *only* `mycelium.llm.calls` site is synthesis.

## Why this matters

During a typical e2e run a user sees:

- `mycelium.llm.calls = 1`
- `cfn.llm.calls = N` (where N is large)

…and reasonably asks "why is mycelium making only 1 call?" The answer is "it isn't — the counter is partial." Either the label should match the reality, or the instrumentation should match the label.

## Proposal

Two-part fix, pick one or both:

1. **Rename for accuracy** (low-risk, ~no code outside metrics):
   - Rename the counter to `mycelium.llm.synthesis.calls` (and `synthesis.input_tokens`, etc.).
   - Update `_render_mycelium_llm_usage_table` to display "Synthesis LLM calls" instead of "Total LLM calls".

2. **Broaden instrumentation** (medium):
   - Add a thin wrapper helper (e.g. `record_llm_call(operation=...)` already supports this) at the other backend completion sites, starting with `llm_health.probe_completion`.
   - Optionally add `kind=` label distinguishing `synthesis` / `health` / future kinds so the existing total stays meaningful.

## Out of scope

- Surfacing CFN or OpenClaw counts under the mycelium-backend rollup (they already have their own sections).
[...]

Issue #295, opened 2026-05-20
metrics: CFN LLM "By room" doesn't aggregate session sub-rooms


sfph:
## Context

`mycelium metrics show cfn` renders a per-room rollup under **CFN LLM Token Usage → By room**, but every session shows up as its own line — all labelled identically as `mycelium_room`:

```
By room:
  mycelium_room    7 calls, 14,205 in / 1,022 out
  mycelium_room    3 calls,  6,118 in /   401 out
  mycelium_room    9 calls, 19,884 in / 1,733 out
  ...
```

The backend stores these under session-scoped counter keys (`by_room.<room>:session:<uuid>.<metric>`). The renderer strips `:session:<uuid>` from the **display label** but still emits one row per raw key, producing N duplicates instead of a rolled-up total.

## Where

`mycelium-cli/src/mycelium/commands/metrics.py:2974-2996` in `_render_cfn_llm_usage_table`:

```python
for room, data in sorted(rooms.items()):
    ...
    label = room
    if ":session:" in label:
        label = label.split(":session:", 1)[0]
    table.add_row(f"  {label}", ...)
```

The bucketing happens *before* the strip — so rooms keyed `mycelium_room:session:A` and `mycelium_room:session:B` are two distinct entries in `rooms` that just get the same label.

## Fix

The same file already has the right pattern in `_render_coordination_table` (lines 2842–2850), which uses a `_parent_room()` helper to bucket *first*, then render:

```python
def _parent_room(label: str) -> str:
    return label.split(":session:", 1)[0] if ":session:" in label else label

rounds_by_room: dict[str, int] = {}
for key, val in coord.items():
    if not key.startswith("by_room."):
        continue
    room = _parent_room(key.removeprefix("by_room."))
    rounds_by_room[room] = rounds_by_room.get(room, 0) + int(val or 0)
```

Apply the same pattern in `_render_cfn_llm_usage_table` so the per-metric dict is keyed by parent room, and metrics are summed across sessions before render. Optionally cap with `[:5]` + `…and N more` like the coordination table does.

## Out of scope

- Persisting per-session breakdown (we still want the raw session keys in the counter store for debugging).
- Backend-side aggregation — the data shape is fine, it's purely a renderer bug.

## References

- `mycelium-cli/src/mycelium/commands/metrics.py:2842` — `_parent_room` helper (reuse this)
- `mycelium-cli/src/mycelium/commands/metrics.py:2974` — buggy `by_room` loop in CFN LLM table

[...]

PR #300, opened 2026-05-21
feat(metrics): containerize collector + per-host architecture + traces UI + cross-host fixes

sfph:
## Summary

This branch turns the OTLP metrics collector into a first-class container, builds an explicit hub/spoke architecture on top of it, adds a trace explorer to the CLI, and fixes a class of cross-host aggregation bugs that were silently undercounting cluster totals. It also picks up #292 (plan/todo namespace), #295/#296/#297 (CFN LLM counter fixes), and a number of correctness fixes the collector branch surfaced along the way.

74 commits, 201 files, +7932/-1923.

### Collector lifecycle
- New `Dockerfile.collector` and `mycelium-collector:dev` compose service; published to GHCR on release (`ci.yml` smoke-builds it on PRs).
- New `.dockerignore` drops the image from **988 MB → 306 MB** by excluding 543 MB of plugin `node_modules` (and the local `.venv`, `build/`, etc.) from the build context. The setuptools wheel already excluded them — the Docker COPY layer didn't.
- `metrics stop` command + daemonized spoke collector.

### Hub/spoke OTLP architecture
- Spoke collectors transparently forward every `/v1/metrics` and `/v1/traces` payload to the hub via fire-and-forget POSTs (agent-to-gateway pattern), so the hub gets the unified cross-host view without each spoke having to ship OTLP straight to the hub.
- Per-host bookkeeping end-to-end (`Spoke Sites` → renamed `Sites` with hub row tagged) across collector, CLI, backend, and UI.
- `--host` filter on `mycelium metrics show` for per-spoke slicing; exits cleanly when the host is unknown.

### Cross-host counter aggregation (correctness)
- `MetricsStore` was storing every host's cumulative-counter pushes in a single shared dict and `set`-ing rather than summing. Whichever host pushed last clobbered the others, so the hub's `Tokens by Channel` and `Cost Estimates` panels silently reported the value from one (essentially random) spoke instead of the cluster total.
- Fixed by per-host bucketing (`counters_by_host` / `histograms_by_host`) with sum-on-snapshot aggregation; persisted across collector restarts so reloads don't re-aggregate (which would double-count on the next push).
- 11 regression tests in `tests/test_collector_aggregation.py` cover tokens, cost, message/lane/session-state counters, and histogram merging.

### Metrics CLI / UI polish
- `mycelium metrics traces` viewer over the collector's local `~/.mycelium/metrics/traces.db` (full table or per-trace event timeline).
- Per-room slicing for coordination, knowledge, and cost panels (#297, #295, plus this branch's continued work).
- Added `MAS id` column to per-room sub-tables so rooms with identical display names but different MAS ids are distinguishable.
- New `--detail` flag expands the truncated "By room" sections and shows full mas_id UUIDs instead of the 8-char prefix.
- Room `mas_id ↔ room_name` mapping is now captured in the backend metrics snapshot at write time so the per-room tables survive hard room deletion (previously the mapping vanished as soon as `/api/rooms/{name}` was DELETEd).

### Backend
- #292 plan/todo namespace (merged in).
- CFN LLM call instrumentation completed (#296), with per-room cost breakdown (#297) and aggregated "By room" across sessions (#295).
- `leave_room` route is now scoped to the correct `CoordinationSession.parent_room_name` — previously a caller who knew a participant UUID could remove that participant by addressing DELETE to any room URL. Regression tests added.
- `_post_message` no longer crashes on FK violation when the room was deleted mid-flow.
- Spawn race fix (defense-in-depth with the existing partial unique index).
- `llm_health.probe_completion` collapsed to a single `finally`-block error recording path so any future except clause can't drift into silent under-counting.

### Other
- Sanitize `Infinity`/`NaN` from collector JSON output.
- Adapter fixes: pin OpenClaw OTLP endpoint to localhost (don't leak the hub URL), tolerate `git pull` failure for deep-observability, auto-configure diagnostics-otel when `--step=deep-observability` is used.
- ty/ruff/format: clean (`pnpm run lint` passes in both `mycelium-cli` and `fastapi-backend`).

## Test plan

- [x] `uv run pytest -q` in `mycelium-cli` → **83 passing** (incl. 11 new collector aggregation tests).
- [x] `uv run pytest -q` in `fastapi-backend` → **254 passing, 12 skipped** (incl. 3 new leave_room tests).
- [x] `pnpm run lint` in `mycelium-cli` and `fastapi-backend` (ruff check + ruff format + ty check) → clean.
- [x] CI green on PR: backend tests, CLI lint, collector docker build, frontend typecheck+build, OpenAPI snapshot staleness, OpenClaw plugin tests — all 6 checks pass.
- [x] Hub: `mycelium metrics show` populates `counters_by_host` with all 3 hosts (`oclw4`, `oclw-3`, `oclw5`) and the aggregated `tokens.by_agent` view sums across hosts instead of clobbering.
- [x] Spoke: `mycelium metrics show` on oclw5 renders the overview with backend-from-hub fallback; `--host nonexistent-host` exits cleanly with "No data found for host 'nonexistent-host'." and the list of known hosts.
- [x] `mycelium metrics traces` renders the recent-trace summary; `mycelium metrics traces list` shows per-span table; `mycelium metrics traces show <trace_id>` renders the parent→child tree.
- [ ] e2e suite (`mycelium-e2e-test`) — running now; results will be posted in a comment.


sfph:
## E2E suite results

Full suite ran 1h 54m: **36 passed / 6 failed** (86%).

| Test | Root cause | Caused by this PR? |
|---|---|---|
| `test_06d_cfn_llm_counters` | 90s budget too short for node-svc 0.1.6's ~125s `start_negotiation` (new `generate_options_with_memory` self-call hangs the full upstream 120s timeout before falling back to LLM-only) | No — pre-existing |
| `test_08_ioc_cfn` | 30s ingest timeout vs ~125s queueing behind peer CFN load (single-worker node-svc) | No — pre-existing |
| `test_10_ioc_negotiation_path` | 120s test timeout loses race to ~125s CFN response by ~5s every time | No — pre-existing |
| `test_17_resource_allocation` | Same upstream `generate_options_with_memory` 120s hang — `coordination_tick received` timed out at 240s | No — pre-existing |
| `test_49_skill_cross_channel_return_trip` | Multi-device Matrix consensus failed at 10 min — real agents on oclw3/4/5 not converging | No — unrelated, real-agent flow |
| `test_60_cross_channel_memory_isolation` | Multiple "no response from agent-X within 90-120s" — Matrix agents stalled | No — unrelated, real-agent flow |

**None of the 6 failures are caused by anything in this PR's diff.** They all trace to either:
- An upstream regression in `ioc-cognition-fabric-node-svc:0.1.6` (4 of 6), or
- Pre-existing real-agent timing fragility on cross-channel Matrix flows (2 of 6).

**Follow-up on `mycelium-e2e-test`**: pushed [`ee76791`](https://github.com/sfph/mycelium-e2e-test/commit/ee76791) (`fix(e2e): react to CFN node-svc 0.1.6 behavior + harden 06d/08/10 timing`) which:
- Updates the CFN-svc health probe URL to `/api/internal/diagnostics/health` (node-svc 0.1.6 dropped public `/health`).
- Bumps `test_10` POST timeout 120s → 240s.
- Bumps `test_08` HTTP timeouts 30s → 180s.
- Rewrites `test_06d` with a two-phase synchronization (Phase 1: `coordination_start` posted; Phase 2: `cfn_llm.calls` advanced) and diagnostic-rich failure messages so the next investigation distinguishes "backend tick didn't fire" from "CFN didn't return usage" without re-reproducing.

Should also file a separate upstream ticket for the `generate_options_with_memory: fabric lookup failed (http_status=None)` self-call hang in node-svc 0.1.6 — that's the underlying issue affecting 4 of the 6 failing tests.

PR #307, opened 2026-05-21
config: faithful ~/.mycelium/.env materialisation — DB URL, image tag, ports (+ fix #287)

### LINKED ISSUE #287: mycelium doctor: `_check_pending_migrations` mis-reports state in three common configurations
## Summary

`mycelium doctor`'s pending-migrations check has three latent failure
modes that combine to make it unreliable on every install topology
except "developer running from a clone with `DATABASE_URL` already
exported". Hit all three today on `feat/metrics-containerization`.

### Mode 1 — silently skipped on installed CLI even when a hub is running locally

```
$ which mycelium
/home/ubuntu/.local/bin/mycelium
$ mycelium doctor | grep -i migration
  ✓ Migrations            Skipped (fastapi-backend not found — installed mode)
```

Backend is running in Docker, alembic is at head, but the check skips
because it can't find `fastapi-backend/` as a sibling of the *installed
package dir* (`~/.local/share/uv/tools/mycelium-cli/lib/...`). The
hop-up loop at `doctor.py:788` only walks parents of the installed
package — for end users on PyPI/Homebrew this can never find a backend
dir. Result: a working hub silently advertises "skipped" instead of
either confirming `head` or warning if behind.

### Mode 2 — phantom "pending migration" parsed out of a Python traceback

When the check **does** find `fastapi-backend/` (e.g. via an editable
install hit by the e2e venv), it shells out to `uv run alembic current`
*without* setting `DATABASE_URL`. `env.py` raises
`ValueError("DATABASE_URL environment variable is not set!")`. The
check then runs this regex on combined stdout+stderr:

```python
rev_match = _re.search(r"([0-9a-f]{4,})", output)
```

…and dutifully matches `aceback` from `Traceback (most recent call
last):`. The user sees:

```
~ Migrations: DB at acebac, latest is 0016 — pending migrations
```

There is no revision called `acebac` and the DB is actually at head.
Reproducible via:

```
$ cd fastapi-backend && unset DATABASE_URL
$ uv run alembic current 2>&1 | head -c 30 | od -c | head -1
0000000   T   r   a   c   e   b   a   c   k       (   m   o   s   t
$ python3 -c "import re; print(re.search(r'[0-9a-f]{4,}', 'Traceback'))"
<re.Match object; span=(2, 8), match='aceback'>
```

### Mode 3 — check requires `DATABASE_URL` from env even when `MyceliumConfig` has it

`doctor.py:824` loads `MyceliumConfig` and tries to pull
`cfg.server.database_url`. On a hub install that field is empty (the
backend reads its URL from the container's env, not from
`~/.mycelium/config.toml`), so the subprocess gets no `DATABASE_URL`
unless the operator has already exported one in the shell. End users
who just want to run `mycelium doctor` to verify their stack hit Mode 2
above.

Confirmed today:

```
$ mycelium doctor              # Mode 1: skipped (installed-mode CLI)
$ export DATABASE_URL=postgresql+asyncpg://postgres:password@localhost:5432/mycelium
$ mycelium migrate              # works
$ mycelium doctor               # now reports: ✓ DB at head (0016)
```

The migrate command succeeds because it can connect; doctor only
*reports* correctly once the URL is in the operator's shell, which it
shouldn't need to be.

## Suggested fix

Three independent patches; each stands on its own:

1. **Fail loudly on subprocess error** — guard the regex match behind
   `result.returncode == 0`, and only then look for revision tokens.
   On non-zero exit, emit a clear "alembic current failed: <stderr
   first line>" warning so phantom revisions stop showing up.

2. **Tighten the revision regex** — match only at start of line and
   require word boundaries, e.g.
   `re.search(r'^([0-9a-f]{8,})\b', output, re.MULTILINE)`. Alembic
   revisions in this repo are 4-digit numeric prefixes (`0016_…`) on
   disk, but the DB-side revision IDs alembic emits are 12-char hex.
   `Traceback`, `lalembic`, package paths under
   `.cache/uv/abcdef…/`, and most other accidental hex substrings stop
   matching.

3. **Query the backend instead of running alembic locally** — add a
   read-only `/api/admin/migration-revision` (or extend `/health` with
   a `database.revision` field) and have doctor prefer that path when a
   reachable backend is available. This is the only path that gives
   end-user installs (PyPI / Homebrew / install.sh) a meaningful answer
   without requiring a clone or DB URL in the operator's shell.

Mode 1's "Skipped (installed mode)" message stays correct for the
no-backend-reachable case, but with #3 it becomes the fallback rather
than the default.

## Test plan

- Unit test: feed the current alembic-traceback string to the parsing
  helper and assert it returns "could not determine" rather than a
  phantom revision.
- Unit test: feed `0016 (head)` and assert OK.
- Integration: run `mycelium doctor` on a hub with `DATABASE_URL` unset
  in the shell — should still report head via the backend endpoint
  (post-#3).

Filed while validating PRs #273, #286 against
`feat/metrics-containerization`. Found via the e2e
`test_06c_doctor_clean` regression test which parsed
`mycelium --json doctor` and surfaced the phantom warning end-to-end.

sfph:
## Summary

Make `~/.mycelium/.env` a **faithful, complete materialisation of `config.toml`** instead of a partial translation. The Postgres URL was the loudest example of `.env` lying about what `config.toml` said (and the root cause of #287's `mycelium doctor` "DB at acebac, latest is 0016 — pending migrations" false-positive), but the same pattern bit image tag, custom passwords, and the metrics port. Every fix in this PR converges on one rule: anything a user sets via `mycelium config set` must reappear in `.env` exactly, and every consumer reads through one helper rather than reassembling.

As a direct consequence, fixes the `mycelium doctor` "DB at acebac, latest is 0016 — pending migrations" false-positive on fully-migrated databases (#287).

## DB URL SSOT (original scope)

Pre-change, three places independently assembled the URL:
1. `compose.yml` — inline `postgresql+asyncpg://postgres:${MYCELIUM_DB_PASSWORD}@mycelium-db:5432/mycelium`
2. `mycelium doctor` — read the never-populated `cfg.server.database_url` field
3. `mycelium migrate` / `_run_migrations` — hoped `DATABASE_URL` happened to be in the user's `.env`

Post-change, the recipe lives on `MyceliumConfig.database_url()` and **every consumer reads through it**:

| Consumer | Mechanism |
|---|---|
| Backend container | `${DATABASE_URL}` materialised into `.env` from the helper |
| Graph indexer | `${GRAPH_DB_URL}` materialised into `.env` from the helper |
| `mycelium doctor` | calls `cfg.database_url(host_side=True)` directly |
| `mycelium migrate` / install-time migrations | prefer `${DATABASE_URL_HOST}` from `.env`, fall back to helper |
| Alembic `env.py` | unchanged — still reads `os.getenv("DATABASE_URL")` |

`config.toml` stays the **single editable source**; `.env` is purely a materialised translation layer for compose's benefit (and now also for host-side tools). The `server.database_url` override becomes a real escape hatch for external/managed Postgres — when set, every variant returns it verbatim.

## Why this is SSOT-compliant

- Only **one file is hand-edited** (`config.toml`)
- `.env` is auto-generated, banner-stamped `# Auto-generated from ~/.mycelium/config.toml — do not edit manually.`
- The URL assembly recipe lives in **exactly one method** (`MyceliumConfig.database_url`)
- The host vs container hostname split becomes **explicit and modeled** instead of a hidden assumption baked into `compose.yml`

## Other `.env` materialisation fixes folded in

Same theme as DB URL — `config.toml` says X, `.env` must reflect X, every consumer reads `.env`:

- **Image tag round-trips through `config apply`.** Previously `MYCELIUM_IMAGE_TAG` in `.env` got clobbered back to the default on every `config apply`, so users pinning a preview build silently lost the pin. Now the helper preserves it.
- **`mycelium up` announces the resolved image tag.** The `:latest` fallback was invisible — operators couldn't tell if the stack was running a pinned tag or had silently fallen back. The post-start summary now prints `mycelium-backend  → ghcr.io/…:<resolved-tag>` so the running version is observable.
- **`MYCELIUM_METRICS_PORT` materialised into `.env`.** `compose.yml` already published the collector via `${MYCELIUM_METRICS_PORT:-4318}:4318` and `runtime.collector_port` was a first-class config field, but `generate_env_file()` never wrote the env var — so a custom `collector_port` silently fell back to 4318 in both the host-side port mapping and the `mycelium up` summary. Now backend/UI/metrics ports all materialise consistently from the same helper.
- **Custom passwords survive compose substitution.** Passwords with special characters (`@`, `:`, `/`) broke the inline URL assembly in `compose.yml`. Fixed by URL-encoding inside the helper and switching `compose.yml` to nested `${VAR:-fallback}` substitution so custom passwords flow through unchanged. Factored `resolve_host_database_url` out so host-side tools share the same fallback chain.

## Docs

Metrics setup had three real gaps the metrics-port fix exposed — none caught users at install time, but all silently produced "no data" outcomes:

- **`docs/openclaw-matrix-setup.md` §9** told users to enable the OTLP plugin (`mycelium adapter add openclaw --step=otel`) but never to start the collector. OpenClaw happily pushed OTLP frames into a port nothing was listening on. Added `mycelium up --metrics` to the setup snippet with rationale + verify block.
- **`hub-and-spoke.md` Step 4** documented the spoke side in isolation, leaving readers to discover that the hub also needs `mycelium up --metrics` running. Added a prerequisite callout and a "How the spoke collector works" sub-block explaining the local-first-then-forward design (no docker assumption on spokes; local survives a hub outage; forwarding is fire-and-forget so hub-side failures surface as gaps in the Spoke Sites table, not spoke-side errors).
- **`troubleshooting.md`** env-var table was missing `MYCELIUM_UI_PORT` and `MYCELIUM_METRICS_PORT`; "Port Already in Use" snippet only showed backend/db. Replaced the hand-edit snippet with the canonical `config set` → `config apply` → restart flow covering all four runtime port keys.

`docs/reference.html` and `docs/index.html` regenerated via `docs/generate_docs.py`; `runtime.frontend_port` also picked up in the `[runtime]` config block (latent post-merge stale-doc).

## Closes #287

Two independent bugs combined to produce the pre-fix warning on a fully-migrated DB:

1. Doctor read `cfg.server.database_url` (never populated by install) → alembic invoked without `DATABASE_URL` → `env.py` raised `ValueError` → traceback in subprocess output
2. The output (stdout + stderr concatenated) was scanned with `/[0-9a-f]{4,}/` → "Traceback" matched `aceback` → reported as the current DB revision

Three end-to-end fixes:
- Source `DATABASE_URL` from the helper (root cause)
- Parse alembic stdout *separately* from stderr (alembic emits INFO logs and tracebacks on stderr)
- Tighten the regex with start-of-line anchor + word boundary (belt-and-suspenders)

The pure parse logic is factored into `_parse_alembic_current()` so it can be tested without a real database.

## Commits

1. `config: add MyceliumConfig.database_url() — single source for the Postgres URL`
2. `docker: materialise DATABASE_URL into .env, drop inline assembly from compose`
3. `migrate: route host-side alembic through DATABASE_URL_HOST + config helper`
4. `doctor: fix migration false-positive — "DB at acebac, latest is 0016" (#287)`
5. `config: round-trip MYCELIUM_IMAGE_TAG across mycelium config apply`
6. `up: announce resolved image tag so :latest fallback is visible`
7. `docker: use nested substitution in compose fallbacks for custom passwords`
8. `config: URL-encode passwords, extract resolve_host_database_url helper`
9. `fix(config): materialise MYCELIUM_METRICS_PORT into .env`
10. `docs: fix metrics-setup gaps for all-in-one and hub-and-spoke`
11. `docs(hub-spoke): clarify --metrics value on the hub itself`
12. `docs(hub-spoke): expand "How the spoke collector works" rationale`

(Two `Merge main` commits along the way to keep diff size sane against the moving target.)

## Tests

181 tests pass on the full mycelium-cli suite. New / updated coverage in this PR:

- `test_config_database_url.py` (12) — container/host/override variants, password injection, custom port, async-vs-plain driver, ClassVar leak guard
- `test_docker_env_materialization.py` (11) — `.env` carries all three URL variants, picks up custom password, honours remapped port, propagates override, **materialises all three service ports** (backend/UI/metrics), defaults `MYCELIUM_METRICS_PORT` to 4318, `resolve_host_database_url` tier fallback
- `test_doctor_migrations.py` (12) — real alembic outputs (`0016 (head)`, hex hashes, branch arrows, branch labels), **the exact #287 traceback regression**, log-line rejection, edge cases (empty, whitespace, leading indent, three-char tokens, uppercase hex)
- `test_instance_announce_image_tag.py` — `mycelium up` post-start summary prints the resolved tag
- `test_refresh_templates.py` — `compose.yml` refresh on upgrade preserves user customisations

End-to-end verified on the live hub:
- `mycelium config apply` → `.env` now carries `DATABASE_URL`, `GRAPH_DB_URL`, `DATABASE_URL_HOST`, `MYCELIUM_METRICS_PORT`
- `mycelium up --metrics` → backend container env contains the three URL variants; collector container publishes the configured port; post-start summary prints the resolved image tag
- Backend `/health` → DB connected, model loaded
- `mycelium doctor` (from inside the repo) → `Migrations DB at head (0016)` (was: `DB at acebac, latest is 0016 — pending migrations`)
- Direct test of `_parse_alembic_current` against the exact pre-fix traceback → returns `None` (was: `acebac`)
- **30-series mycelium-e2e-tests** (distributed coordination across the 3-spoke fleet) → pass

## Test plan

- [x] Unit tests pass (181/181 green)
- [x] `ruff check` clean
- [x] `ruff format --check` clean
- [x] `ty check` clean
- [x] End-to-end on hub: regenerate `.env`, restart backend, run doctor → migration check now correct
- [x] Direct regression test against the exact #287 traceback
- [x] Spoke verification (30-series e2e against the live 3-spoke hub passed; collector-only nodes confirmed unaffected by DB-side changes)
- [ ] Reviewer sanity-check on commit boundaries / messages
- [ ] CI green on PR

## Backward compatibility

- Upgraded installs whose `.env` was generated before this change still boot — `compose.yml` keeps literal fallback values for the `DATABASE_URL` / `GRAPH_DB_URL` env entries that match the legacy defaults.
- Running `mycelium config apply` once regenerates `.env` with the canonical URLs, image tag, and the full set of port env vars; the compose fallbacks become inert.
- Users with an explicit external `DATABASE_URL` in `fastapi-backend/.env` continue to work (the migrate-side fallback chain honours it as the last resort).
- `server.database_url` was previously a dead config field; it now functions as the documented escape hatch for external Postgres.
- Pre-fix `.env` files that lack `MYCELIUM_METRICS_PORT` still resolve via the `${MYCELIUM_METRICS_PORT:-4318}` fallback in `compose.yml` — `mycelium config apply` writes the explicit value on next regeneration.


juliarvalenti (approved):
LGTM. Solid SSOT cleanup — the database_url() helper is genuinely the only assembly site, and the #287 fix (separate streams + tightened regex + helper-sourced URL) is correct end-to-end with a literal regression test against the exact traceback. Merging as squash to match the repo convention.

One follow-up worth filing: `resolve_host_database_url` in docker_utils silently overrides a pre-existing `DATABASE_URL` in .env. Old behavior was to honor a user-pasted external URL; new behavior clobbers it (only `DATABASE_URL_HOST` survives). Not blocking — but worth a release-note callout or a quick precedence check on env["DATABASE_URL"] before falling through. Will file as a separate issue.

PR #377, opened 2026-06-23
fix(openclaw): un-sandbox created agents so they can exec the mycelium CLI

### LINKED ISSUE #376: agent create --adapter openclaw allowlists mycelium CLI but leaves agent sandboxed, so the CLI is 'not found'
## Summary

`mycelium agent create --adapter openclaw` reports `allowlisted mycelium CLI for <agent>`, but the newly-created agent inherits the default Docker sandbox (`agents.defaults.sandbox.mode = "all"`, `workspaceAccess: "none"`). Inside that sandbox the `mycelium` binary doesn't exist, so the agent gets:

```
/bin/sh: 1: mycelium: not found
```

The agent can still participate in a room via the `mycelium-room` channel (the channel does the SSE I/O on its behalf), but it **cannot run any `mycelium` command itself** — `session join`, `message propose`, `message respond`, etc. all fail. That defeats the point of allowlisting the CLI for it.

## Root cause

Two independent layers that aren't aligned at create time:

1. **Allowlisting** (what `agent create` does): adds `mycelium` to the gateway's allowed-commands list so it isn't blocked as a "dangerous command" (see `integrations/openclaw/dispatch.py:111`).
2. **Sandbox presence**: with `sandbox.mode = "all"`, the agent executes in `openclaw-sandbox-python:latest`, which has neither the host-installed `mycelium` binary (`~/.local/bin/mycelium`) nor access to `~/.mycelium` (`workspaceAccess: "none"`).

Allowlisting a command is meaningless if the binary isn't present in the execution environment. `agent create` does (1) but not (2), so an openclaw agent is broken by default for CLI use.

## Current behavior

We already **detect** this in `commands/doctor.py:_check_openclaw_agent_sandbox()` (`doctor.py:1308`), which warns:

> `N channel agent(s) are sandboxed — mycelium CLI invisible`
> fix: set `sandbox.mode = 'off'` in openclaw.json for each agent, restart gateway

But the warning only surfaces if you run `mycelium doctor` after the fact. Create-time gives no hint, so the failure shows up much later as a confusing `not found` inside the agent.

## Repro

1. `mycelium agent create devops --adapter openclaw --room <room>` (inherits default sandbox)
2. Invoke it and have it try to run any `mycelium ...` command
3. Agent reports `mycelium: not found`

## Suggested fix

Pick one (in rough order of preference):

- **At create time, set `sandbox.mode = "off"` for mycelium-controlled openclaw agents** (matches the pattern existing working agents like `julia-agent`/`selina-agent` already use), OR
- **Warn loudly at create time** if the agent will inherit a sandboxed mode that hides the CLI — reuse the existing `doctor` check logic, OR
- **Make the CLI available inside the sandbox** (bake `mycelium` into `openclaw-sandbox-python` + mount `~/.mycelium`) if isolation is desired.

The first option is the least-surprising for the common local case; un-sandboxing is a security decision, so if `create` does it automatically it should at least print that it did.

## Workaround

Set `sandbox.mode = "off"` for each affected agent in `~/.openclaw/openclaw.json`, then `openclaw gateway restart`.

juliarvalenti:
Closes #376.

## Problem

`mycelium agent create --adapter openclaw` prints `allowlisted mycelium CLI for <agent>`, but the new agent inherits openclaw's default per-agent Docker sandbox (`agents.defaults.sandbox.mode = "all"`). Inside that Linux container the host `mycelium` binary doesn't exist, so the agent reports `mycelium: not found` on first dispatch — `session join`/`message propose`/`message respond` all fail. Allowlisting a command is meaningless when the binary isn't present in the execution environment.

We confirmed there's **no config-only escape**: openclaw's sandbox supports `docker.binds`, but the host is macOS and the sandbox is a Linux container, so a host-built binary can't be bind-mounted and executed across the OS/arch boundary. Putting the CLI *inside* the sandbox is a real feature (bake it into the image) tracked separately. The pragmatic, reliable fix is to run created agents on the host.

## Change

- **Created (greenfield) openclaw agents**: `register()` now calls `_disable_sandbox()`, pinning `sandbox.mode = "off"` in `openclaw.json` — matching how existing wired-in agents (e.g. `julia-agent`, `selina-agent`) are already configured, and what `mycelium doctor`'s `_check_openclaw_agent_sandbox()` recommends after the fact. We just apply it at create time so the agent works on first dispatch instead of failing silently until someone runs `doctor`.
- **Adopted agents**: user-owned, so we don't silently override their sandbox. `_warn_if_sandboxed()` surfaces the same diagnosis `doctor` would, with the manual fix.
- Un-sandboxing is logged loudly (it's security-relevant). Both helpers are best-effort — a missing record / unreadable config warns rather than aborting registration.

## Tests

- `_disable_sandbox` flips a created agent's `sandbox` to `{mode: off}`.
- `_disable_sandbox` on a missing record doesn't raise.
- `register()` calls `_disable_sandbox` for created agents and only `_warn_if_sandboxed` for adopted ones.

Full suite: 372 passed; ruff + ty clean.

### DIFF
--- a/mycelium-cli/src/mycelium/integrations/openclaw/dispatch.py
+++ b/mycelium-cli/src/mycelium/integrations/openclaw/dispatch.py
@@ -136,6 +136,68 @@ class OpenClawIntegration(Integration):
             f'"{mycelium_bin}"'
         )
 
+    def _disable_sandbox(self, agent_id: str) -> None:
+        """Set ``sandbox.mode = "off"`` for *agent_id* in openclaw.json.
+
+        A wired-in agent must be able to exec the host ``mycelium`` binary
+        (negotiate, memory writes, plan task ops). OpenClaw's default
+        per-agent sandbox runs every command inside a Linux container
+        (``openclaw-sandbox-python``) that has neither the host-installed
+        ``mycelium`` binary nor access to ``~/.mycelium`` — so the agent
+        reports ``mycelium: not found`` mid-conversation. Mounting the host
+        binary in via ``sandbox.docker.binds`` can't help: the host binary
+        is built for the host OS/arch and won't execute in the Linux
+        container. The only reliable fix today is to run the agent on the
+        host (``mode = "off"``), matching how existing wired-in agents are
+        configured. ``mycelium doctor`` would otherwise flag this after the
+        fact (``N channel agent(s) are sandboxed — mycelium CLI invisible``);
+        we apply it at create time so the agent works on first dispatch.
+
+        Un-sandboxing is a security-relevant change, so it's logged loudly.
+        Best-effort: a missing agent record or unreadable config warns rather
+        than aborting the registration.
+        """
+        oc, oc_json = self._read_block()
+        agents = ((oc.get("agents") or {}).get("list")) or []
+        record = next((a for a in agents if a.get("id") == agent_id), None)
+        if record is None:
+            console.print(
+                f"[yellow]could not set sandbox=off for {agent_id}[/yellow] "
+                f"(no agent record in {oc_json}). If the agent can't run the "
+                f"mycelium CLI, set sandbox.mode = 'off' for it manually and "
+                f"restart the gateway."
+            )
+            return
+        if (record.get("sandbox") or {}).get("mode") == "off":
+            return  # already on the host
+        record["sandbox"] = {"mode": "off"}
+        oc_json.write_text(json.dumps(oc, indent=2) + "\n")
+        console.print(
+            f"  [green]sandbox[/green] off for [cyan]{agent_id}[/cyan] "
+            f"[dim](runs on host so it can exec the mycelium CLI)[/dim]"
+        )
+
+    def _warn_if_sandboxed(self, agent_id: str) -> None:
+        """Warn (don't change) if an adopted agent's sandbox hides the CLI.
+
+        Adopted agents are user-owned; flipping their sandbox silently would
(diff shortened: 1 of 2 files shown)


mycelium-cli/src/mycelium/integrations/openclaw/dispatch.py
```diff
@@ -430,10 +492,17 @@ def register(
...
             self._create_agent(agent_id=agent_id, description=manifest.description)
+            # Greenfield agents inherit openclaw's default per-agent sandbox,
+            # which hides the host mycelium CLI (see _disable_sandbox). Run
+            # them on the host so they can coordinate on first dispatch.
+            self._disable_sandbox(agent_id)
         else:
             console.print(
                 f"  [green]adopting[/green] existing OpenClaw agent [cyan]{agent_id}[/cyan]"
             )
+            # Don't override a user's deliberate sandbox config on adopt; just
```
sfph, line 503: This is nice :)

sfph (approved):
LGTM

juliarvalenti:
Superseded by #388, which is the better fix: it routes `tools.exec.host=gateway` so agents can exec the mycelium CLI **while keeping the sandbox on**, instead of disabling the sandbox entirely as this PR did. #388 also identified the second failure mechanism (the exec-approvals allowlist is bypassed for sandbox-hosted exec) that this PR missed. The unit tests + None-safety hardening from here were carried over to #388. Closing in favor of #388.

PR #388, opened 2026-06-30
fix(openclaw): route exec to gateway host instead of disabling sandbox

sfph:
## Problem

When an OpenClaw agent's sandbox is enabled (the default for non-main sessions), `mycelium` CLI calls fail silently in two compounding ways:

1. **Wrong spawn host** — exec lands inside the sandbox container where `mycelium` isn't installed.
2. **Allowlist bypassed** — the exec-approvals allowlist written by `mycelium agent add openclaw` is only consulted when exec is gateway-hosted; it's unconditionally skipped (`approvalPolicy = undefined`) for sandbox-hosted exec.

The existing guidance was to set `sandbox.mode = 'off'`, which lifts all sandbox isolation — a broader change than necessary.

## Fix

`tools.exec.host = 'gateway'` is the minimal correct alternative: it routes only the `exec` tool to the gateway host (where `mycelium` lives and the allowlist is evaluated), while keeping sandbox isolation intact for `read`, `write`, `edit`, and `apply_patch`.

- `_configure_exec_host_gateway(agent_id)` writes the setting into `openclaw.json` automatically during `mycelium agent add openclaw` and `mycelium adapter add openclaw`, paired with the existing `_allowlist_mycelium()` call.
- `mycelium doctor` now accepts either fix — it flags an agent only when both `sandbox.mode != 'off'` AND `tools.exec.host != 'gateway'`. Warning details explain both options.
- Agents not in `agents.list` (e.g. the implicit `main` handle) are skipped silently — they cannot be patched programmatically.

## Changes

- `dispatch.py` — new `_configure_exec_host_gateway()`, called alongside `_allowlist_mycelium()` in `register()` and `register_adopted_batch()`
- `doctor.py` — `_check_openclaw_agent_sandbox()` accepts either fix; updated warning details list both options
- `docs/design/openclaw-exec-host-vs-sandbox.md` — explains the two-layer failure, the comparison table between options A and B, and annotated `openclaw.json` examples
- Skill docs updated to mention both options

## Test plan

- [x] `uv run ruff check . && uv run ruff format --check . && uv run ty check .` — all pass
- [x] Live negotiation with `lawyer-a` (OpenClaw agent, sandbox enabled, `exec.host=gateway` set) reached consensus (`broken: false`, `plan/tasks.md` compiled) — verified on `negotiation-test` room session `91c3f034`
- [ ] `mycelium doctor` on a host with a sandboxed agent that has `exec.host=gateway` shows green
- [ ] `mycelium doctor` on a host with a sandboxed agent missing both fixes shows warning with both option A and B in details

### DIFF
--- a/.claude/skills/e2e/SKILL.md
+++ b/.claude/skills/e2e/SKILL.md
@@ -178,7 +178,9 @@ mycelium session join --handle agent-delta -m "Ship safe" -r e2e-test-room
 
 Test that OpenClaw agents get woken by coordination ticks and respond autonomously.
 
-**Prerequisites**: OpenClaw gateway running, mycelium adapter installed, agents configured with sandbox=off.
+**Prerequisites**: OpenClaw gateway running, mycelium adapter installed, agents able to exec mycelium CLI. Two ways to achieve this per agent in `openclaw.json`:
+- Option A (simpler): `"sandbox": {"mode": "off"}`
+- Option B (preserves container isolation): `"tools": {"exec": {"host": "gateway"}}` — routes exec to the gateway host where mycelium is installed while keeping sandbox isolation for read/write/edit.
 
 ```bash
 # Verify gateway + plugin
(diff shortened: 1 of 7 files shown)


juliarvalenti (approved):
Approving. Reviewed the two-layer root cause (wrong exec host + bypassed allowlist) and the `exec.host=gateway` fix — it's the minimal correct routing. Verified locally end-to-end: a genuinely sandboxed agent (runtime: sandboxed, mode: all) successfully ran `mycelium --version` from inside via exec.host=gateway, which previously failed with 'not found'. Added unit tests + None-safety hardening (pushed to the branch). CI green.

PR #389, opened 2026-06-30
feat: web-triggered demos via a host-side hub runner

juliarvalenti:
## Problem

In a hub-and-spoke deployment the stack and the OpenClaw gateway run on a shared remote host. Today the only way to start a coordination is to SSH into that host and drive the gateway CLI by hand — so a fresh Mycelium instance is an empty world with no obvious way in.

This adds a **"Run a sample coordination"** button to the web GUI that provisions real agents on the hub and routes the user straight into the live room — no terminal, no SSH.

## Why a host-side runner

Provisioning an OpenClaw agent edits `~/.openclaw/openclaw.json` and runs `openclaw gateway restart` — host-level operations a container can't do without mounting the host FS and baking in the CLIs. The gateway is itself a host process. So the executor is a **host process** (`mycelium hub serve`) next to the gateway; the containerized backend reaches it over the host gateway.

## What's here

- **`mycelium hub serve`** (CLI) — localhost HTTP runner reusing the `mycelium demo` orchestration in-process; background provisioning + job polling; optional `MYCELIUM_HUB_TOKEN` bearer auth.
- **`/api/demos` proxy** (backend) — env-gated on `HUB_RUNNER_URL`; **404s when unset**, so local single-host installs are unchanged.
- **GUI** — empty-state CTA + launcher dialog (scenario picker → start → poll → route into room). All affordances hidden unless the runner answers.
- **Dataset fetch** uses `GITHUB_TOKEN`/`GH_TOKEN` to dodge GitHub's unauthenticated 60/hr limit on a shared host.
- **compose** passes the env through + adds `host.docker.internal:host-gateway` (inert when unset).
- **Runbook**: `docs/design/hub-runner-web-demos.md` (runner service, backend wiring, the mandatory pre-authenticated seed agent, security gate).

## Tests

- CLI: HTTP contract (routing, auth, validation, job lifecycle) over a loopback server with provisioning stubbed.
- Backend: proxy gate (404 unconfigured), forwarding contract, token attach, error/unreachable passthrough.

## Not verified end-to-end yet

The actual `POST /demos` happy path needs a real gateway and was not exercised in CI (no gateway available). To validate on a hub, follow the runbook; the **pre-authenticated seed agent** is the easy-to-miss step (fresh agents have no model token and hang silently).

## Open question

The hub GUI spawns LLM agents on a click = spend + RCE surface. It should sit behind an auth front door (SSO/reverse proxy) before it's broadly reachable — flagged as a TODO in the runbook.


juliarvalenti:
This is confusing cuz we have "hub and spoke" but "hub" is being reused here in a diff context

juliarvalenti:
Superseded by the SLIM-native rewrite (#418): built on removed subsystems (see the diff). The salvageable idea is tracked in #444 — closing this stale PR; the branch stays in git history as the reference implementation.

PR #391, opened 2026-07-02
feat!: Mycelium 2.0.0 — IOC L9 protocol + Go CFN (ioc-cfn-svc) cutover

juliarvalenti:
## What

**This is the 2.0.0 branching point.** Mycelium natively speaks the [IOC L9 epistemic protocol](https://github.com/outshift-open/ioc-protocols-models) and targets the Go CFN ([ioc-cfn-svc](https://github.com/outshift-open/ioc-cfn-svc)) **exclusively** — the python CFN client and all backwards-compat machinery are removed. Full scoping doc: `docs/l9-integration-plan.md` (in this PR, with a 2.0.0 addendum).

The agent-facing surface is **additive**: L9 keys ride inside existing coordination message content, all epistemic fields are optional, and a session run by non-L9 agents produces byte-identical payloads to today. The breaking change is the engine underneath: the cfn profile now runs `ghcr.io/outshift-open/ioc-cfn-svc` and the semantic-negotiation API path no longer exists.

## L9 integration

- **Envelopes on coordination messages** — ticks are `exchange`, consensus is `commit:converged`/`commit:abort`, with episode URNs and causal `message.parents`. The backend synthesizes reply envelopes, so the causal graph is complete without L9-aware agents. Every negotiation leaves a full episode record at `log/episodes/{short_id}.md` in room memory.
- **Epistemic negotiation** — on negotiate replies: `--confidence`, `--reasoning`, the 3-way evidence split `--supporting-evidence`/`--against-evidence`/`--addresses`, `--revision-cause` (why a belief moved: `grounded_argument`/`new_evidence`/`semantic_memory`/`repair_resolution`/`social_compliance`), and `--defer-to <handle>` (shorthand for `social_compliance`). Consensus gets quality metrics computed by mycelium: **MPC** (mean final confidence), **GAR** (genuine agreement ratio, guarded at the MPC=0.5 degeneracy), **SCR** (social compliance ratio — now the spec definition: fraction of belief *revisions* that were compliance, over agents that revised, replacing the old accepts-that-deferred proxy), provenance weight `(1−SCR)×GAR`. Grounding is scored via `contingency_score = |addresses ∩ prior evidence| / |addresses|` (≥0.40 grounded); a move with no grounding flags gets the benefit of the doubt (genuine), so non-adopters are never penalized. Rendered in the UI, CLI watch output, and both openclaw formatters; taught to agents in all three adapter SKILL.md files.
- **Live metrics + trust gate** — `mycelium negotiate status` surfaces the interim MPC/GAR/SCR/provenance_weight mid-negotiation; `--contested` exits 2 when provenance_weight < 0.60, a CI/orchestration gate on acting on a weakly-supported agreement.
- **Team priors, local by default** — a converged episode writes its rule to the room's own memory at `l9/rule_update/topic` (posterior/gar/scr/provenance_weight, `episode_count`), and the next negotiation reads it back as the tick `team_prior`. Fully local (markdown + frontmatter), no external service. The CFN knowledge fabric (`L9_CFN_ENABLED=false`, `POST /api/l9/messages`, requires a knowledge CE) is now only a fallback when no local rule exists.

> **Honest framing, documented in the tree.** These are *self-reported* epistemics: the confidence float and evidence strings are agent-generated; mycelium validates their shape (range/enum/non-empty) but not their truth or calibration, and grounding is a gameable token-overlap. The metrics are exact arithmetic over self-report — read `provenance_weight` as "how the team described its reasoning," not a measured trust score. Observed-vs-reported (Theory-of-Mind / drift detection) is the upstream Drift Detection CE (`ioc-cfn-cognition-engines#33`) we'd consume, not build. Spelled out on the internal L9 page below.

**Docs.** The consumer-facing L9 page (`concepts.html`) is updated for the new flags and genuine SCR. A new **internal (unlinked) reference** at `docs/l9-integration.html` maps the mycelium ⇄ CFN ⇄ CE relationship, pins every deployed version, shows a Mermaid sequence diagram of the round loop and a command→payload walkthrough, and states the self-reported caveat plainly.

## Go CFN cutover (the 2.0.0 part)

- **Removed** the short-lived `CFN_API_FLAVOR` dual-stack flag (never shipped in a release). The python CFN's semantic-negotiation path is gone.
- `cfn_negotiation.py` speaks semantic-alignment only: `participant_id` reply keying, `final_result` envelope agreement, `trace`/`meta`/`shared_memory` extras surfaced as telemetry + `cfn_persisted` on consensus.
- The generated `ioc_cfn_svc_api_client` was **regenerated from the Go CFN's swagger** (not deleted) — see the typed-client section below.
- `compose.yml` cfn profile runs `ghcr.io/outshift-open/ioc-cfn-svc:${CFN_IMAGE_TAG:-v0.2.0}` — **pinned** to a released tag (not `:latest`), overridable via `CFN_IMAGE_TAG` for the smoke test, multi-arch (no more `platform: linux/amd64` pin). The service keeps the historical `ioc-cognition-fabric-node-svc` DNS name so existing installs' saved `COGNITION_FABRIC_NODE_URL` values survive the upgrade. Same `cfn_cp` database, port, and mgmt-plane wiring — MAS IDs and workspaces live in the mgmt plane and carry over.

## Typed CFN client (the data-contract boundary)

The negotiation path — the one that silently produces an empty plan on a shape change — goes through a **generated typed client**, stronger than the pre-2.0.0 client (whose responses were `Any`). Three enforcement layers in `cfn_negotiation.py`:

1. **Typed requests** (`StartRequest`/`DecideRequest`/`AgentReply`) — `ty` catches a wrong field at the call site; `agent_id`→`participant_id` mapping and "epistemic extras never forwarded to CFN" are guaranteed by the type.
2. **Schema-validated responses** parsed into `StartResponse`/`DecideResponse`.
3. **Presence assertions** (`_require`) — swaggo marks every field optional, so a CFN rename would silently be `UNSET`; we raise `CfnNegotiationError` instead of writing an empty agreement.

Generation (`scripts/gen-cfn-client.sh`, deterministic from the committed `fastapi-backend/cfn_swagger.json`): Swagger 2.0 → OpenAPI 3 via `npx swagger2openapi`, then `openapi-python-client`. CI-able drift check: regenerate + `git diff --exit-code`.

**Already earned its keep:** the typed layer surfaced a latent bug — the code read the python CFN's `_usage`/`payload.status`, but the Go CFN returns usage in a typed `meta` field and `status` top-level, so token metrics were silently recording nothing. Fixed.

**Two honest limits, both the CFN's own swagger:** `messages` (Go `json.RawMessage` → swaggo `[]int`, really a list of envelope objects) and `final_result` (`map[string]interface{}`) are under-typed at the source, so those stay dict-navigated via the validated model's `to_dict()`. `cfn_knowledge.py` stays on httpx **by design** — its `payload.data` is also `json.RawMessage` and the `graph/*` endpoints aren't in the swagger; typing them would be fragile, not robust. (Worth a `swaggertype` annotation upstream on ioc-cfn-svc to close both.)

## Decisions encoded

- **Go CFN subkind table is authoritative** (team decision): `commit: converged|resolved|abort` — failed negotiations commit as `abort`.
- **L9 bindings vendored** (`app/services/l9_models.py`): the `ioc-l9-all-models` PyPI package requires `litellm>=1.89.3`, conflicting with our security-pinned `litellm[proxy]==1.83.14`. Upstream fix worth filing: make that dep optional.
- **CFN contract enforcement**: a regenerated typed client for the negotiation path (details below). The regenerate-and-`ty check` ritual is restored and now CI-able from a committed swagger snapshot.

## Before cutting v2.0.0

**Known follow-up:** the openclaw channel `route.ts` formatters don't yet render the new `coordination_retry` message or the consensus `reason` field (two-delivery-paths rule) — openclaw agents just won't see the retry notice (informational, non-blocking). CLI/UI render it.

One live smoke test against a running Go CFN stack (`e2e` coordination scenario) — mgmt-plane 0.1.1 ↔ ioc-cfn-svc config-push compatibility and a full multi-round SAO loop through the semantic-alignment CE. The typed client validates response *structure* against the swagger, but the runtime message-flow (ongoing-round `messages` fan-out, `final_result` agreement extraction — the two under-typed fields) has not hit a live instance. Release via `/release` with tag `v2.0.0` (tag-driven versioning; pyprojects stay at placeholder per the release workflow).

## Live-integration port (from sfph's play_with_cfn_golang)

Merged the empirically-validated fixes from the branch sfph ran end-to-end against the live Go CFN. The typed client validates *structure*; these are the runtime behaviours you only find by running it.

**Naming (team decision):** `COGNITION_FABRIC_NODE_URL` → `CFN_SVC_URL`; compose service → `ioc-cfn-svc`. A 2.0.0 break; `mycelium config apply` regenerates `.env` so a stale key self-heals.

**Full CFN stack (the "images unknown" gap, answered):** the cfn profile now runs the validated 4-service set pinned to a coherent `2026-06-23-*` group — `ioc-cfn-mgmt-plane-svc`, `ioc-knowledge-memory-svc`, `ioc-cfn-svc`, `ioc-cfn-cognition-engine` (auto-registers) — with the real env wiring. `mycelium-db` gains TimescaleDB (Dockerfile.agensgraph — the published image must be rebuilt+republished) and an `ioc-graph-db` database (initdb). Per-service `*_IMAGE_TAG` overrides.

**Empirical Go-CFN shapes (only findable live):** token usage in `.meta` not `_usage`, status top-level not `payload.status` (both fixed in coordination); memory-provider registration wants `name`+bool (not `memory_provider_name`+`"True"`); agent-registration now logs 4xx/5xx instead of masking them; knowledge ingest is HTTP 202 + otel-trace/ioa_observe format (`knowledge_fanin`, `routes/knowledge`); indexer stale-room guard + atomic upsert.

**Retry/timeout fix (Selina's headache):** the CFN default `retry_max_attempts=3` silently restarts a negotiation from round 1 (score < `validation_score_intervention` 0.6), running a session several times over. Fix: `_ensure_mas` sends `mas_config={retry_max_attempts, validation_score_intervention}` (settings default `CFN_RETRY_MAX_ATTEMPTS=1`). A CFN-internal retry is surfaced as a `coordination_retry` message (round-regression detection) and rendered in the UI; `_finish_cfn(reason=...)` threads agreed/abort/timeout onto consensus.

**Timeout boundary made explicit:** the CFN never waits for agents (it takes the full reply batch in one `/decide` and returns a structured `status:"timeout"`). Agent responsiveness is mycelium's round watchdog, which restarts per agent reply — essential for single-threaded/serialized agent runtimes, and it runs entirely before `/decide`. So `CFN_DECIDE_TIMEOUT_SECONDS` (300 → 600) is a **dead-connection backstop above** the CFN's internal timeout, not a control. Documented so it isn't collapsed into a session-wide deadline (which would re-break the single-threaded case).

## Upgrade notes (for release)

- **`CFN_API_FLAVOR` in `~/.mycelium/.env`**: users who ran `mycelium config apply` during the dual-stack window have a stale `CFN_API_FLAVOR=negotiation` line. Harmless (the backend ignores unknown env vars), and `mycelium config apply` fully regenerates `.env` so it self-heals on the next apply. Safe to delete by hand.
- **CFN image pins**: the cfn profile pins the validated `2026-06-23-*` set (per-service `*_IMAGE_TAG` overrides). Bump the set together once a newer group is validated.
- **Naming break**: `COGNITION_FABRIC_NODE_URL` → `CFN_SVC_URL`. Run `mycelium config apply` after upgrade; `.env` regenerates.
- **Retry default**: a room's MAS now uses `retry_max_attempts=1` (predictable timing). Set `CFN_RETRY_MAX_ATTEMPTS` higher, or lower `CFN_VALIDATION_SCORE_INTERVENTION`, to change it.

## Testing

- backend: 346 passed (L9 episode/genuine-SCR/grounding/writeback + typed-client boundary + 4 CFN-contract tests), ruff + ty clean
- CLI: 439 passed (epistemic reply flags + status `--contested`), full gate clean
- openclaw plugin: tsc clean, 123 tests passed; frontend build clean
- compose: `docker compose config` validates (cfn profile, both files)


sfph:
## PR #391 Review: IOC L9 Protocol Integration

**Scope:** +3,635 / −138 lines across 36 files

---

### Overall Assessment

Strong integration. The additive design (L9 keys optional inside existing content JSON, no schema migration, fail-soft everywhere) means this genuinely achieves zero-disruption for non-L9 agents. The dual-delivery-path rule is honored, the vendoring rationale is sound, and the test suite is thorough.

That said, there are a few issues worth addressing before merge.

---

### Bug: Broken HTML in generated `concepts.html`

The epistemic annotations section has garbled rendering. The markdown-to-HTML generator is splitting list continuations into separate `<ul>`/`<p>` blocks:

```html
<h2 id="sessions-epistemic-annotations">Epistemic annotations</h2>
<p>Sessions carry an optional epistemic layer from the <a href="#l9-protocol">L9 protocol</a>:</p>
<ul>
  <li>Replies may include <code>--confidence</code>, <code>--evidence</code>, <code>--reasoning</code>, and (on an</li>
</ul>
<p>accept that yields without persuasion) <code>--defer-to &lt;handle&gt;</code>.</p>
<ul>
  <li>Ticks may include a <code>team_prior</code> — how this team has agreed on this topic in</li>
</ul>
<p>past episodes (requires the L9 CFN knowledge fabric).</p>
```

Each bullet's continuation line becomes a separate `<p>`, producing mangled output. The source `sessions.md` probably has line-wrapped bullets that the generator doesn't handle. Either unwrap those bullets to single lines or fix the generator.

---

### Performance: httpx client created per request (alignment path)

`_post_alignment` instantiates and tears down an `httpx.AsyncClient` on every call in `fastapi-backend/app/services/cfn_negotiation.py`. Same issue in `l9_cfn.post_l9`.

This defeats connection pooling and adds TCP+TLS setup latency to every round. The negotiation path runs multiple `/decide` calls per session, so this compounds. Consider a module-level client (with lazy init + shutdown hook) or pass a shared client through `_CfnRoundState`.

---

### Fire-and-forget task reference

In `coordination.py`, `asyncio.ensure_future` is used for `l9_cfn.knowledge_write_consensus` without storing the returned task.

`asyncio.ensure_future` without storing the returned task means:

1. The task can be garbage collected before completion (unlikely but spec-undefined).
2. If the coro raises, Python 3.12 emits a noisy "Task exception was never retrieved" warning.

Since this is intentionally fire-and-forget, I'd suggest `asyncio.create_task` plus swallowing exceptions inside `knowledge_write_consensus` (which already does this via the `post_l9` wrapper). Or store the task reference on the state and let it be GC'd with the state dict entry.

---

### Episode record doesn't match documentation

The docs (integration plan section 2.2, concepts page) say "YAML frontmatter plus a JSONL block," but the actual output in `l9_episode.write_episode_record` is markdown headers and bullet points with a fenced JSONL block.

Not a bug (the memory system handles it fine), but the docs should say "markdown file with a JSONL fenced block" rather than "YAML frontmatter."

---

### Minor: `bool` subclass of `int` inconsistency

`sanitize_epistemic_fields` correctly guards against bool-as-confidence:

```python
if isinstance(conf, int | float) and not isinstance(conf, bool) and 0.0 <= conf <= 1.0:
```

But `record_reply` doesn't:

```python
if isinstance(conf, int | float) and 0.0 <= conf <= 1.0:
```

In practice the sanitizer runs upstream so booleans are already stripped, but for consistency the `record_reply` guard should also exclude `bool`.

---

### Things I Like

**Fail-soft discipline is consistent.** Every L9 operation is wrapped in try/except with logger.exception and a continuation path. The team-prior query, episode setup, tick/reply recording, consensus envelope, episode persistence, and knowledge write are all individually fail-soft. One failure doesn't cascade.

**Backward-compatibility tests are explicit.** `test_propose_reply_legacy_wire_shape_is_byte_identical` and `test_respond_reply_legacy_wire_shape_is_byte_identical` assert exact JSON strings, not just structural equivalence. This is the right test for an additive protocol change.

**GAR/SCR metric design is thoughtful.** The `compute_metrics` threshold (at least N-1 agents must report confidence) avoids generating meaningless scores from partial data, and the GAR definition — confidence moved in the direction of the outcome relative to the agent's first stated confidence — correctly distinguishes earned agreement from starting-position agreement.

**Dual-stack is clean.** The alignment path branches early in `start_negotiation`/`decide_negotiation`, uses the same timing/metrics instrumentation, and maps `agent_id` → `participant_id` transparently. The `_normalize_cfn_decide_response` test covers the `final_result` envelope shape.

**SKILL.md updates are protocol-shaped, not augmentation-shaped.** The epistemic instructions ("state your confidence and cite your sources," "defer honestly") extend the negotiate step without turning the skill into an LLM prompting layer.

---

### Summary

| Priority | Item |
|---|---|
| Fix before merge | Broken HTML rendering in `concepts.html` epistemic annotations section |
| Fix before merge | `asyncio.ensure_future` → `create_task` or store task ref |
| Should fix | httpx client-per-request in `_post_alignment` and `post_l9` (perf) |
| Nit | `bool` guard missing in `record_reply` |
| Nit | Episode record docs say "YAML frontmatter" but it's markdown |

juliarvalenti:
Thanks @sfph — all five items addressed in b4954c6. Note your review predates the 2.0.0 cutover commit (8cb69e5), which changes the picture on a couple of them:

- **Broken HTML in concepts.html** — real; the generator splits line-wrapped bullets into separate `<ul>`/`<p>` blocks. Unwrapped the `sessions.md` bullets and regenerated; the epistemic-annotations section now renders as one list. (The similar-looking `memory.md` case turned out to be a legit list-then-paragraph, not the bug.)
- **httpx client-per-request** — fixed, and it matters more now than when you reviewed: the cutover made the alignment path the *only* path and converted `cfn_knowledge.py` to httpx as well. New `app/services/cfn_http.py` keeps a shared `AsyncClient` cached per event loop (so pytest-asyncio's per-test loops can't reuse a client bound to a dead loop), closed via the app lifespan. `_post_alignment` and `post_l9` both use it with per-request timeout/header overrides.
- **`ensure_future` without a task ref** — fixed with a module-level strong-ref set + done-callback discard (`_spawn_background`). Exceptions can't surface as "never retrieved" warnings since `knowledge_write_consensus` swallows everything via `post_l9`, but the GC concern was legit.
- **Episode record docs** — corrected to "markdown summary + fenced JSONL block" in both the L9 concepts section and the integration plan. (`write_memory_file` does add its own frontmatter on disk, but the metrics/outcome live in the body, so the old wording was wrong either way.)
- **`bool` guard in `record_reply`** — added, matching the sanitizer.

One heads-up on the review's "dual-stack is clean" note: the dual-stack no longer exists — 8cb69e5 removed the python CFN path, the `CFN_API_FLAVOR` flag, and the entire generated `ioc_cfn_svc_api_client` by team decision (this PR is the 2.0.0 branching point; see the updated PR description). The `_normalize_cfn_decide_response` coverage you called out still applies — it's now the only decide shape.

Gates after the fixes: backend 319 passed, ruff/ty clean; docs regenerate clean.

sfph:
## Follow-up: 2.0.0 cutover review (`8cb69e5`)

This commit is a **−3,298 net line deletion** that removes the entire generated `ioc_cfn_svc_api_client`, the `CFN_API_FLAVOR` dual-stack flag, and the Python CFN's `semantic-negotiation` path. Mycelium now talks to **ioc-cfn-svc only** via plain httpx (`semantic-alignment` + shared-memories). Combined with the follow-up fix commit (`b4954c6`, shared `cfn_http.py`), the PR is in good shape — with a few pre-release gaps.

---

### What the cutover does well

**The upgrade path is thoughtful.** Keeping the compose service name `ioc-cognition-fabric-node-svc` while swapping the image to `ghcr.io/outshift-open/ioc-cfn-svc` means existing `COGNITION_FABRIC_NODE_URL` values (`http://ioc-cognition-fabric-node-svc:9002`) keep working without a config migration. Same `cfn_cp` DB, port 9002, and mgmt-plane wiring — workspace/MAS IDs should carry over.

**The code simplification is real.** `cfn_negotiation.py` drops from ~400 lines of dual-branch + typed client to a single alignment httpx path. `cfn_knowledge.py` follows the same pattern. Deleting ~3k lines of generated client code is the right call if you're never going back.

**Docs are internally consistent.** CLAUDE.md, concepts page, L9 protocol docs, config reference, precommit skill, and CI notes all reflect alignment-only. The integration plan has a clear 2.0.0 addendum explaining why dual-stack was built then removed.

**Agent surface stays additive.** The breaking change is entirely under the hood (CFN image + API path). L9 epistemic fields, CLI flags, UI rendering — all still optional for agents.

---

### Concerns before tagging v2.0.0

#### 1. No live smoke test yet (acknowledged blocker)

The updated PR description explicitly calls this out:

> One live smoke test against a running Go CFN stack — specifically mgmt-plane 0.1.1 ↔ ioc-cfn-svc config-push compatibility and a full multi-round SAO loop through the semantic-alignment CE.

All CFN coverage is unit-tested with mocked httpx. The alignment request shapes, `participant_id` keying, and `final_result` parsing are tested — but nothing has hit a real ioc-cfn-svc instance. For a `feat!` that deletes the only working CFN path, I'd treat this as a **merge blocker for release**, not just a nice-to-have.

#### 2. `:latest` tag is a step backward from pinning

```yaml
# before
image: ghcr.io/outshift-open/ioc-cognition-fabric-node-svc:0.1.7
platform: linux/amd64

# after
image: ghcr.io/outshift-open/ioc-cfn-svc:latest
pull_policy: always
# platform pin removed (good for arm64)
```

Removing `platform: linux/amd64` is correct for multi-arch. But going from a pinned `0.1.7` to `:latest` with `pull_policy: always` means every `docker compose up` can silently pull a breaking CFN change. The integration plan itself notes `{date}-{sha}` tags exist on ghcr. Consider pinning to a known-good digest or dated tag once the smoke test passes.

#### 3. Typed client → manual swagger diff is a weaker contract gate

The old workflow was: regenerate client → `ty check` catches field renames at call sites. The replacement is manual: diff upstream `docs/swagger.json` against httpx call sites when the image is bumped. That's documented in precommit + CI, but it's **not automated** and easy to forget. Worth considering a lightweight contract test (even just asserting URL paths + required JSON keys against a checked-in swagger fragment) before the first 2.0.0 release.

#### 4. README still references the old CFN repo

The acknowledgements section still links to `ioc-cognition-fabric-node-svc` as the fabric orchestration component. Should point at `ioc-cfn-svc` now. Small but visible on the landing page.

#### 5. Stale `CFN_API_FLAVOR` in existing `.env` files

The config key, validator, and `.env` emission are all removed (good). But users who ran `mycelium config apply` during the dual-stack window will have `CFN_API_FLAVOR=negotiation` sitting in `~/.mycelium/.env`. Harmless (backend ignores unknown env vars), but confusing. A one-line note in release notes ("safe to delete `CFN_API_FLAVOR` from `.env`") would help.

---

### What changed between cutover and fix commit

`8cb69e5` still had per-request httpx clients in `cfn_negotiation.py` and `l9_cfn.py`. `b4954c6` added `cfn_http.py` with a per-event-loop shared client and wired it into the app lifespan. So the performance concern from the first review is addressed — and matters more now that **every** CFN call path uses httpx.

---

### Verdict

| Area | Status |
|---|---|
| Architecture (Go CFN only) | Sound — team decision is clear, code matches |
| Upgrade compatibility (DNS/DB/port) | Good |
| L9 agent surface | Unchanged — still additive |
| Code quality / deletion cleanliness | Good — no orphaned imports, tests updated |
| Live integration validation | **Missing** — author agrees, blocks v2.0.0 tag |
| Image pinning | **Weaker** than before (`:latest`) |
| Contract safety net | **Weaker** than typed client (manual process) |
| Docs | Good except README ack link |

**Recommendation:** Approve the L9 integration and cutover direction. Hold the **v2.0.0 tag** until someone runs the live smoke test the PR description describes (mgmt-plane config-push + full SAO loop). Pin the CFN image to something other than `:latest` once you know which build works.

juliarvalenti:
Update on the CFN data contract (658b0e1) — reverses the earlier "no generated client" decision after @juliarvalenti pushed on it:

The generated client is **back and stronger** than pre-2.0.0. `cfn_negotiation.py` now goes through a typed client regenerated from the Go CFN's swagger (`scripts/gen-cfn-client.sh`: Swagger 2.0 → OAS3 → openapi-python-client, deterministic from the committed `cfn_swagger.json` snapshot). Typed requests + schema-validated responses + presence assertions on depended-on fields. The old client typed requests but left responses `Any`; the Go swagger types both, so this is a net upgrade.

It immediately caught a latent bug: the branch was reading the python CFN's `_usage`/`payload.status`, but the Go CFN returns usage in a typed `meta` field and `status` top-level — token metrics were silently recording zero. Fixed in coordination.py.

Two fields stay dict-navigated because the CFN's own swagger under-types them (`messages` = `json.RawMessage` → `[]int`; `final_result` = `map[string]interface{}`), and `cfn_knowledge.py` stays httpx by design (same RawMessage issue on `payload.data` + graph endpoints absent from swagger). Both documented inline; both closeable with a `swaggertype` annotation upstream.

PR description updated with a dedicated "Typed CFN client" section. Backend 337 passed, ruff/ty clean. The live-CFN smoke test is still the one gate before v2.0.0 — the typed client validates structure, not the runtime message-flow.

juliarvalenti:
Thanks @sfph — sharp review. Status on all five (dc66ee1 + the earlier 658b0e1):

**2. `:latest` → pinned.** Now `ioc-cfn-svc:${CFN_IMAGE_TAG:-v0.2.0}` (v0.2.0 is a real released tag on ghcr), `pull_policy: always` dropped. Overridable via `CFN_IMAGE_TAG` so the smoke test can try a dated-sha without editing compose; bump the default once a build is validated against mgmt-plane 0.1.1.

**3. Typed client → this is now resolved, not just documented.** After you filed this, I regenerated a typed client from the Go CFN's swagger (658b0e1). So the contract gate is back to the automated form you wanted: `ty check` catches field renames at call sites, and `gen-cfn-client.sh` regenerate + `git diff --exit-code` is CI-able from the committed `cfn_swagger.json` snapshot — plus presence-assertions catch swaggo's everything-optional silent drift. Your "assert URL paths + required keys against a checked-in swagger fragment" suggestion is essentially what the generated client + `_require()` now do.

**4. README ack link.** Fixed — points at `ioc-cfn-svc` now. (The compose service/DNS name stays `ioc-cognition-fabric-node-svc` by design for URL compat; that's not the old repo, so it stays.)

**5. Stale `CFN_API_FLAVOR` in `.env`.** Audited: the emission is clean (no longer written; every remaining CFN key is still consumed by the Go stack). `mycelium config apply` fully overwrites `.env` preserving only `MYCELIUM_IMAGE_TAG`, so the stale line self-heals on the next apply — harmless until then. Added an upgrade note to the PR body ("safe to delete") rather than active cleanup code, matching your suggestion.

**1. Live smoke test.** Agreed, hard blocker for the v2.0.0 tag — tracked, waiting on a reachable Go CFN (local Docker's been down; a dev instance URL also works). The typed client validates response *structure* against the swagger, but the runtime message-flow (ongoing-round `messages` fan-out, `final_result` extraction — the two fields the CFN's swagger under-types) still needs a real instance.

Net: 2, 3, 4, 5 addressed in code/docs; 1 is the standing release gate. Appreciate the thorough pass.

fastapi-backend/app/services/cfn_knowledge.py
```diff
@@ -3,19 +3,29 @@
...
+- ``shared-memories`` create sends ``payload.data`` which the CFN declares as
+  Go ``json.RawMessage`` (``swaggertype:"object"``). At runtime it accepts the
+  openclaw records as a *list*; the generated ``ExtractionPayloadData`` model
+  is object-only, so typing it would misrepresent the real payload.
+- The ``graph/*`` read endpoints are not published in the CFN's swagger
+  (``paths`` omits them), so the generated client has no functions for them.
+
+Shapes are documented inline. If the CFN's shared-memory contract firms up
+(a non-RawMessage data field, graph endpoints added to swagger), migrate these
+onto the typed client the way ``cfn_negotiation.py`` uses it.
```
juliarvalenti, line 28: Knowledge typing issue

fastapi-backend/app/services/cfn_negotiation.py
```diff
@@ -2,16 +2,34 @@
...
+   fields we depend on and raise :class:`CfnNegotiationError` on drift.
+
+Two response fields are deliberately NOT trusted from the typed model:
+``messages`` (Go ``[]json.RawMessage`` — swaggo mis-renders as ``[]int``, but
+the runtime shape is a list of JSON envelope objects) and ``final_result``
+(Go ``map[string]interface{}`` — opaque by the CFN's own contract). We return
+the validated model's ``to_dict()`` to coordination so those two are handed
+over as their real dict form, never as a mis-typed field. The returned dict is
+in the Go CFN's JSON shape (top-level ``status``/``meta``/``final_result``/
+``shared_memory``).
```
juliarvalenti, line 41: Types issue two

akanksha276:
## Live smoke test findings — 2.0.0rc1 on this server

### ioc-cfn-mgmt-plane-svc crash-loops on fresh 2.0.0rc1 install

Upgraded from v1.1.0rc7 → v2.0.0rc1 via `mycelium upgrade --version 2.0.0rc1 && mycelium pull --version 2.0.0rc1`. Backend and DB start healthy, but `ioc-cfn-mgmt-plane-svc` immediately crash-loops with:

```
sqlalchemy.exc.ProgrammingError: (psycopg2.errors.UndefinedTable) relation "agent" does not exist

[SQL: CREATE UNIQUE INDEX IF NOT EXISTS "uq_agent_mas_name" ON "agent" ("mas_id", "name") WHERE "name" IS NOT NULL]
```

The migration is trying to create a unique index on the `agent` table before the table itself exists — a missing dependency ordering in the Alembic migration chain. The `cfn_mgmt` database exists (created by the initdb script), but the `agent` table hasn't been created before the index migration runs because Alembic sees the existing DB at version `0008` (python CFN schema) and skips the table-creation migration.

**Reproducer:** `mycelium pull --version 2.0.0rc1` on an existing install where the `cfn_mgmt` DB was already at schema version `0008` from the python CFN mgmt-plane.

**Workaround:**
```bash
docker stop ioc-cfn-mgmt-plane-svc
docker exec mycelium-db psql -U postgres -c 'DROP DATABASE cfn_mgmt;'
docker exec mycelium-db psql -U postgres -c 'CREATE DATABASE cfn_mgmt;'
docker start ioc-cfn-mgmt-plane-svc
```

The Go CFN mgmt-plane then runs its full migration chain from scratch and comes up healthy. MAS IDs and workspace data from the old schema are lost, but those are recreatable.

Stack is now healthy. Proceeding with negotiation smoke tests.

akanksha276:
## Follow-up: ioc-graph-db also missing on upgrade (same root cause)

After fixing `cfn_mgmt` per the workaround above, `mycelium up` then failed on `ioc-knowledge-memory-svc`:

```
RuntimeError: Database connection failed: FATAL: database "ioc-graph-db" does not exist
```

Same root cause as `cfn_mgmt` — `initdb` doesn't run on an existing volume, so `ioc-graph-db` was never created. Fix:

```bash
docker exec mycelium-db psql -U postgres -c 'CREATE DATABASE "ioc-graph-db";'
```

Then `mycelium up` succeeds and all services come up healthy. The `cfn-db-init` service from the latest commit (36dfdad) should cover both — worth making sure it also handles `ioc-graph-db` in addition to `cfn_mgmt`/`cfn_cp`.

sfph:
> ## Follow-up: ioc-graph-db also missing on upgrade (same root cause)
> After fixing `cfn_mgmt` per the workaround above, `mycelium up` then failed on `ioc-knowledge-memory-svc`:
> 
> ```
> RuntimeError: Database connection failed: FATAL: database "ioc-graph-db" does not exist
> ```
> 
> Same root cause as `cfn_mgmt` — `initdb` doesn't run on an existing volume, so `ioc-graph-db` was never created. Fix:
> 
> ```shell
> docker exec mycelium-db psql -U postgres -c 'CREATE DATABASE "ioc-graph-db";'
> ```
> 
> Then `mycelium up` succeeds and all services come up healthy. The `cfn-db-init` service from the latest commit ([36dfdad](https://github.com/mycelium-io/mycelium/commit/36dfdad8ec78a947b4d64c885091e94586f886c3)) should cover both — worth making sure it also handles `ioc-graph-db` in addition to `cfn_mgmt`/`cfn_cp`.

This should all be taken care of in chore/play_with_cfn_golang.  Does require starting with a clean database.  And making sure you build the db locally (until we're ready to cut over) - so check out the compose-dev.yaml too :)

juliarvalenti:
## 🐛 Snag hit during 2.0.0rc1 semantic negotiation testing: `COGNITION_FABRIC_NODE_URL` missing from `.env` template

**Environment:** Mycelium 2.0.0rc1, CFN stack via `mycelium up --cfn` (OpenClaw + `persona-before-and-after` skill)

**Symptom:** Sessions created successfully and reached `coordination_start`, but no `coordination_tick` ever fired. The session stayed stuck in `waiting`/`negotiating` state indefinitely with no CE activity.

**Root cause:** `fastapi-backend/app/services/coordination.py:_run_tick` has a guard:

```python
if not settings.COGNITION_FABRIC_NODE_URL or not mas_id or not workspace_id:
    logger.error("Coordination requested for %s but CFN not configured...", room_name)
    # posts coordination_error, sets session state=failed, returns
    return
```

`COGNITION_FABRIC_NODE_URL` was not present in `~/.mycelium/.env` (generated by `mycelium config apply`) and not documented anywhere as a required variable. The variable is **distinct from `CFN_SVC_URL`** — `CFN_SVC_URL` is consumed by the mgmt-plane container; `COGNITION_FABRIC_NODE_URL` is what the backend itself uses to reach the CFN `/decide` endpoint at tick time.

**Fix:** Added to `~/.mycelium/.env` manually:

```
COGNITION_FABRIC_NODE_URL=http://ioc-cfn-svc:9002
```

Then had to fully recreate the backend container (`docker stop && docker rm` then `docker compose up`) because `docker restart` does not re-read `env_file:` directives.

**Suggested fixes for this PR:**
1. Add `COGNITION_FABRIC_NODE_URL` to the `.env` template generated by `mycelium config apply` (or derive it from `CFN_SVC_URL` when CFN profile is enabled)
2. Add it to `mycelium doctor` checks alongside the existing CFN config checks — currently `mycelium doctor` validates `CFN_SVC_URL` but not `COGNITION_FABRIC_NODE_URL`, so ticks silently fail while doctor reports green
3. Consider making `_run_tick` log a more prominent error (or surface it via `mycelium session status`) when this variable is missing — the silent-fail makes it very hard to diagnose

**Additional context:** Also found a related snag with container recreation: using `--no-deps` from the wrong working directory puts the backend on `docker_default` instead of `mycelium_default`, breaking DNS for `mycelium-db`. The correct incantation to recreate just the backend is:

```bash
docker compose -p mycelium -f ~/.mycelium/docker/compose.yml --env-file ~/.mycelium/.env --profile cfn up -d --no-deps mycelium-backend
```

This is worth adding to the troubleshooting docs.

juliarvalenti:
## 📊 Semantic negotiation test results — exp-0568 (ex09_ci_cd_release, Haiku, 2.0.0rc1)

Full evaluation: https://gist.github.com/juliarvalenti/7df4709a82d7faf8b1f5ae711b264741

**Setup:** 3 agents (deploy_agent, qa_agent, sre_agent) from the `agent-personas` dataset, `ex09_ci_cd_release` scenario. Before case = plain chat room; after case = Mycelium structured negotiation via OpenClaw channel plugin.

### Before case
- 12 messages, compatible terms reached informally
- sre_agent hallucinated conversation state (said "I haven't heard from the others" after both had agreed) — 2/3 explicit agreements, no shared written plan

### After case (session 9f1f7335)
- 7 complete CFN rounds, clear structured convergence on all 3 issues (deployment cadence, rollback policy, test gates)
- **sre_agent never joined** — posted broadcasts but didn't run `mycelium session join`; context window confusion from earlier failed sessions likely cause
- **CFN RemoteProtocolError at R7** — `ioc-cfn-svc` dropped the `/decide` connection mid-call; agents had both accepted but consensus was lost. Session ended `broken: true`

### Issues surfaced

| # | Issue | Severity |
|---|-------|----------|
| 1 | `COGNITION_FABRIC_NODE_URL` missing from `.env` template (see previous comment) | High |
| 2 | CFN `RemoteProtocolError` drops confirmed consensus — agents accepted but result lost | High |
| 3 | sre_agent couldn't distinguish fresh session after prior failures — didn't join | Medium |
| 4 | Zeuthen oscillation: agents alternate accept/counter_offer for 6+ rounds when both near agreement (strategy instructs counter when `can_counter_offer=true` regardless of bottom line) | Medium |
| 5 | `my-test:session:064c00cb` 404 loop persisted throughout — gateway doesn't expire subscriptions to deleted sub-rooms | Low |

juliarvalenti:
## 🔍 Root cause of `broken: true` — pgvector missing in `ioc-knowledge-memory-svc`

Following up on the CFN `RemoteProtocolError` from the previous comment. Dug into the CFN service logs.

**Causal chain:**

1. `ioc-knowledge-memory-svc` errors with `type "vector" does not exist` (pgvector not installed) starting at **18:51 UTC** — 47 minutes before our session started.

2. During R7 of the negotiation, `ioc-cfn-cognition-engine` ran concept similarity searches against `ioc-knowledge-memory-svc` to build its evidence context for the `/decide` call. Every `concepts/similarity-search` request returned **500** after 3 retries × 7s each (~21s per attempt).

3. `shared-memories/query` requests sent at 19:47:39 didn't complete until 19:49:01 — **82 seconds** — all 500s. The CE was blocked waiting on evidence retrieval.

4. The CE spent the entire 300s session budget in this retry loop and never returned a `/decide` response.

5. `ioc-cfn-svc` closed the hung HTTP connection → `RemoteProtocolError: Server disconnected without sending a response` at backend side.

**Why R0–R6 succeeded:** Earlier rounds didn't trigger the blocking evidence path (RAG similarity-search fast-failed with 404; CE fell through to degraded mode). R7 hit a code path that needed concept similarity results and blocked on the retry loop.

**Fix:** `CREATE EXTENSION IF NOT EXISTS vector;` in the `ioc-knowledge-memory-svc` DB, or guarantee pgvector is present at init. `mycelium doctor` doesn't currently check for this extension.

Gist updated: https://gist.github.com/juliarvalenti/7df4709a82d7faf8b1f5ae711b264741

juliarvalenti:
Thanks @akanksha276 — that fresh-install pass was exactly what we needed, and the headline is the one that matters: **the pgvector fix (`36dfdad`) is confirmed working end-to-end** in your run (consensus reached, `broken:false` / `reason:agreed` / `cfn_persisted:true`, plan compiled). The original exp-0568 root cause is resolved.

Fixed the packaging gaps you found around it — all on `feat/l9-integration`:

**Gap 3 (cfn-db-init DOA — the regression I introduced)** → fixed in `1584133`. `_refresh_compose_templates()` now materializes `cfn-db-prep.sh` into `~/.mycelium/docker/` alongside `compose.yml`, so Docker bind-mounts a real file instead of creating an empty directory. Verified it writes a 2904-byte `0755` file; `cfn-db-init` runs instead of exit-126'ing.

**Gap 1 (`:latest` db without TimescaleDB)** → fixed in `1584133`. Added `mycelium install --tag <version>`, which pins `MYCELIUM_IMAGE_TAG` in `.env` (operator-managed, survives `config apply`) on both the fresh and re-install paths. So `mycelium install --tag 2.0.0rc3 --ioc` brings up a TimescaleDB-capable db without a separate `mycelium pull --version`.

**Gap 2 (`COGNITION_FABRIC_NODE_URL` silent-tick)** → already resolved on this branch. Current `coordination.py:_run_tick` guards on `settings.CFN_SVC_URL` (the var was renamed back in `ca2f199`), which the installer writes and `doctor` checks — there's no `COGNITION_FABRIC_NODE_URL` anywhere in current source. Your rc2 run did get ticks firing to consensus, so this one is carried forward from the rc1 comment; no change needed.

**Bonus** → `64d59a5` adds a `mycelium doctor` check for the `vector` extension in `ioc-graph-db`, so the exact failure mode that kicked this off (missing pgvector → `broken:true`) is caught up front with a `CREATE EXTENSION` remediation.

**Gap 4** (`mycelium pull --version` before `install`) → parked pending a clean re-repro of the intended first-time-setup ordering.

If you can cut an **rc3** off the current branch tip (`64d59a5`) and re-run the fresh-install → negotiation path, that should exercise all three fixes together with no manual `psql`/`.env` steps. Thanks again — genuinely high-signal report.

juliarvalenti:
## L9 payload faithfulness: audit vs deployed CFN + `ioc-protocols-models` drift

Checked our L9 envelope construction against (a) the CFN version this PR actually deploys and (b) `ioc-protocols-models` `main`, both freshly pulled locally.

**TL;DR:** Our envelope is faithful to the CFN we deploy. Nothing here blocks the PR. But the spec repo has moved 3 commits ahead of our vendored binding, and one of them (a commit-subkind rename) is worth tracking because we sit on the soon-to-be-old name.

### Faithful to the deployed CFN

Deployed stack: `ioc-cfn-svc` pinned `2026-06-26-dd16901`, which pins `ioc-protocols-models/.../golang v0.0.6` (its `go.mod`).

- `VALID_SUBKINDS` in `l9.py` matches the Go `validSubkinds` in `pkg/app/handlers_l9.go` exactly: `knowledge {query, distillation, extraction, feedback}`, `commit {converged, resolved, abort}`, `intent {coordinator-assignment, mission}`, `exchange {team-formation}`, `contingency {negotiation}`.
- `VERSION = "0.0.6"` matches the CFN's pinned golang binding.
- Every binding-required field is populated: header `{protocol, subprotocol, version, kind, participants, message}`; `message {id, parents, episode}`; `participants {actors, groups}` (groups is required-but-nullable, and `envelope_to_dict` correctly restores an explicit null); payload `{type, data}`.
- Optional fields we deliberately omit, none required by the binding or by CFN routing: `header.policy`, `header.attributes`, `context.epistemic`, `context.semantic`, `actor.attestation`. So there are no missing required "flags"; the CFN routes on kind/subkind + `participants.groups` only.

### Drift: our vendored `93773a5` to spec `main` (`c63f8c5`)

Three commits, newest first:

| PR | Change upstream | Impact on us |
|----|-----------------|--------------|
| outshift-open/ioc-protocols-models#32 | Commit subkinds renamed `converged` to `resolved`, `disagreement` to `unresolved` (in the SAB subprotocol) | **Track this.** We emit `commit:converged` on consensus (`l9_episode.py:230`). The deployed CFN still accepts `converged` (both `converged` and `resolved` are in `handlers_l9.go`), so no break today. When the CFN bumps its binding past this rename, `converged` envelopes will 400 and we switch to `resolved`. |
| outshift-open/ioc-protocols-models#29 | `message` made **required** on `L9Header`; canonical python L9 classes moved from `language_bindings/python/.../data_model.py` to `src/l9.py` | Safe. We always populate `message`, so the now-required field is covered. Two stale bits to tidy: our vendored `l9_models.py` still types `message` as optional, and its regen header points at the old `data_model.py` path (now `src/l9.py`). |
| outshift-open/ioc-protocols-models#28 | SAB drops `SABHeader`, builds on the canonical L9 envelope | Future. This is the "negotiation migrates onto the L9/SAB layer" path CLAUDE.md already flags as transitional. Not consumed yet. |

Also: a golang binding `v0.0.7` now exists, but the CFN still pins `v0.0.6`, so tracking the deployed version rather than spec HEAD is the correct call.

### Suggested low-cost follow-ups (not blockers)

1. Drop a `TODO` near `l9_episode.py:230` noting `converged` is the pre-rename name (see outshift-open/ioc-protocols-models#32) and becomes `resolved` once the CFN adopts the renamed binding.
2. Fix the regen path in `l9_models.py`'s vendoring header (`data_model.py` to `src/l9.py`).

_Audit run against local checkouts: `ioc-protocols-models` @ `c63f8c5`, `ioc-cfn-svc` @ `7c55a85` (main), deploy pin `2026-06-26-dd16901`._

juliarvalenti:
## L9 spec coverage: what's implemented vs. what "implement the L9 spec" still entails

Follow-up to the drift audit above. This maps mycelium's current L9 epistemic implementation against the formal spec (`ioc-protocols-models` `SSTP/documentation/L9-formal.md` + `L9-old.md`, the SNP/SIEP layer) so there's a concrete gap list to work from.

Correction to an earlier take in this thread: MPC / GAR / SCR / `provenance_weight` are **not** a mycelium invention. They are formal L9 spec primitives (defined with exact math in `L9-formal.md`); they just live in the spec's documentation + `hcpanel` reference code, not in the python/golang bindings or the CE service, which is why a code-only grep missed them.

### Faithful today (do not rebuild)

- Envelope scaffolding: `intent:mission` at open, `exchange` for ticks + synthesized replies, `commit:converged`/`abort` at close, with causal `parents` and episode URNs.
- Three of the four metrics match the spec formula exactly: **MPC**, **GAR**, **`provenance_weight` = (1 - SCR) x GAR**.
- A thin belief slice collected from agents: `confidence` (used as both prior and posterior), flat `evidence`, `reasoning`, `deferred_to`.

### Missing: the belief-revision substrate the metrics are supposed to sit on

| Spec primitive | Spec location | Mycelium today |
|---|---|---|
| `prior` / `posterior` as distinct per-turn beliefs | `belief` block | collapsed into one `confidence`; prior = first-seen |
| `uncertainty` (= 1 - posterior) | epistemic block | absent (derivable) |
| `revision_cause` (grounded_argument / social_compliance / semantic_memory / new_evidence / repair_resolution) | `belief` block | **absent** (this is the foundation SCR needs) |
| `belief_status` (asserted / deferred / challenged / revised / retracted / unresolved) | epistemic block | absent |
| `message_act` (assertion / compliance / challenge) | epistemic block | absent (inferred from action) |
| `supporting` / `against` / `addresses_evidence` (3-way) | `proposal_payload` | flat `evidence` list only |
| `contingency_score` (grounding test) | derived metric | not computed |
| **SCR** = fraction of revisions with `revision_cause = social_compliance` | derived from ToM | **proxy**: accepts-with-`deferred_to` (`l9_episode.py:206`) |
| Theory-of-Mind drift detection | source of SCR/GAR raw signal | absent |
| `knowledge` `rule_update` writeback (posterior/gar/scr/provenance_weight/revision_cause to SemanticMemory) | post-commit | metrics stay internal (episode md + consensus payload); no spec knowledge envelope |
| `snp-convergence` payload type on commit | commit envelope | metrics ride an ad-hoc consensus field |

Through-line: without `revision_cause` + grounding, **SCR can only be a proxy, and `provenance_weight` is only as real as SCR.** That is the load-bearing gap.

### Proposed plan (4 phases, dependency order)

**Phase 1 - speak the belief model (data layer).** Extend the agent reply so agents volunteer the SNP/SIEP fields; carry them in envelopes as real `cip`/`siep` belief + epistemic blocks instead of the ad-hoc `reply` dict.
- `protocol.py`: add `posterior` (alias `confidence`), `belief_status`, `message_act`, `revision_cause`; split evidence into `supporting`/`against`/`addresses`.
- `commands/negotiate.py` + the 3 `SKILL.md` files: new flags + protocol guidance so agents populate them.
- `cfn_negotiation.py:sanitize_epistemic_fields` + `l9_episode.record_reply`: fold into the belief block; `l9.build_envelope` payloads become spec-shaped.

**Phase 2 - derive grounding + real SCR (correctness upgrade).**
- Compute `contingency_score = |addresses_evidence intersect prior evidence| / |prior evidence|`; flag turns grounded/ungrounded.
- Derive `revision_cause`: posterior moved + grounded -> `grounded_argument`; moved + ungrounded/weak-evidence -> `social_compliance`.
- Replace the SCR proxy at `l9_episode.py:206` with the spec definition (fraction of `social_compliance` revisions). Highest-value single change: makes SCR and `provenance_weight` mean what the spec says.
- Fix the MPC=0.5 GAR degeneracy while here (spec shares it, but we can guard it).

**Phase 3 - Theory-of-Mind (hard, ambitious).** Per-agent predictive model + drift detection, the spec's actual source for genuine-vs-compliance. Real AI-modeling effort, and arguably the Cognition Engine's job. Recommend **defer**; let Phase 2's heuristic `revision_cause` stand in.

**Phase 4 - close the loop (knowledge writeback).**
- At consensus emit the spec `knowledge` `rule_update` envelope (posterior=MPC, gar, scr, provenance_weight, revision_cause=`converged_episode`) to the knowledge store; add `snp-convergence` payload on the commit.
- Makes `provenance_weight` usable: the `team_prior`-on-ticks path (`l9_cfn.knowledge_query_team_prior`, already stubbed) reads a concept's prior weight into the next episode. Today we compute `provenance_weight` and drop it.

### Scoping caveat

Nothing in the **deployed** CFN/CE consumes SNP/SIEP payloads yet (zero refs in `ioc-cfn-svc` / `ioc-cfn-cognition-engines`). So Phases 1, 2, 4 make mycelium's own records/metrics/priors genuinely spec-faithful and self-consuming (UI, episode history, provenance-weighted priors), which is real value today; Phase 3 and CE-native metric computation are ahead of the platform. Since negotiation is slated to migrate onto the L9/SAB layer natively, Phases 1-2-4 are the non-throwaway work because that migration needs these belief semantics regardless.

Recommended deliverable for this branch: **Phase 1 + Phase 2 + the Phase 4 writeback**, with ToM as a documented follow-up.

juliarvalenti:
## L9 spec coverage: the CFN (`ioc-cfn-svc`) side

Companion to the mycelium gap analysis above, from the CFN's perspective. Audited against `ioc-cfn-svc` main (`7c55a85`, freshly pulled) and `ioc-cfn-cognition-engines` main (`a58dbab`).

**Framing that makes the rest make sense:** the CFN is a **routing/transport fabric + coordination mechanism**, not the epistemic compute engine. `l9Handler` (`pkg/app/handlers_l9.go`) parses the envelope, validates kind/subkind, extracts routing IDs, and forwards verbatim to the matching Cognition Engine (CE). All belief semantics are delegated to CEs.

### Implemented by the CFN

- **L9 envelope transport + validation** (`POST /api/l9/messages`): parses the full `l9.L9` (golang binding `v0.0.6`), validates kind/subkind against the authoritative `validSubkinds` table, content-routes by `participants.groups` `{workspace_id, mas_id}` + kind/subkind/subprotocol to the matching CE, forwards verbatim. This is the routing layer of the L9 header.
- **All 5 kinds as routing categories**: `intent`, `contingency`, `exchange`, `commit`, `knowledge` are validated and routed (not semantically processed).
- **Header policy + provenance blocks**: modeled and passed through (`sstpPolicyLabels` = sensitivity/propagation/retention, `sstpProvenance` = sources/transforms).
- **Semantic alignment / SAO bargaining** (`/semantic-alignment/start|decide`): runs the Stacked Alternating Offers mechanism (rounds, offers, accept/reject/counter_offer decisions, final agreement, `timedout`/`broken`). Emits a commit-level `confidence_score` scalar.
- **Knowledge / shared memory**: extraction (concepts/relations/RAG), `shared-memories` write/query, graph store (neighbors/paths/similarity), vector store. Persists agreements to shared memory as format `semneg`.
- **CE registry + operational metrics**: `cognition-engines` CRUD, heartbeat, token/latency/cost metrics.

### NOT implemented by the CFN (delegated to CEs, or absent everywhere)

- **SNP/SIEP epistemic compute**: no `belief`, `prior`, `posterior`, `uncertainty`, `revision_cause`, `belief_status`, `message_act`. Zero code refs.
- **Convergence metrics MPC / GAR / SCR / `provenance_weight`**: not computed anywhere in the CFN. Its only score is the commit `confidence_score` scalar, produced by the bargaining layer, not derived from a belief model.
- **Grounding / `contingency_score`**: absent.
- **Theory-of-Mind drift detection**: absent.
- **Epistemic `rule_update` writeback**: it persists the agreement (`semneg`), but not an SNP knowledge rule-update carrying `gar`/`scr`/`provenance_weight`/`revision_cause`.
- **CIP / SIEP subprotocol semantics**: routes by subprotocol but implements none; negotiation is SAO/SAB bargaining via the CE.
- **Causal threading semantics**: `message.parents` + `episode` are required binding fields, parsed and transported, but routing ignores them (kind/subkind/groups only). Carried, not acted on.
- **`context.epistemic` / `context.semantic`**: present in the binding; docs claim `Semantic` drives CE selection, but the code routes on kind/subkind/subprotocol via the `KindsSubkinds` map. Transported, effectively unused for routing.

### The key takeaway

**Neither the CFN nor mycelium currently computes the SNP epistemic metrics from a real belief model.** The CFN owns transport/routing + SAO bargaining + knowledge plumbing; the epistemic-semantic layer (belief revision, MPC/GAR/SCR, grounding, ToM) is delegated to CEs. The deployed CE (`ioc-cfn-cognition-engines`) computes agreement (NegMAS/SAO + `confidence_score`) but also does not compute the SNP convergence metrics (zero refs there too).

That is exactly why mycelium computes MPC/GAR/SCR itself (`l9_episode.py`, "interim until the CE computes them natively"). The "interim" is real: **no deployed component computes them today.** So the Phase 1-2-4 work proposed in the mycelium comment above is not duplicating the CFN; it is filling a layer the whole deployed stack currently leaves empty.

juliarvalenti:
## L9 protocol styles in the Cognition Engine (`ioc-cfn-cognition-engines`): consumed vs. latest

Final piece of the audit, and it comes with a **correction to my CFN-side comment above**. That comment said "zero refs in `ioc-cfn-cognition-engines`" for the epistemic layer. That grep ran against a **stale local `main` (`a58dbab`)**. Re-checked against `origin/main` (`f673c15`, outshift-open/ioc-cfn-cognition-engines#39) and against the exact commit we deploy. The corrected picture below.

### Two L9 "styles" in play

- **SSTP style (legacy):** raw `SSTPNegotiateMessage` over REST `/negotiate/*`. This is what mycelium's typed client (`cfn_negotiation.py`) speaks.
- **L9/SAB style (new):** L9 envelopes over `POST /api/l9/messages`, SAB-wrapped at the CFN<->CE boundary.

### What we actually consume: image `2026-06-26-c01c60a` = commit `c01c60a` (outshift-open/ioc-cfn-cognition-engines#34)

- **No L9 protocol endpoint.** `semantic_alignment/app/api/l9_routes.py` = **ABSENT**; `protocol/sab/l9_adapter.py` = **ABSENT**. The consumed CE speaks legacy SSTP REST only. No `/api/l9/messages`, no SAB envelope wrapping.
- **But it does have an epistemic-validation layer (SAV = Semantic Alignment Validation).** `semantic_alignment/app/agent/sav/` + `semantic_alignment_validation_pipeline.py`, plus a standalone `semantic_validation/` CE. Its `models.py` computes instability/drift signals: `InteractionSignals` (positional_instability, per_issue_divergence, oscillation_rate, overall_instability_score), `AlignmentEvaluation` (agreement_coherence, cross_issue_conflicts), Severity LOW/MEDIUM/HIGH against a `GoalSpec` (mission_constraints, success_criteria).

So my "zero epistemic anything in the CE" was too strong. **Correction:** even the consumed CE has a drift/alignment-quality validation layer. What it is *not* is the SNP belief-revision metric family: there is no `posterior`/`prior`/`belief`/`revision_cause`, and no MPC/GAR/SCR/`provenance_weight`. SAV measures *instability of the bargaining trace*; it does not model agent beliefs or compute the convergence metrics. Those remain uncomputed by any deployed component, which is the point that stands.

### Latest CE main (`f673c15`)

- **L9 endpoint now PRESENT:** `l9_routes.py` + `protocol/sab/l9_adapter.py` (merged outshift-open/ioc-cfn-cognition-engines#38 "SAB L9 routing for Semantic Alignment CE", outshift-open/ioc-cfn-cognition-engines#35 per-MAS config wire-in). The CFN<->CE boundary is L9/SAB-wrapped as of ~4 days ago.
- SAV validation layer still present, same instability-based signals.

### Still open upstream (the belief layer is being actively built, not merged)

- **outshift-open/ioc-cfn-cognition-engines#33** (open): renames SAV -> **Drift Detection**, adds a CE-side L9 adapter that reads `header.attributes['drift']`, `header.context.topic`, and `header.message` (episode/parents) to build episode-lineage interaction history, plus 8 per-MAS process-failure-mode thresholds. Note: its backend-retrieval helpers are **mocked** ("real backend doesn't exist yet across `ioc-cfn-svc`/mgmt/fabric-node").
- **outshift-open/ioc-cfn-cognition-engines#41** (open, ~1h): SAB-wrap negotiation **inward** (L9 end-to-end to agents).
- **outshift-open/ioc-protocols-models#25** (open, ~1h): **genuine GAR/SCR metrics** + grounding in the reference hcpanel, plus a new `SSTP/l9/` module (deliver/emit/episode/grounding). This is the canonical reference implementation of exactly the belief-revision + grounding + real-GAR/SCR layer.

### Bottom line for this branch

1. We consume a CE (`c01c60a`) that is **SSTP-style only**: no `/api/l9/messages`, no SAB. mycelium's optional native-L9-to-CFN path (`L9_CFN_ENABLED` -> `l9_cfn.post_l9`) would route through the CFN to a CE with no L9 handler, hitting the CFN's "CE-not-found -> fallback URL" TODO. **This is exactly why `L9_CFN_ENABLED` is off by default, and why it can't be turned on until the CE image is bumped past the outshift-open/ioc-cfn-cognition-engines#38 merge (newer than the `2026-06-26` pin).** mycelium's additive L9 envelopes inside coordination content are unaffected; they never hit the CE.
2. The epistemic layer I described as "empty" is more precisely: **the deployed stack has SAV (instability/drift validation) but not the SNP belief-revision metrics.** The belief model + genuine GAR/SCR + grounding + drift-as-social-compliance are in-flight upstream (outshift-open/ioc-protocols-models#25, outshift-open/ioc-cfn-cognition-engines#33, outshift-open/ioc-cfn-cognition-engines#41), not released.
3. Mapping to the mycelium plan: **Phase 3 (ToM/drift) maps directly onto the upstream Drift Detection CE (outshift-open/ioc-cfn-cognition-engines#33)** — mycelium should consume that once released rather than build its own. Phases 1-2-4 (belief model, genuine SCR, knowledge writeback) remain mycelium's to own, and align with outshift-open/ioc-protocols-models#25's direction.

juliarvalenti:
## Minimal L9 spec-adoption suggestion, expressed as CLI changes

Following up on the phased gap analysis above — here's one way to land Phases 1/2/4 as small, additive changes to the existing `negotiate` commands rather than new subsystems. Keeping it scoped to flags/output on commands that already exist, consistent with the additive/fail-soft pattern already in this PR.

### 1. `negotiate propose` — split evidence, get prior for free

```
mycelium negotiate propose budget=high scope=full \
  --confidence 0.8 \
  --supporting-evidence "vendor_quote,timeline_data" \
  --against-evidence "budget_precedent_q3" \
  --addresses "prior_offer_id_or_evidence_key" \
  --reasoning "..."
```

- `--evidence` (flat) → split into `--supporting-evidence` / `--against-evidence`, matching the spec's 3-way `proposal_payload` shape.
- `--addresses` is new: names which prior evidence keys this offer engages — the one field the grounding check actually needs.
- On the **first** `propose` in an episode, store that confidence as the immutable prior (never overwritten by later posteriors) — no new flag required, just: first propose = prior, everything after = posterior. Cheaper than an explicit `declare-prior` subcommand and gets you prior/posterior separation for free.

### 2. `negotiate respond` — declare *why* you moved

```
mycelium negotiate respond accept \
  --revision-cause grounded_argument \
  --addresses "vendor_quote,timeline_data"
```

- `--revision-cause` takes the spec enum: `grounded_argument | social_compliance | new_evidence | semantic_memory | repair_resolution`.
- If omitted, derive a fallback from grounding: `contingency_score = |--addresses ∩ prior's --supporting-evidence| / max(len(addresses), len(prior_evidence))`; score ≥ 0.40 → `grounded_argument`, else → `social_compliance`. Gives a real signal even when the agent doesn't self-report, and having both a declared cause *and* a computed score is more robust than deriving cause from grounding alone.
- `--defer-to` becomes sugar for `--revision-cause social_compliance` — keep both one release, then fold.

This is the load-bearing change: without a real `revision_cause`, SCR stays a proxy and `provenance_weight` is only as real as SCR.

### 3. `negotiate status` — surface what's already computed and currently dropped

```
mycelium negotiate status --room sprint-plan

Round 3/5 · 4/5 agents replied
Standing offer: budget=high, scope=full

L9 metrics (interim, episode-scoped):
  MPC: 0.68        GAR: 0.72        SCR: 0.22
  provenance_weight: 0.56  [CONTESTED — below 0.60 threshold]
  ungrounded turns: 1 (agent: reviewer-2, round 2)
```

Purely surfacing values `l9_episode.py` already computes and currently discards. Cheapest possible consumer.

### 4. `negotiate status --contested` — the minimal control-flow hook for `provenance_weight`

```
mycelium negotiate status --room sprint-plan --contested
```

Exits non-zero if `provenance_weight < 0.60` at commit. Usable in CI/orchestration to gate on "was this negotiation actually trustworthy" before an agent acts on the outcome — without building a full re-negotiation/dual-episode mechanism. Gives the metric a consumer, which today it doesn't have.

### 5. `memory` — knowledge writeback via the existing command

```
mycelium memory set "l9/rule_update/<concept_id>" \
  '{"posterior": 0.68, "gar": 0.72, "scr": 0.22, "provenance_weight": 0.56, "revision_cause": "converged_episode", "episode_id": "<id>"}' \
  --room sprint-plan
```

No new command — the negotiation-commit path calls existing `memory set` under a `l9/rule_update/*` key convention. This is what gives `knowledge_query_team_prior` (already stubbed per the audit above) something real to read via `memory get` on the next episode, closing the loop for `team_prior`-on-ticks.

### Deliberately out of scope

No ToM CLI surface, no predictive belief-model subcommand — matches the deferral already proposed for Phase 3. Consume the upstream Drift Detection CE once released rather than building a CLI for it now.

### Suggested order

1. `--revision-cause` on `respond` — single flag, highest leverage
2. `--supporting-evidence` / `--against-evidence` / `--addresses` on `propose` — needed for #1's fallback derivation
3. `status` output extension — free, no new flags
4. `--contested` exit-code flag — gives `provenance_weight` a consumer
5. `memory` writeback convention — closes the loop

Each step is additive to an existing command; no new subcommand tree, no schema migration.

PR #393, opened 2026-07-02
feat: first-class typed `event` message primitive (#392)

### LINKED ISSUE #392: feat: first-class typed `event` message primitive (source_event, action, concern)
## Summary

Add **one** first-class structured message type — `event` — to Mycelium rooms, discriminated by a `kind` field and carrying a structured `metadata` payload. A single general primitive covers a live source-activity feed and a persistent, status-bearing action ledger — without adding a new bespoke message type per use case.

Kinds in scope: `source_event`, `action`, `concern`.

## Why one primitive, not a type per use case

These kinds differ along **two independent axes** — persistence and statefulness — not one:

| Kind           | Persistence         | Stateful?             |
|----------------|---------------------|-----------------------|
| `source_event` | stored, capped/TTL  | no                    |
| `action`       | durable             | yes (open→resolved)   |
| `concern`      | durable             | yes                   |

Modeling **retention** and **status** as message *attributes* (not as distinct message types) lets one `event` type serve all three today and anything future (`note`, `decision`, `ci_result`, …) without re-litigating storage/filtering/rendering each time.

### Message shape

```json
{
  "message_type": "event",
  "room_name": "platform-eng",
  "sender_handle": "github-poller",
  "content": "New PR: \"fix recordings window\" (#48)",
  "metadata": {
    "kind": "source_event",
    "ttl_seconds": 1209600,
    "status": null,
    "payload": { "source": "github", "event": "pr_opened", "number": 48, "repo": "org/repo" },
    "provenance": [
      { "type": "pr", "ref": "org/repo#48", "url": "https://github.com/org/repo/pull/48" }
    ],
    "correlation_id": "…"
  }
}
```

- `content` — human-readable label (what a client renders if it doesn't understand the kind).
- `metadata.kind` — open string vocabulary (`source_event`, `action`, `concern`, …). Indexed, queryable.
- `metadata.ttl_seconds` — optional retention cap for transient kinds (`source_event`). Absent = durable (`action`/`concern`).
- `metadata.status` — for stateful kinds: `open | in_progress | resolved` (null for stateless). Backs the ledger.
- `metadata.payload` — kind-specific structured data.
- `metadata.provenance` — array of cited refs (`pr | commit | issue | page | message`), each `ref` + optional `url`.
- `metadata.correlation_id` — groups related events / links a status update to its origin.

### Behavior

- **`POST /rooms/{name}/messages`** accepts `message_type: "event"` with a `metadata.kind`. Stored; broadcast over SSE.
- **`GET /rooms/{name}/messages`** gains optional filters: `?kind=source_event`, `?status=open`, `?since=<ts>`. Feed and ledger become **server-side filters over the room**, not client-side scans.
- **Status transitions** (stateful kinds): move an `action`/`concern` open→resolved. *Open design question for the implementer:* a follow-up `event` referencing the original `correlation_id`, or a `PATCH /messages/{id}` — pick one; the current status must be queryable.
- **Retention:** a sweep (or TTL index) drops stored events past `ttl_seconds`. Durable kinds (no TTL) are never swept.
- **SSE (`room-sse.ts`):** emit `event` frames to subscribers; clients filter/render by `kind` (feed rows, ledger cards) or ignore. No change to existing `broadcast`/`announce`/`direct`/`delegate` subscribers.

### Filtering & indexing

- Index `(room_name, message_type, metadata.kind, created_at)` and `(room_name, metadata.kind, metadata.status)` so feed/ledger queries stay cheap.

## Acceptance criteria

- [ ] `POST /rooms/{name}/messages` accepts `message_type: "event"` with a `metadata.kind`
- [ ] Stored events respect `ttl_seconds` (capped kinds) or persist durably (no TTL)
- [ ] `GET /rooms/{name}/messages` supports `?kind=`, `?status=`, `?since=` filters
- [ ] Stateful kinds (`action`/`concern`) carry a `status` that can transition and be queried
- [ ] `metadata.provenance` is accepted and returned intact
- [ ] `room-sse.ts` emits `event` frames to subscribers; existing message types unchanged
- [ ] Docs: the `kind` vocabulary and the ttl/status/provenance contract

juliarvalenti:
Closes #392.

One structured message type — `event` — discriminated by `metadata.kind`, with retention (`ttl_seconds`) and statefulness (`status`) as message **attributes**, per the issue's one-primitive design. Kinds in scope: `source_event` (transient, stateless), `action` / `concern` (durable, `open → in_progress → resolved`). Unknown kinds are accepted with sane defaults, so future uses need no schema change.

## Implementation notes

- **Storage**: `messages` gains `metadata` (JSONB, returned intact) plus `event_kind`/`event_status`/`event_expires_at` **promoted at write time** — feed/ledger filters and the TTL sweep hit plain indexed columns instead of JSON path expressions (which differ between AgensGraph and the SQLite test backend). Migration `0018` adds the two composite indexes from the issue. All columns NULL for non-event types; existing rows/writers untouched.
- **Status transitions** — the issue left the design open (follow-up event vs PATCH). Went with **in-place `PATCH /rooms/{name}/messages/{id}`** `{"status": "resolved"}`: the current status must be cheaply queryable, and a follow-up-event model forces a latest-per-correlation aggregation on every ledger query. The transition is broadcast over the room SSE stream (payload carries `status_transition`). 409 for stateless kinds / non-events, 404 across rooms.
- **TTL**: `event_expires_at` precomputed at write; a lifespan-managed sweep (5 min) reclaims expired rows, and GET independently excludes expired-but-unswept events so sweep latency is invisible to readers. No TTL = durable, never swept.
- **SSE**: NOTIFY payloads include `metadata` for events; `room-sse.ts` is a transparent pump so subscribers get event frames as-is and filter/render by kind or ignore. Existing message types unchanged.
- **Knowledge ingest**: events are excluded from CFN fan-in — they're machine feed, not agent speech.
- **Docs**: "Typed events" section in the Rooms concept doc (kind vocabulary + ttl/status/provenance contract); site regenerated. `openapi.json` snapshot and vendored typed clients regenerated (CI staleness check will pass).

## Acceptance criteria

- [x] `POST /rooms/{name}/messages` accepts `message_type: "event"` with `metadata.kind`
- [x] `ttl_seconds` respected (sweep + read-time exclusion); no-TTL kinds durable
- [x] `GET` supports `?kind=`, `?status=`, `?since=`
- [x] `action`/`concern` carry transitionable, queryable status
- [x] `provenance` accepted and returned intact
- [x] `room-sse.ts` emits event frames; existing types unchanged
- [x] Docs: kind vocabulary + ttl/status/provenance contract

## Testing

16 new tests (validation incl. source_event-rejects-status and metadata-only-on-event, kind defaults, ttl expiry column, filters, expired-hidden-from-GET, transitions with 409/404 paths, sweep deletes-only-expired). Backend suite 293 passed, CLI 404 passed, ruff + ty clean on both.

### DIFF
--- a/fastapi-backend/alembic_migrations/versions/0018_event_message_primitive.py
+++ b/fastapi-backend/alembic_migrations/versions/0018_event_message_primitive.py
@@ -1,0 +1,67 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2026 Mycelium Contributors
+
+"""Add the typed ``event`` message primitive (#392).
+
+Four new nullable columns on ``messages``:
+
+- ``metadata``          — full structured event metadata (kind, ttl_seconds,
+                          status, payload, provenance, correlation_id),
+                          returned intact to clients.
+- ``event_kind``        — promoted from metadata.kind at write time.
+- ``event_status``      — promoted from metadata.status; backs the ledger.
+- ``event_expires_at``  — created_at + ttl_seconds, precomputed for the TTL
+                          sweep. NULL = durable.
+
+Plus the two composite indexes that keep feed (?kind=) and ledger (?status=)
+queries cheap. All columns are NULL for non-event message types, so existing
+rows and writers are unaffected.
+
+Revision ID: 0018
+Revises: 0017
+"""
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects.postgresql import JSONB
+
+revision = "0018"
+down_revision = "0017"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    with op.batch_alter_table("messages") as batch:
+        batch.add_column(
+            sa.Column(
+                "metadata",
+                JSONB().with_variant(sa.JSON(), "sqlite"),
+                nullable=True,
+            )
+        )
+        batch.add_column(sa.Column("event_kind", sa.String(), nullable=True))
+        batch.add_column(sa.Column("event_status", sa.String(), nullable=True))
+        batch.add_column(sa.Column("event_expires_at", sa.DateTime(timezone=True), nullable=True))
+    op.create_index("ix_messages_event_expires_at", "messages", ["event_expires_at"])
+    op.create_index(
+        "ix_messages_room_kind_created",
+        "messages",
+        ["room_name", "event_kind", "created_at"],
+    )
+    op.create_index(
+        "ix_messages_room_kind_status",
+        "messages",
+        ["room_name", "event_kind", "event_status"],
+    )
+
+
+def downgrade() -> None:
+    op.drop_index("ix_messages_room_kind_status", table_name="messages")
+    op.drop_index("ix_messages_room_kind_created", table_name="messages")
+    op.drop_index("ix_messages_event_expires_at", table_name="messages")
+    with op.batch_alter_table("messages") as batch:
+        batch.drop_column("event_expires_at")
+        batch.drop_column("event_status")
+        batch.drop_column("event_kind")
+        batch.drop_column("metadata")
--- a/fastapi-backend/alembic_migrations/versions/0018_event_message_primitive.py
+++ b/fastapi-backend/alembic_migrations/versions/0018_event_message_primitive.py
@@ -43,17 +43,22 @@ def upgrade() -> None:
         batch.add_column(sa.Column("event_kind", sa.String(), nullable=True))
         batch.add_column(sa.Column("event_status", sa.String(), nullable=True))
         batch.add_column(sa.Column("event_expires_at", sa.DateTime(timezone=True), nullable=True))
-    op.create_index("ix_messages_event_expires_at", "messages", ["event_expires_at"])
-    op.create_index(
-        "ix_messages_room_kind_created",
-        "messages",
-        ["room_name", "event_kind", "created_at"],
-    )
-    op.create_index(
-        "ix_messages_room_kind_status",
-        "messages",
-        ["room_name", "event_kind", "event_status"],
-    )
+    # Plain CREATE INDEX takes an ACCESS EXCLUSIVE lock on messages for the
+    # build — on a busy deployment that blocks every room's writes. Build
+    # concurrently on postgres (needs autocommit); sqlite has no such lock
+    # semantics (or CONCURRENTLY support), so it takes the plain path.
+    indexes = [
+        ("ix_messages_event_expires_at", ["event_expires_at"]),
+        ("ix_messages_room_kind_created", ["room_name", "event_kind", "created_at"]),
+        ("ix_messages_room_kind_status", ["room_name", "event_kind", "event_status"]),
+    ]
+    if op.get_bind().dialect.name == "postgresql":
+        with op.get_context().autocommit_block():
+            for name, cols in indexes:
+                op.create_index(name, "messages", cols, postgresql_concurrently=True)
+    else:
+        for name, cols in indexes:
+            op.create_index(name, "messages", cols)
 
 
 def downgrade() -> None:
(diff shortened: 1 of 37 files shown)


fastapi-backend/app/routes/messages.py
```diff
@@ -182,3 +229,78 @@ async def list_messages(
         messages=[MessageRead.model_validate(m) for m in messages],
         total=len(messages),
```
juliarvalenti: `total` here is the returned page length (<= `limit`), not the total rows matching the filter. Clients that paginate on `total` break — the Tome Talk panel does reverse infinite-scroll comparing loaded-count vs `total`, so "load older" would stop early. Suggest a separate `select(func.count())` with the same `where` clauses (including the expiry exclusion) for the real total. If this predates the PR, still worth fixing while the file is open.

juliarvalenti (commented):
Reviewed the primitive end-to-end — design matches #392 and the tradeoffs are well-chosen (in-place PATCH over follow-up events; promoting kind/status/expires_at to indexed columns to dodge JSON-path differences; read-time expiry exclusion so sweep lag is invisible; excluding events from CFN fan-in and coordination response-parsing). Three things worth a look, left inline. Only the first is functional, and it may predate this PR.

fastapi-backend/app/routes/messages.py
```diff
@@ -182,3 +229,78 @@ async def list_messages(
...
+        "metadata": msg.event_metadata,
+        "created_at": msg.created_at.isoformat(),
+        "status_transition": payload.status,
+    }
+    channel_name = room.name if room is not None else coord.display_name  # type: ignore[union-attr]
+    notify_payload["room_name"] = channel_name
+    try:
+        from urllib.parse import urlparse
+
+        parsed = urlparse(settings.DATABASE_URL)
```
juliarvalenti: This opens a brand-new asyncpg connection (parse URL + connect + auth + close) per message just to fire NOTIFY, and the block is copy-pasted verbatim from `send_message`. Human chat is low-volume so it was tolerable, but `source_event` is driven by a poller emitting bursts — this connection churn becomes a hot path. Suggest a pooled/long-lived NOTIFY connection and factoring the duplicated block into one `_notify_room(channel, payload)` helper used by both POST and PATCH.

fastapi-backend/alembic_migrations/versions/0018_event_message_primitive.py
```diff
@@ -0,0 +1,67 @@
...
+            sa.Column(
+                "metadata",
+                JSONB().with_variant(sa.JSON(), "sqlite"),
+                nullable=True,
+            )
+        )
+        batch.add_column(sa.Column("event_kind", sa.String(), nullable=True))
+        batch.add_column(sa.Column("event_status", sa.String(), nullable=True))
+        batch.add_column(sa.Column("event_expires_at", sa.DateTime(timezone=True), nullable=True))
+    op.create_index("ix_messages_event_expires_at", "messages", ["event_expires_at"])
```
juliarvalenti: `op.create_index` takes an ACCESS EXCLUSIVE lock on `messages` for the duration of each build, blocking writes to every room while it runs. The nullable column adds above are fine (metadata-only in PG11+); it is just these three index builds. On a large/busy messages table consider `CREATE INDEX CONCURRENTLY` (autocommit migration) or call out a maintenance window.

juliarvalenti:
All three addressed in c8b449d:

- **`total` = page length** — real bug, and yes it predates the PR (`total=len(messages)` was there before). Fixed properly: the count runs over the same filtered query (subquery + `func.count()`), so it includes the kind/status/since filters *and* the expiry exclusion. Regression test asserts `limit=1` returns one row with `total == 3`.
- **Per-message NOTIFY connect** — both POST and PATCH now go through one `_notify_room()` helper backed by a shared long-lived asyncpg connection: lazy connect, one retry on a fresh connection if the server dropped it, never raises into the write path, closed in lifespan shutdown. The duplicated block is gone.
- **Index locks** — migration 0018 now builds all three indexes with `CREATE INDEX CONCURRENTLY` inside an autocommit block on postgres; sqlite (no such lock semantics, no CONCURRENTLY) takes the plain path. No maintenance window needed.

Gate after fixes: backend 293 passed (17 event tests), ruff + ty clean.

PR #406, opened 2026-07-16
fix(session): scope await missed-tick scan to the handle's session (#405)

### LINKED ISSUE #405: session await replays a stale consensus from an older session in the same room
## Summary

`mycelium session await` returns a stale `coordination_consensus` from an
*older* session in the same room instead of waiting for the current
session's tick, once a room has more than one coordination session over
its lifetime. Since rooms are documented as persistent, reusable
namespaces, this is a normal usage pattern — not an edge case — and it
permanently breaks `await`-based polling for any room that gets a second
negotiation.

## Version

`2.0.0rc4`, commit `5c4eb9c68831a73e5dfc648439d54f1cfc2aeb71`.

## Repro

1. Complete a negotiation in room `R` (session `A`) — a
   `coordination_consensus` message is persisted for `A`.
2. Start a new session `B` in the same room `R` (`mycelium session join`
   for at least one agent), and let the CognitiveEngine fire a real tick
   for `B`.
3. Run `mycelium session await -H <handle> -r R`.

**Expected:** blocks until (or immediately returns, if already fired)
session `B`'s tick.

**Actual:** immediately returns session `A`'s old consensus, tagged
`"replayed": true` — even while `B` is actively `negotiating` with its
own real, unconsumed ticks (confirmed directly via the backend API:
`GET /api/coordination-sessions?parent_room=R` shows `B` in state
`negotiating`, and `GET /api/rooms/{B}/messages` shows live
`coordination_tick` messages for it).

This reproduced consistently across every subsequent `await` call for
`B` in my testing, well after `B`'s round-2 tick had actually posted.

## Root cause

`mycelium-cli/src/mycelium/commands/session.py`, `await_tick()`
(~lines 312-377). Before opening the SSE stream, it does a "missed tick"
pre-check: it fetches *every* coordination session under the room
(`GET /api/coordination-sessions?parent_room=<room>&limit=200`), then
scans each one's message history and returns on the **first**
`coordination_consensus` message found — across any of those sessions,
not just the one the handle actually joined:

```python
rooms_to_scan = [resolved_room]
coord_resp = http.get(
    f"{config.server.api_url}/api/coordination-sessions",
    params={"parent_room": resolved_room, "limit": 200},
)
if coord_resp.status_code == 200:
    for c in coord_resp.json():
        rooms_to_scan.append(c["display_name"])   # adds ALL sessions, old + new

for scan_room in rooms_to_scan:
    ...
    for msg in msgs:                                # newest-first
        if mtype == "coordination_consensus":
            ...
            return                                  # returns on first hit, any session
```

There's no filter for "is this the session my handle actually joined,"
and no recency check across sessions — just a scan that stops at the
first consensus it sees, in whatever order the coordination-sessions
endpoint returns them. Once any session in a room reaches `complete`,
its consensus permanently shadows every later session's ticks.

## Suggested fix

Scope the missed-tick scan to the specific session the handle is a
participant in (the backend already knows this from the join), rather
than every session under the parent room. At minimum, only consider a
`coordination_consensus`/`coordination_tick` message if its
`coordination_session_id` matches the session the handle actually
joined.

## Workaround

Drive the negotiation via `mycelium negotiate respond` /
`mycelium negotiate status` instead of `session await` — both correctly
resolve the room's currently-active session rather than scanning all
historical sessions, so they don't exhibit this bug.

## Impact

Doesn't block a single negotiation in a brand-new room (the common
quick-start path), but breaks the documented `session await` loop for
any room reused across multiple negotiations — which the CLI's own docs
describe as the expected usage ("rooms are persistent namespaces").

juliarvalenti:
Fixes #405.

## Root cause

`await_tick`'s missed-tick pre-check (`session.py`) built `rooms_to_scan = [parent_room, *every session under the room]` and the inner `coordination_consensus` branch `return`ed on the **first consensus in any scanned room, with no session- or handle-scoping** (the tick branch *is* handle-filtered; the consensus branch isn't). So once any session reached `complete`, its consensus permanently shadowed later sessions' live ticks.

Two verified leak vectors:

- **The parent room carries the consensus.** `_finish_cfn` posts the consensus session-scoped *and* mirrors it to the parent room for the chat channel. The parent room is scanned first, so an older session's mirrored consensus is returned before the current session is looked at.
- **Older sessions leak too.** Newest-first ordering doesn't save it: each round the CE addresses one agent, so the current session's latest tick is often for the *other* handle, fails the tick branch's handle filter, and the loop falls through to the older session's unconditional consensus branch.

## Fix

Scope the pre-check to the session **the handle actually joined** — filtered, not guessed.

The backend already records participation (the `participants` table, written by `join_room`), so this adds a `participant` filter to `GET /api/coordination-sessions` and has `session await` query `?parent_room=<room>&participant=<handle>&limit=1`. That deterministically returns the caller's own most-recent session; ticks/consensus are session-scoped, so scanning only that session's messages is complete (catches a missed tick for the handle, or its own session's consensus if it concluded between join and await). It **drops the parent-room scan entirely** — the parent room only ever holds the mirrored consensus, never ticks.

(An earlier revision of this branch scoped to "the newest session in the room." That was a heuristic, not a filter — it leaned on the one-active-session-per-room invariant and could point at a session the handle never joined once a room has multiple agents opening different sessions. Replaced with the participant filter above.)

## Tests

- Backend: `?participant=<handle>` returns only sessions the handle joined (cross-room discriminating case).
- CLI: `tests/test_session_await.py` reproduces the exact `"replayed": true` stale-consensus symptom (verified it **fails pre-fix**), asserts await sends `participant=<handle>`, never scans the old session, and that no joined-session → straight-to-SSE. Full gates green (backend 347, CLI 442, ruff/ty clean).

## Known limitation (pre-existing, not addressed here)

The pre-check has no notion of *which tick the handle has already consumed* — it returns the newest matching tick in the session. So if the agent responds to round N and calls `await` again before round N+1 fires, the pre-check can replay round N's (already-handled) tick. This fix removes the cross-session/other-agent shadowing but not that within-session replay window. A proper follow-up would give `await` a watermark — e.g. pass `since=<handle's last reply time>` (the messages endpoint already supports `since`) or a client-side cursor across `await` invocations — so it only returns ticks newer than the handle's last action. Happy to do that as a separate PR.

## Cleanup

Removed a stale `protocol.py` docstring referencing the long-removed "inline NegMAS mode" — the phantom-mode drift that seeded the bad first-draft comment in this fix.

### DIFF
--- a/fastapi-backend/app/routes/coordination_sessions.py
+++ b/fastapi-backend/app/routes/coordination_sessions.py
@@ -25,7 +25,7 @@ from sqlalchemy.ext.asyncio import AsyncSession
 from app.bus import room_channel
 from app.config import settings
 from app.database import async_session_maker, get_async_session
-from app.models import CoordinationSession, Message
+from app.models import CoordinationSession, Message, Participant
 from app.routes.stream import _close_listen_conn, _open_listen_conn, _stream_with_disconnect_watcher
 from app.schemas import (
     CoordinationSessionRead,
@@ -50,13 +50,20 @@ async def list_coordination_sessions(
         None,
         description="Comma-separated state filter (e.g. 'waiting,negotiating')",
     ),
+    participant: str | None = Query(
+        None,
+        description="Only sessions this agent handle actually joined (via the "
+        "participants table). Used by `session await` to scope to the caller's "
+        "own session rather than guessing the newest one in the room.",
+    ),
     limit: int = Query(200, le=1000),
 ):
     """List coordination sessions, newest first.
 
-    Supports two filters used by the OpenClaw channel plugin's polling loop
-    (``state=waiting,negotiating``) and the frontend sessions rail
-    (``parent_room=<name>``).
+    Supports the OpenClaw channel plugin's polling loop
+    (``state=waiting,negotiating``), the frontend sessions rail
+    (``parent_room=<name>``), and the CLI's missed-tick scope
+    (``participant=<handle>``).
     """
     query = select(CoordinationSession).order_by(CoordinationSession.created_at.desc())
 
@@ -66,6 +73,14 @@ async def list_coordination_sessions(
         wanted = [s.strip() for s in state.split(",") if s.strip()]
         if wanted:
             query = query.where(CoordinationSession.state.in_(wanted))
+    if participant:
+        query = query.where(
+            CoordinationSession.id.in_(
+                select(Participant.coordination_session_id).where(
+                    Participant.agent_handle == participant
+                )
+            )
+        )
 
     result = await db.execute(query.limit(limit))
     return [CoordinationSessionRead.model_validate(s) for s in result.scalars().all()]
--- a/fastapi-backend/tests/test_coordination_sessions.py
+++ b/fastapi-backend/tests/test_coordination_sessions.py
@@ -113,3 +113,29 @@ async def test_sessions_are_a_subset_of_a_room(client):
     # display_name preserves the parent → session relationship for legacy URLs.
     assert alpha_only[0]["display_name"].startswith("alpha:session:")
     assert beta_only[0]["display_name"].startswith("beta:session:")
+
+
+@pytest.mark.asyncio
+async def test_participant_filter_scopes_to_joined_sessions(client):
+    """?participant=<handle> returns only sessions that handle actually joined
+    (via the participants table) — the scope `session await` uses so an older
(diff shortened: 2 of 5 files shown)


juliarvalenti:
Superseded by the SLIM-native rewrite (#418): `coordination_sessions.py` and the `session` command were removed (sessions became episodes), and the missed-tick problem this fixes is now handled by the rewrite's durable inbox + per-handle `await` cursor. Closing as dead code.

PR #409, opened 2026-07-20
fix: SAV retry UX + config knob to disable

### LINKED ISSUE #408: SAV auto-retry: add UI affordance + mycelium config knob to disable it
## Problem

When the CFN's Semantic Alignment Validator (SAV) rejects an agreement at the end of a negotiation round, it silently restarts the round counter from 1. From an agent or operator perspective this looks identical to being stuck — `mycelium negotiate status` shows \"Round 1\" again with no indication that a retry is in progress. There is also no way to disable SAV retries from mycelium config today.

Two separate issues are filed here together because they share the same root cause.

---

## Issue 1 — No UI affordance for SAV retries

When a `coordination_retry` message is posted (detected in `coordination.py` via round regression), nothing surfaces it to:
- `mycelium negotiate status` — the output resets to \"Round 1\" with no retry banner
- `mycelium session await` / the OpenClaw channel plugin — agents receive a fresh tick with no context that this is attempt N
- The frontend (if a session is being watched there)

### What should happen

- `mycelium negotiate status` should show `Attempt 2/N — Round 1` (or equivalent) when a retry is active
- The `coordination_retry` message payload (`attempt`, `prior_rounds`) should be rendered by both the CLI await path and `formatTickInstruction()` in the OpenClaw channel plugin (`route.ts`), following the same two-delivery-path discipline already in CLAUDE.md
- Optionally: surface a `coordination_retry` event in the frontend room timeline

---

## Issue 2 — No mycelium-side config to disable SAV

Today mycelium sends `retry_max_attempts` and `validation_score_intervention` to the CFN at MAS creation time (`rooms.py:_ensure_mas`). The defaults are:

```
CFN_RETRY_MAX_ATTEMPTS=1          # mycelium default (CFN default is 3)
CFN_VALIDATION_SCORE_INTERVENTION=0.6
```

`retry_max_attempts=1` means the CFN will still run SAV once and may silently restart before closing the session. Setting it to `0` would bypass SAV entirely — **but it's currently unknown whether the CFN honors `retry_max_attempts=0` as a disable signal** (the ioc-cfn-cognition-engines repo is not public/indexed; needs a quick check with the CFN team).

We also don't know if there is a flag to lower the SAV scoring baseline independently of the retry count. The `validation_score_intervention` threshold is the only lever we have on the mycelium side today.

### What should happen

1. **Verify with CFN team** whether `retry_max_attempts=0` disables SAV outright or causes an error.
2. Add a `mycelium config` surface for both knobs so operators can set them per-deployment:
   ```
   mycelium config set negotiation.retry_max_attempts 0
   mycelium config set negotiation.validation_score_intervention 0.3
   ```
   These should map directly to `CFN_RETRY_MAX_ATTEMPTS` / `CFN_VALIDATION_SCORE_INTERVENTION` in `config.py`.
3. **Default `CFN_RETRY_MAX_ATTEMPTS` to `0`** (off by default) — SAV retries are opt-in for operators who want multi-attempt negotiation hardening, not the default experience.

---

## Relevant files

- `fastapi-backend/app/config.py:73-74` — the two env-var defaults
- `fastapi-backend/app/routes/rooms.py:115-121` — where they are POSTed to the CFN at MAS creation
- `fastapi-backend/app/services/coordination.py` — `coordination_retry` detection (round regression)
- `mycelium-cli/src/mycelium/integrations/openclaw/assets/mycelium/plugin/src/channel/route.ts` — `formatTickInstruction()` / `formatConsensusSummary()` (needs retry rendering)

juliarvalenti:
Closes #408

## Summary

- **`negotiate status`** shows `[SAV retry — attempt N]` banner when a SAV restart is in progress — no more "why is it Round 1 again?"
- **`session await`** emits `{"type":"retry",...}` on `coordination_retry` instead of silently dropping it, then keeps streaming so agents catch the next tick
- **OpenClaw `route.ts`** — new `routeRetry()` dispatches a `[CognitiveEngine — SAV retry]` notice to all channel agents via `formatRetryNotice()`
- **`CFN_RETRY_MAX_ATTEMPTS` default `1→0`** — SAV retries off by default, opt-in per deployment
- **`NegotiationConfig`** gains `retry_max_attempts` + `validation_score_intervention`; both wired through `docker_utils.py → .env` so `mycelium config set negotiation.retry_max_attempts 1` + `mycelium config apply` enables them

## Open question for CFN team

Whether `retry_max_attempts=0` is honored as a hard disable signal needs confirmation (noted in `config.py` comment). If not, `1` is the effective minimum.

## Test plan

- [ ] `uv run pytest tests/ -x -q` (backend) — 346 passed
- [ ] `uv run pytest tests/ -x -q` (CLI) — passes (pre-existing `test_register_adopted_batch_restarts_gateway_once` failure unrelated)
- [ ] Smoke: trigger a SAV retry in a live demo session and confirm `negotiate status` shows the attempt banner
- [ ] Smoke: confirm `session await` prints the `retry` JSON event and then the next tick arrives correctly

### DIFF
--- a/fastapi-backend/app/config.py
+++ b/fastapi-backend/app/config.py
@@ -64,13 +64,15 @@ class Settings(BaseSettings):
     # run several times the rounds you'd expect:
     #   retry_max_attempts: how many times the cognition engine may reject an
     #     agreement (alignment score below the intervention threshold) and
-    #     restart negotiation from round 1. CFN default is 3; we default to 1
-    #     (take the first agreement; mycelium's own round loop + plan compiler
-    #     handle quality) for predictable timing.
+    #     restart negotiation from round 1. CFN default is 3; we default to 0
+    #     (disabled) — SAV retries are opt-in. Set >0 to enable multi-attempt
+    #     hardening for deployments that need it. Note: whether 0 is a valid
+    #     CFN disable signal needs confirmation from the CFN team; if not, use
+    #     1 as the effective minimum.
     #   validation_score_intervention: the alignment score below which the
     #     engine intervenes/retries. Lower it if LLM-callback agents can't
     #     reach the CFN default (0.6).
-    CFN_RETRY_MAX_ATTEMPTS: int = 1
+    CFN_RETRY_MAX_ATTEMPTS: int = 0
     CFN_VALIDATION_SCORE_INTERVENTION: float = 0.6
 
     # Per-call HTTP timeout (seconds) for CFN start/decide: a DEAD-CONNECTION
--- a/fastapi-backend/app/routes/rooms.py
+++ b/fastapi-backend/app/routes/rooms.py
@@ -381,6 +381,7 @@ async def get_negotiation_status(
         "active": True,
         "session_id": state.session_id,
         "round": state.current_round,
+        "attempt": state.negotiation_attempt,
         "issues": state.issues,
         "issue_options": state.issue_options,
         "current_offer": state.current_offer,
--- a/fastapi-backend/tests/test_cfn_contract.py
+++ b/fastapi-backend/tests/test_cfn_contract.py
@@ -49,13 +49,13 @@ async def test_ensure_mas_sends_mas_config(monkeypatch):
     """_ensure_mas sends mycelium's retry/validation policy under `config` (the
     live MultiAgenticSystemRequest field; an earlier `mas_config` key was
     silently dropped) so a room's MAS doesn't inherit the CFN default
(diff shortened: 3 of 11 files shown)


juliarvalenti:
Superseded by the SLIM-native rewrite (#418): this builds on `commands/negotiate.py`, `commands/session.py`, and `test_cfn_contract.py` — all removed in the rewrite. Closing as dead code.

PR #420, opened 2026-08-04
Step 2: SLIM node + hello-world over a group

juliarvalenti:
Implements **Part V · Step 2** of the SLIM-native rebuild ([START_HERE_STEP_2.md](docs/START_HERE_STEP_2.md)): stand up the SLIM messaging fabric and prove a two-client group broadcast. **Plumbing only** — nothing is wired into rooms/coordination (that's Step 3), so Step 1's green stays green.

## What changed

**SLIM node (Docker)**
- `compose.yml`: new `slim` core service — `ghcr.io/agntcy/slim`, port 46357, with an **inline** node config (Compose `configs.content`, no external file to ship/install). Pinned to **`1.4.0`** (see version trap below).
- `compose-dev.yml`: documented that the node is inherited from `compose.yml` (dev file is a pure override layer always applied on top — nothing dev-specific to build for a pulled image).

**Backend wrapper** — `app/services/slim_client.py` (new)
- Lazy/isolated `slim_bindings` import (missing native wheel → clear `SlimUnavailableError`, no import-time crash across backend modules).
- `workspace/room/agent` ↔ SLIM `Name` (`org/ns/app`) mapping, `to_channel_name`, channel-scoped shared-secret minter.
- `SlimClient`: connect / create_group / invite / listen / publish / receive — reusable by backend + daemon.
- Adds `slim-bindings>=1.4,<2`.

**CLI**
- `config.py`: `SlimConfig.node_endpoint` (scheme-normalizing) + `MYCELIUM_SLIM_ENDPOINT` env.
- `commands/hub.py` (new): `mycelium hub host` (start node, print connect address) and `mycelium connect <address>` (store endpoint) — same command whether self- or mycelium-hosted.

**Tests / harness**
- `tests/test_slim_naming.py` — offline name round-trip / channels / secret minting / segment validation.
- `tests/test_slim_roundtrip.py` — guarded integration test (skips if no reachable node).
- `scripts/slim_roundtrip.py` — throwaway two-client broadcast harness.
- `mycelium-cli/tests/test_slim_config.py` — endpoint normalization + `connect` persistence.

## DoD ✅
`mycelium hub host` runs a node on :46357; the harness exchanges a broadcast between two clients on one group channel:
```
OK — moderator received: b'hello over a slim group'
```
Name mapping covered by the offline unit test.

## Verification gate ✅
- Backend: ruff / format / ty clean; **203 passed, 1 skipped** (pre-existing live-LLM test). Round-trip test runs+passes with a node up.
- CLI: ruff / format / ty clean; **346 passed**.

## Two non-obvious findings
1. **Version coupling** — the node image and `slim-bindings` share a link-negotiation handshake. `:latest` (2.0 line) rejects a 1.4 client with `public key length is invalid`. Node **1.4.0** matches bindings **1.4.1**; bump both together.
2. **Single-process constraints** — one dataplane connection per endpoint per process (→ per-endpoint conn cache), and all group members must share one secret that seeds the group key (→ mint keyed on `workspace/room`, not agent). In prod the moderator and each connector are separate processes.

## Judgment call
"Add the node to both compose files" resolved by defining it fully in `compose.yml` and documenting inheritance in `compose-dev.yml` rather than duplicating the service (the dev file is only ever layered on `compose.yml`, so a duplicate would be dead weight to keep in sync).

## Deferrals (per START_HERE)
SLIM not yet wired into rooms/coordination (Step 3); no durable inbox/persister (Step 4); daemon still uses httpx SSE (Step 5); `abort`→`rejected` L9 subkind is Step 3.

### DIFF
--- a/fastapi-backend/app/services/slim_client.py
+++ b/fastapi-backend/app/services/slim_client.py
@@ -1,0 +1,291 @@
+# SPDX-License-Identifier: Apache-2.0
+# Copyright 2026 Mycelium Contributors
+
+"""Thin async wrapper around ``slim-bindings`` for mycelium's SLIM fabric.
+
+**Step 2 scope only.** This module provides two things:
+
+1. **Naming / identity** — map a mycelium ``workspace/room/agent`` triple to a
+   SLIM :class:`Name` (``org/namespace/app``, with org = workspace/tenant,
+   namespace = room, app = agent id) and back, plus a dev shared-secret minter.
+2. **A small client** (:class:`SlimClient`) that stands up a SLIM app, connects
+   to a node, and creates/joins a **group** channel to exchange broadcasts.
+
+It is deliberately **isolated** from the room/coordination flow — nothing here
+is wired into routes or the bus yet. Room-becomes-a-channel and L9-over-SLIM are
+Step 3; the durable inbox/persister is Step 4. Keeping this standalone is what
+lets Step 1's green stay green.
+
+The ``slim_bindings`` import is **lazy** (native Rust wheel, availability is
+per-platform): modules that never touch SLIM import cleanly even where no wheel
+exists, and a missing wheel degrades to a clear :class:`SlimUnavailableError`
+rather than an import-time crash across every backend module.
+
+Ground truth for the binding API is the cloned examples under
+``~/Documents/GitHub/_slim-research/slim-bindings/python/examples/`` (``group.py``
++ ``common.py``); this wrapper follows their lifecycle.
+"""
+
+from __future__ import annotations
+
+import datetime
+import hashlib
+import hmac
+from dataclasses import dataclass
+from types import ModuleType
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    import slim_bindings
+
+# The node's default listen address (matches ghcr.io/agntcy/slim's default port).
+DEFAULT_NODE_ENDPOINT = "http://127.0.0.1:46357"
+
+# Minimum shared-secret length required by SLIM's dev auth (also seeds MLS).
+MIN_SECRET_LEN = 32
+
+# Default third segment for a room's group-channel Name when no explicit topic
+# is given. A channel is a Name whose app segment is the topic (bible §7b); the
+# members' own apps are their agent ids.
+DEFAULT_CHANNEL_TOPIC = "room"
+
+# Dev-only master secret from which per-channel shared secrets are
+# *deterministically* derived, so every agent on a room's channel and the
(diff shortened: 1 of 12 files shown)


juliarvalenti:
## Review — Step 2 gate ✅ (excellent, on-plan)

Stacks cleanly on merged Step 1; **purely additive** (12 files, 0 deletions), so Step 1's green is untouched. Reviewed against bible **§7** + Step 2 DoD.

**Verified:**
- **`services/slim_client.py`** — the reusable foundation, and it's careful, correct work:
  - Name mapping is exactly §7b: `workspace/room/agent` → `org/ns/app`; channel = `workspace/room/topic` (third segment is topic, distinct from the member's agent id). ✓
  - **Shared secret keyed per-channel** (`workspace/room`, not agent) via HMAC-SHA256 so every member reconstructs it offline without a key exchange — matches the SLIM constraint that all group members share the one secret that seeds the group key. ✓
  - **One dataplane connection per endpoint per process** (`_connections` cache) with the correct rationale + note that prod separates moderator/connector into their own processes. ✓
  - **Lazy `slim_bindings` import** → `SlimUnavailableError`, no import-time crash across backend modules. ✓
  - Lifecycle (`connect / create_group / invite / listen_for_session / publish / receive`) faithfully mirrors the binding examples; `receive` is flagged as the future wake monitor. ✓
- **Version-match pair** — compose pins `ghcr.io/agntcy/slim:1.4.0`, backend pins `slim-bindings>=1.4,<2` (→1.4.1). The `:latest`/2.0 handshake trap is documented. ✓
- **Green** — backend 203 passed / CLI 346 passed / ruff / ty; round-trip harness prints the received broadcast. ✓
- **Judgment call** (node defined in `compose.yml`, inheritance documented in `compose-dev.yml` rather than duplicated) — agreed, right call.
- **Deferrals correct** — wiring into rooms (Step 3), persister (Step 4), daemon SSE→SLIM (Step 5), `abort`→`rejected` (Step 3).

**Two forward-looking notes (non-blocking, for later steps):**
1. **MLS is off** for the hello-world (reasonable). But turning MLS **on** for the room channel matters beyond encryption — the hosted-rendezvous de-risk story (bible §16) relies on intermediate nodes seeing only ciphertext. Please enable it when the room channel goes live (Step 3/4), don't let it stay off by default.
2. The `_connections` cache has **no teardown/close** yet — fine for a one-shot, but Steps 4–5 (long-lived connections) need connection lifecycle (close on shutdown) to avoid leaks/stale conns.

**Verdict:** meets Step 2 DoD, green, foundation is solid. Merging.

PR #421, opened 2026-08-05
Step 3: L9 over SLIM (the bus) + room = channel

juliarvalenti:
Picks up **Step 3** of the SLIM-native rebuild (after #419 Step 1, #420 Step 2), against `slim-native-rewrite`. This is the step where SLIM stops being a standalone hello-world and becomes the coordination bus, wired into rooms.

## What changed

**L9↔SLIM binding** — new `app/services/l9_slim.py`, the seam between `l9.py` (envelope construction) and `slim_client.py` (transport):
- `serialize_envelope` / `deserialize_envelope` — an L9 envelope rides under the additive `l9` key of a message's content JSON (agents never speak L9).
- `CausalOrderBuffer` — holds a message back until every id in its `message.parents` has been delivered, then releases it (and any dependents) in causal order.
- `EpisodeLifecycle` + `build_episode_abort_envelope` — a mid-episode membership change aborts the episode (`commit:rejected`), channel untouched.
- `L9SlimChannel` — composes serialize + causal ordering over one SLIM group session.

**Room = SLIM channel** — new `app/services/room_channels.py`: backend-as-moderator registry. Creating/opening a room provisions one long-lived group channel (backend is moderator), invites members, tracks membership, enforces the episode↔channel lifecycle. Everything is **best-effort** — no reachable node ⇒ no channel, never a failed room create (a `node_reachable` pre-flight keeps the no-node path fast).

**Routes / lifecycle:**
- `rooms.py` provisions on create, closes on delete.
- `sessions.py` invites on join (**fire-and-forget** — the SLIM handshake, which retries against agents that have no connector until Step 5, must never block the HTTP join) and removes on leave; presence reflects SLIM channel membership when a channel is live, falling back to `local_state` otherwise.
- `main.py` tears down channels + the shared connection on shutdown.

**`slim_client.py`** (Step 2 review notes): **MLS ON** for real room channels (hosted-rendezvous story rests on intermediate nodes seeing only ciphertext); connection **close/teardown** path (`close_connection` + `SlimClient.close`, tracking created/joined sessions); `node_reachable` helper.

**Subkind flip** — a failed negotiation now commits as `rejected` (was `abort` under the removed Go CFN) in `l9.py:VALID_SUBKINDS`; the now-stale CFN-authoritative notes are corrected in `l9.py`, `l9_episode.py`, and `CLAUDE.md`.

## Definition of Done ✅

Proven by hand against a live `slim:1.4.0` node (`scripts/l9_slim_roundtrip.py`):
[...]

PR #422, opened 2026-08-05
Step 4: Backend as room infrastructure (persister + durable inbox + wake)

juliarvalenti:
## Step 4 — Backend as room infrastructure

Picks up [`docs/START_HERE_STEP_4.md`](https://github.com/mycelium-io/mycelium/blob/step-4-persister-durable-inbox/docs/START_HERE_STEP_4.md) (Step 3 merged in #421). This is the step where the backend stops being a passive moderator and becomes the always-on room infrastructure: it **joins the channel's read loop** as the persister / durable inbox / trigger-watcher.

### What changed

**New — `app/services/persister.py`**
- **`RoomPersister`** — a long-lived task per room, started by `room_channels` on `provision`, that pulls `L9SlimChannel.receive_with_context()` and, for each message: records the transcript, feeds the UI bus, and runs the trigger-watcher.
- **Persister / transcript** — the full stream is written to the room's markdown at `log/transcript.md` (a jsonl-bearing memory file): git-shareable, reindexed the normal way, and a **distinct artifact** from the episode-scoped `log/episodes/*` (no clobber).
- **Durable inbox** — SLIM keeps nothing for an offline member (bible §7d). `DeliveryLog` tracks a per-`(room, handle)` delivery cursor; on **reconnect** (a membership add for a handle the persister has already seen) the backend **re-serves the missed tail point-to-point** via `publish_to`, so a broadcast an agent missed while offline is delivered in order on rejoin. The receiver's own `CausalOrderBuffer` re-orders/de-dupes.
- **Trigger-watcher (skeletons)** — recognizes `@`-summon tokens (`find_summons`) → summon hook (real engine is **Step 7**) and `commit:converged` (`is_converged`) → plan-compile seam (fired in **Step 8**). Both hooks log only here.

**Extended**
- `slim_client.py` — `publish_to` (targeted point-to-point send) + `receive_message` (whole message with `.context`).
- `l9_slim.py` — `receive_with_context()` / `send_content_to(ctx, content)` + a `serialize_content` helper (content dict form kept so a missed message can be re-served verbatim).
- `room_channels.py` — owns the persister start/stop lifecycle (cancel on `close`/`close_all`, strong task ref) and surfaces the reconnect → re-serve on `invite`. Reconnect re-serve is kept **orthogonal** to episode-abort (a reconnect re-serves regardless of episode state; a re-serve never resurrects an aborted episode).
- Updated the now-stale `# TODO(step4)` / `# TODO(step3)` notes in `local_state.py` / `routes/messages.py` (the UI HTTP/SSE feed stays until Step 10).

### Definition of Done ✅

An agent offline during a broadcast **receives the missed message(s) on reconnect, in order**, and the **channel transcript is persisted to markdown**. The trigger-watcher recognizes an `@`-summon and calls the (skeleton) summon hook; a `commit:converged` fires the plan-compile seam (without compiling).

### Tests

**Fast unit gate (no node)** — `tests/test_persister.py`: delivery cursor (advances for present members, yields exactly the missed tail for an offline agent, empty when caught up); transcript persistence + reload + coexists with `log/episodes/*`; `@`-summon recognition (and the `l9` key is never scanned); `commit:converged`-only plan hook; ingest wiring fires the hooks.

**Live-node integration slice (guarded)** — `test_l9_over_slim_roundtrip.py::test_durable_inbox_reserves_missed_message_on_reconnect` + a `run_durable_inbox(...)` scenario in `scripts/l9_slim_roundtrip.py`.

Verified end-to-end against a standalone `ghcr.io/agntcy/slim:1.4.0` node:

```
228 passed, 5 skipped         # backend fast gate
4 passed                       # backend live-node slices (all prior slices still pass)
346 passed                     # CLI gate (ruff + format + ty + pytest)

OK — moderator received in causal order: ['root-envelope-id', 'child-envelope-id']
OK — mid-episode membership change aborted with commit:rejected
OK — durable inbox re-served missed message on reconnect: ['missed-while-offline']
```

### Deferrals (named per the plan)

Claude Code connector + wake bridge and the daemon SLIM retarget are **Step 5** (`daemon/dispatch.py` stays on httpx SSE here); human `@`-mention + consent is **Step 6**; cognition engines (wiring `on_summon`) are **Step 7**; plan-compile firing (wiring `on_converged`) + memory sync are **Step 8**; SSE/`stream.py` retires in **Step 10**.

Handoff for the next agent: [`docs/START_HERE_STEP_5.md`](https://github.com/mycelium-io/mycelium/blob/step-4-persister-durable-inbox/docs/START_HERE_STEP_5.md).

### DIFF
--- a/docs/START_HERE_STEP_5.md
+++ b/docs/START_HERE_STEP_5.md
@@ -1,0 +1,213 @@
+# START HERE — Step 5 (Claude Code connector + wake bridge — the dogfood milestone)
+
+Companion to [`START_HERE.md`](./START_HERE.md). Step 4 is **done** (this PR into
+`slim-native-rewrite`); you are picking up **Step 5**. Read `START_HERE.md`,
+[`START_HERE_STEP_1.md`](./START_HERE_STEP_1.md) → [`START_HERE_STEP_4.md`](./START_HERE_STEP_4.md)
+first if you haven't — the same rules apply. This file gives you (a) the exact state Step 4
+left behind, (b) your Step 5 marching orders, (c) the facts you must internalize, and (d) the
+traps specific to this step.
+
+**This is the step where a real agent finally rides the fabric.** Step 4 made the backend the
+always-on room infrastructure (persister + durable inbox + trigger-watcher). But no agent yet
+holds its own SLIM connection: the daemon still streams over httpx SSE. Step 5 retargets the
+Claude Code connector onto SLIM so a registered agent is **woken** by an inbound message,
+**spawned** for a turn, and its **reply lands back in the channel** — the first end-to-end
+dogfood loop.
+
+## Ground rules (unchanged from START_HERE.md)
+
+1. **The bible is authoritative:** [`slim-native-rebuild-bible.md`](./slim-native-rebuild-bible.md).
+   Your work is **Part V · Step 5**, with **§12 (wake / `@`-mention)**, **§10 (engines — not
+   yet, but the wake seam is shared)**, and **§13 (the full cycle)** as the design reference.
+2. Leave the project **runnable and green** — backend + CLI quality gates pass at the end.
+3. Code/config blocks in the bible are **reference, not paste**. The cloned `slim-bindings`
+   examples under `~/Documents/GitHub/_slim-research/` remain ground truth for the binding API
+   (the **participant** side — `listen_for_session` / `get_message` loop — is what you'll build
+   on now; the moderator side is the backend's).
+4. **Verify before you edit** — the paths below were accurate at the end of Step 4; confirm
+   shape before changing a file.
+5. Fixed decisions are not up for debate; the one open question (native-vs-CLI) has a default:
+   **keep cold-spawn** (the daemon holds the SLIM connection; the agent's contract is unchanged).
+
+## Where Step 4 left things (your starting state)
+
+Branch off `slim-native-rewrite` (Step 4 is merged). The backend is now a full room participant
+that reads the channel; the **agent side is still off-fabric**. Concretely:
+
+- **The backend consumes the channel.** `app/services/persister.py` holds `RoomPersister`: a
+  long-lived task started per room by `room_channels.py` on `provision`. It pulls
+  `L9SlimChannel.receive_with_context()`, records the transcript to `log/transcript.md`, feeds
+  the UI bus, and runs the trigger-watcher. **This is the moderator's read loop** — your
+  connector is the *member's* read loop, a different process/app.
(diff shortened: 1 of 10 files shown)


juliarvalenti:
## Review — Step 4 gate ✅ (durable inbox correct; three non-blocking notes)

The load-bearing step (SLIM drops offline members' messages, so the persister *is* the durable inbox). Reviewed against bible §7d/§9 + Step 4 DoD. Green (228 backend / 5 skipped, 4 live-node slices, 346 CLI; ruff/ty).

**Verified — the durable-inbox logic is correct (for a process lifetime):**
- `DeliveryLog` cursor model is right: a fresh join starts caught-up (missed nothing prior); `record(delivered_to=present)` advances present members and freezes absent ones; `undelivered()` yields exactly the missed tail; re-serve marks caught-up.
- `note_join` correctly distinguishes fresh-join vs reconnect; `reserve` re-serves the tail **point-to-point** (`send_content_to`) in order, no-op without a cached context (expected until Step 5).
- Receive loop handles shutdown (`CancelledError` re-raised) + gives up after 3 consecutive failures (no hot-loop on a torn-down channel). Transcript is a distinct `log/transcript.md` artifact, reindexed normally; triggers (`@`-summon, `commit:converged`) are skeleton hooks logging only, as planned.
- **Integration slice present** — `test_durable_inbox_reserves_missed_message_on_reconnect` seeds the suite's 3rd slice; all prior slices still pass.

**Non-blocking notes:**
1. **Cursors are in-memory only → the re-serve does NOT actually survive a backend restart** (the `RoomPersister` docstring claims "cursors/re-serve survive a backend restart" — that's not true: `_cursors` starts empty on restart, so a reconnecting *known* handle is misclassified as a fresh join by `note_join` and its pre-restart missed tail is silently dropped). The **transcript** survives; the **per-handle delivery position** does not. DoD (single-process offline→reconnect) is met, so non-blocking — but please either **persist the cursors** alongside the transcript or, at minimum, **fix the docstring** so it doesn't overclaim.
2. **Full transcript rewrite per message** — `_ingest` calls `write_transcript(room, self.log.records)` (whole file) on every message → O(n)/msg, O(n²) over a room, and the transcript grows unbounded. Append instead of rewrite; consider rotation/segmentation for long-lived rooms.
3. **Carry-forward from Step 3** — `CausalOrderBuffer._delivered`/`_pending` are still unbounded; prune at episode close (this was the "address in Step 4" note; fine to push to a hardening pass, just don't lose it).

**Verdict:** meets Step 4 DoD, correct for the primary case, integration slice shipped. Merging. (Note #1 is worth a quick follow-up — persist cursors or fix the comment — but doesn't gate.)

PR #423, opened 2026-08-05
Step 5: Claude Code connector + wake bridge (daemon on SLIM)

juliarvalenti:
## What this is

**Step 5 of the SLIM-native rebuild** (bible Part V · Step 5) — the dogfood milestone. Retargets the daemon from an httpx **SSE** stream onto a **SLIM group subscription** so a registered Claude Code agent finally rides the fabric: it is **woken** by an inbound message, **spawned** for a headless turn, and its **reply lands back in the channel**.

This is the **first CLI-side step** of the rewrite (Steps 0–4 were backend-only), so it also stands up the daemon's own SLIM+L9 layer.

## How it works

- **Connector (`daemon/connector.py`).** Per owned `(room, handle)` it connects a member `SlimClient`, `listen_for_session` (the backend moderator invites it), sends a presence **hello** (seeding the durable-inbox re-serve route — Step 5 trap option (a)), then pumps a `get_message` wake loop. Wake fires on an inbound L9 **`exchange`** addressed to the handle (via L9 `participants` recipients **or** a raw `@handle` token; `sender != handle` loop guard). The reply is an `exchange` parented on the message that woke it (`parents = [woke_id]`), published to the channel — the backend persister records it.
- **`runner.py`** launches one connector per owned cold-spawn handle (`connector_targets`), reconciles on SIGHUP, cancels on shutdown (then drops the shared SLIM connection). The legacy SSE stream + coordination poller stay in `dispatch.py` (unit tests still pass) but are **no longer wired** — they retire with the backend SSE in Step 10.
- **Daemon SLIM+L9 layer (`mycelium/slim/`).** The CLI installs as a thin `uv tool` and can't drag in the FastAPI/ML backend, so the daemon gets its own small mirror: `naming.py` (Name mapping + `mint_shared_secret`, **byte-for-byte** the backend's so MLS matches), `client.py` (`SlimClient`), and a lean stdlib-only `l9.py`. The reply shape is proven **byte-for-byte identical** to the backend's `l9.envelope_to_dict` and round-trips through `l9.parse_envelope` / `deserialize_envelope` unchanged. Adds `slim-bindings>=1.4,<2`.
- **Gates preserved.** `allow_from`/budget/depth, the per-handle serial lock, and `abort`/`status` control verbs are reused wholesale from `dispatch.py` — only the transport + reply sink change. The agent's contract is unchanged.

## DoD

✅ A registered Claude Code agent joins a room, is woken by an inbound message, spawns a turn, and its reply appears in the room. Budget/depth/ownership gates and `abort`/`status` still hold.

## Tests

- **Fast gate (merge gate, no node, `claude` mocked):** wake path, reply-is-valid-L9, budget/depth/allow_from gates, per-handle lock serialization, abort/status control verbs, L9 helper round-trip. **CLI: 368 passed.**
- **Guarded live-node slice:** `test_connector_wake_over_slim.py` — connector wakes on an inbound L9 over a real MLS group and its reply is received by the moderator. ✅ passes on a live `slim:1.4.0` node.
- **Backend untouched** — all prior slices (Steps 1–4, incl. `test_l9_over_slim_roundtrip.py`) still pass live. **Backend: 232 passed, 1 skipped.**
- Backend + CLI `ruff` / `ruff format` / `ty` all clean.

## Notable decision (flagged)

The integration slice the Step 5 doc suggested adding to the backend's `scripts/l9_slim_roundtrip.py` lives **CLI-side** instead (`tests/test_connector_wake_over_slim.py`) — the connector is in the CLI and the backend script can't import it. It drives the connector's real transport primitives against a live node with a fake backend moderator.

## Handoff

Adds `docs/START_HERE_STEP_6.md` (human-in-the-room + `@`-mention + consent).

Deferrals: cognition engines (`on_summon`) → Step 7; plan-compile (`on_converged`) + memory sync → Step 8; cross-machine → Step 9; SSE/`stream.py` (and the now-unwired daemon SSE helpers) → Step 10.

### DIFF
--- a/docs/START_HERE_STEP_6.md
+++ b/docs/START_HERE_STEP_6.md
@@ -1,0 +1,179 @@
+# START HERE — Step 6 (Human-in-the-room + `@`-mention + consent)
+
+Companion to [`START_HERE.md`](./START_HERE.md). Step 5 is **done** (this PR into
+`slim-native-rewrite`); you are picking up **Step 6**. Read `START_HERE.md`,
+[`START_HERE_STEP_1.md`](./START_HERE_STEP_1.md) → [`START_HERE_STEP_5.md`](./START_HERE_STEP_5.md)
+first if you haven't — the same rules apply. This file gives you (a) the exact state Step 5
+left behind, (b) your Step 6 marching orders, (c) the facts you must internalize, and (d) the
+traps specific to this step.
+
+**Step 5 put a real agent on the fabric.** A registered Claude Code agent now holds its own SLIM
+connection (via the daemon connector), is **woken** by an inbound L9 message addressed to it,
+**spawns** a headless turn, and its **reply lands back in the channel** — the first end-to-end
+dogfood loop. But the only thing that can address it today is another agent's connector or a raw
+test publisher. **Step 6 puts the human in the room:** the backend publishes the human's message
+onto the channel, `@`-parses it into L9 recipients, wakes in-room agents (via Step 5's bridge),
+and — the hero-demo differentiator — surfaces a **consent-to-be-woken** prompt when an agent is
+`@`-invited into a room it isn't in yet.
+
+## Ground rules (unchanged from START_HERE.md)
+
+1. **The bible is authoritative:** [`slim-native-rebuild-bible.md`](./slim-native-rebuild-bible.md).
+   Your work is **Part V · Step 6**, with **§12 (wake / `@`-mention / invite / consent)** and
+   **§13 (the full cycle)** as the design reference.
+2. Leave the project **runnable and green** — backend + CLI quality gates pass at the end.
+3. Code/config blocks in the bible are **reference, not paste**.
+4. **Verify before you edit** — the paths below were accurate at the end of Step 5; confirm shape
+   before changing a file.
+5. Fixed decisions are not up for debate; the one open question (human identity) has a default:
+   **the human is spoken-for by the backend** — no human connector for the MVP.
+
+## Where Step 5 left things (your starting state)
+
+Branch off `slim-native-rewrite` (Step 5 is merged). The agent side is now on the fabric; the
+**human side is still HTTP-only** and no `@`-parse maps mentions to L9 recipients. Concretely:
+
+- **The daemon is a SLIM member, not SSE.** `mycelium-cli/src/mycelium/daemon/connector.py`
+  holds `run_connector(room, handle, …)`: per owned `(room, handle)` it connects a member
+  `SlimClient`, `listen_for_session` (the backend moderator invites it), sends a presence
+  **hello** (seeding the durable-inbox reply route — Step 5 trap option (a)), then pumps a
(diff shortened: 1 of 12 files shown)


juliarvalenti:
## Review — Step 5 gate ✅ (the dogfood milestone; this is the one that makes it real)

Retargets the daemon from httpx SSE → a SLIM subscription so a Claude Code agent wakes and replies over the fabric. Reviewed hard against bible §12 + Step 5 DoD. Green (CLI 368, backend 232 + live slices; ruff/format/ty clean).

**Verified — the connector is careful, correct work:**
- `should_wake` has the critical **loop guard** (`sender != handle`, so it never wakes on its own broadcast reply), wakes only on `exchange`, addressed via L9 `participants` **or** a guarded `@handle` token.
- **All gates applied** — `allow_from`/budget/**depth** (the depth cap is the backstop that bounds multi-agent ping-pong) + the per-handle serial lock; control verbs (`abort`/`status`) run outside the lock and skip budget/depth, matching `dispatch.py`. Reused wholesale — good.
- `build_reply` threads causality (`parents=[woke_id]`) and stays in the woke episode/topic, so the backend's causal buffer + transcript stay correct.
- The **presence hello seeds the durable-inbox re-serve route** — this closes the exact Step 4 gap I flagged (re-serve had no route to a never-spoke member). 👍
- Resilient reconnect loop, `CancelledError` re-raised, `client.close()` in `finally`, inbound dispatched off-loop so control verbs stay responsive during a turn.
- **Integration slice present** — `test_connector_wake_over_slim.py` (connector wakes on inbound L9 over a real MLS group; moderator receives the reply). This *is* the cross-package drift guard: it exercises both copies end-to-end, and a secret-mint drift would fail the MLS join outright.

**Non-blocking notes:**
1. **The drift guard is only in the guarded (skip-if-no-node) suite.** The CLI's `mycelium/slim/` mirrors the backend's naming/secret/envelope/URN forms out of necessity (the thin `uv tool` CLI can't import the FastAPI backend). It's faithful today, but a future divergence (secret literal, `workspace/room` scope, envelope shape, or the hardcoded `_room_episode`/`_room_topic` URN forms) could **merge green** and only surface when the live suite runs. Add a **fast-gate golden test** — a hardcoded expected envelope dict + secret digest that *both* packages assert against — or, better, extract the shared primitives into a tiny package both depend on. Cheap insurance against a very hard-to-debug "agents silently can't join" failure.
2. **Fire-and-forget inbound task isn't ref-held** — `asyncio.create_task(_guarded_inbound(...))` (RUF006 suppressed) has no strong reference, so it can be GC'd mid-spawn (orphaning the reply + budget accounting). `room_channels.py` handles exactly this with a `self._tasks` set + done-callback; mirror that here for consistency and safety.

**Verdict:** meets Step 5 DoD — a Claude Code agent genuinely rides the fabric now, gates + loop-guard intact, integration slice shipped, and it closes the Step 4 re-serve-route gap. Merging.

PR #425, opened 2026-08-05
Debt D2: extract shared SLIM+L9 package (+ golden drift test)

juliarvalenti:
Pays down **D2** (bible Part IV, Known debt register): the CLI daemon carries a copy of the backend's SLIM+L9 primitives so the thin `uv tool` CLI need not import the FastAPI/ML backend. Drift between the copies is silent and vicious — a diverging `mint_shared_secret` / master-secret literal / `workspace/room` HMAC scope / L9 envelope shape / URN form means MLS group keys mismatch and connectors **silently can't join** (no app-level error), or messages get dropped/misrouted. Until now, only the guarded live-node slice caught this.

## Outcome: the (A) golden-test safety net (full extraction (B) deferred)

Per the register's fallback, this delivers **(A)** — the fast-gate golden drift test — and defers **(B)**, the shared installable package, with a documented rationale. **(A) is the register's minimum acceptable outcome** ("closing the merges-green gap"); it fully removes the *correctness* risk now.

### What shipped

- **`slim-l9-golden.json`** (repo root) — the single frozen source of truth for the shared wire constants: the `mint_shared_secret` HMAC digest for a known `workspace/room`, the master-secret / workspace / channel-topic / node literals, the `episode`/`topic` URN forms, and a **golden serialized `exchange` envelope dict**.
- **`fastapi-backend/tests/test_slim_l9_golden.py`** — the backend primitives (`app.services.slim_client` + `app.services.l9`) assert against the golden file.
- **`mycelium-cli/tests/test_slim_l9_golden.py`** — the CLI copy (`mycelium.slim.{naming,client,l9}` + `daemon.connector` URN helpers) asserts against the **same** file.

One frozen source, **two asserters**: neither copy can move without turning a **unit** gate red (previously drift was caught only by the guarded live-node slice). I verified up front that the two serializers already produce **byte-for-byte identical** output — the backend's `l9.envelope_to_dict(l9.build_envelope(kind=exchange, ...))` and the CLI's `l9.build_reply_content(...)` — and the golden freezes that exact dict.

- **`docs/slim-native-rebuild-bible.md`** — the D2 note now records **(A) delivered / (B) deferred** and why.

### What moved vs. stayed

Nothing *moved* — no code was relocated (that's (B)). The genuinely-shared code is already stdlib-only and byte-for-byte identical on both sides; the backend's richer `l9.py` (pydantic `L9` models, `VALID_SUBKINDS`, validation) stays backend-side by design. (A) locks the shared *values/shape* without touching either implementation.

### Byte-for-byte compat

Preserved and now **enforced at the unit gate**: the golden dict is the exact serialized `exchange` content both sides emit, and the digest/URN/constant assertions pin `mint_shared_secret`, the Name-mapping literals, and the URN forms.

### Why B (shared installable package) was deferred

Both delivery paths make a new shared *distribution* a cross-cutting release/packaging change, not a local refactor:
- The **backend `Dockerfile`** installs deps by parsing `pyproject.toml`'s `project.dependencies` and `pip install`-ing them **from PyPI**, then `COPY . .` with a build context scoped to `fastapi-backend/` only — a sibling path/workspace dependency is unreachable there.
- The **end-user CLI install** (`install.sh`) is a **single self-contained wheel** from GitHub releases (or `mycelium-cli` on PyPI); a path dep on a sibling package would neither be bundled into that wheel nor exist on PyPI, so the installed CLI would fail to import.

Making B green end-to-end therefore means publishing/bundling a new package into the release-wheel + Docker + `install.sh` flows — numbered-step infra explicitly out of scope for this debt paydown, and a live risk to green installs. B remains the ideal and should land alongside the Step 7–8 packaging pass that revisits those flows anyway.

## Gate results (fast unit tier — all green)

| | ruff check | ruff format --check | ty check | pytest |
|---|---|---|---|---|
| **fastapi-backend** | ✅ | ✅ | ✅ | ✅ 241 passed, 5 skipped |
| **mycelium-cli** | ✅ | ✅ | ✅ | ✅ 374 passed, 2 skipped |

The guarded **live-node integration slices** were not run here (they need a running `slim:1.4.0` node); as with prior steps they should be run separately. This PR's value is precisely that it moves drift detection *below* that guarded suite into the fast unit gate.

### DIFF
--- a/docs/slim-native-rebuild-bible.md
+++ b/docs/slim-native-rebuild-bible.md
@@ -525,6 +525,32 @@ but only in the guarded suite. Fix: (A) a fast-gate golden test both packages as
 then (B) extract a shared `mycelium-slim-l9` package **before Steps 7–8** grow the shared
 surface.
 
+**Status (A delivered; B deferred).** The fast-gate golden test is **shipped**:
+`slim-l9-golden.json` at the repo root freezes the shared wire constants (the
+`mint_shared_secret` digest for a known `workspace/room`, the master-secret /
+workspace / channel-topic / node literals, the `episode`/`topic` URN forms, and a
+golden serialized `exchange` envelope dict). Both suites read that **one** file and
+assert their own copy against it — `fastapi-backend/tests/test_slim_l9_golden.py`
+and `mycelium-cli/tests/test_slim_l9_golden.py`. One frozen source, two asserters:
+neither copy can drift without turning a **unit** gate red (previously drift was
+caught only by the guarded live-node slice). This closes the "merges green" gap.
+
+**Why B (the shared installable package) was deferred:** both delivery paths make a
+new shared *distribution* a cross-cutting release/packaging change, not a local
+refactor. The backend `Dockerfile` installs deps by parsing `pyproject.toml`'s
+`project.dependencies` and `pip install`-ing them from **PyPI**, then `COPY . .`
+with a build context scoped to `fastapi-backend/` only — a sibling path/workspace
+dependency is unreachable there. The end-user CLI install (`install.sh`) is a
+**single self-contained wheel** from GitHub releases (or `mycelium-cli` on PyPI);
+a path dep on a sibling package would neither be bundled into that wheel nor exist
+on PyPI, so the installed CLI would fail to import. Making B green end-to-end
+therefore means publishing/bundling a new package into the release-wheel + Docker
++ `install.sh` flows — numbered-step infra explicitly out of scope for this debt
+paydown, and a live risk to green installs. The genuinely-shared code is already
+stdlib-only and byte-for-byte identical on both sides, so (A) removes the
+*correctness* risk now; (B) remains the ideal and should land alongside the
+Step 7–8 packaging pass that revisits the release/Docker flows anyway.
+
 **D11 — the other harnesses will be stale.** The MVP migrates **claude_code** only.
 - **cursor** (cold_spawn) shares the daemon's SLIM connector path (`connector_targets`
   includes it) and the family-agnostic spawn, so it *probably works over SLIM already* — but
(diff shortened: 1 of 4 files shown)


juliarvalenti:
## Review — Debt D2 ✅ (golden drift guard shipped; full extraction correctly deferred)

The (A) fallback from the register — and it's the right call, well-executed. Purely additive (+310/−0, 4 files), green (backend 241/5, CLI 374/2, ruff/format/ty clean).

**Verified:**
- **`slim-l9-golden.json`** freezes the shared wire constants: the `mint_shared_secret` digest for a known `workspace/room`, the master-secret / workspace / channel-topic / node literals, the `episode`/`topic` URN forms, and a golden serialized `exchange` envelope.
- **Both suites assert their own copy against the same file** — the backend test computes backend `mint_shared_secret`/URNs/`envelope_to_dict` output; the CLI test computes `mycelium.slim` + `connector._room_episode/_room_topic`/`build_reply_content` output; both compare to the single golden. **One frozen source, two asserters** → a drift in either copy turns a **fast unit gate** red (the core guards need no node). This closes the "drift merges green" gap that made D2 worrying.
- **The packaging finding is valuable and correct** — full (B) extraction is blocked not by code entanglement but by the two independent install pipelines (backend Dockerfile pip-installs from PyPI with context scoped to `fastapi-backend/`; the CLI ships as a single self-contained release wheel). A local sibling path dep is unreachable in both without release/Docker/`install.sh` changes. Deferring B to the Step 7–8 packaging pass, with A removing the correctness risk now, is the right sequencing.

**Note for Step 7/8:** the golden guard covers today's shared surface. When those steps add new shared L9 shapes (e.g. `knowledge` envelopes), **extend `slim-l9-golden.json`** so the new surface is guarded too.

**Verdict:** de-risks D2 (drift now caught at the unit gate), full extraction correctly deferred with clear rationale. Merging — Step 7 is unblocked.

PR #427, opened 2026-08-05
Step 8: plan + memory sync (the full loop, same-machine)

juliarvalenti:
Picks up **Step 8** from [`START_HERE_STEP_8.md`](docs/START_HERE_STEP_8.md), on `slim-native-rewrite` (Step 7 merged). Closes the `join → exchange → converge → plan → work` loop **on one machine**: when the aligner emits `commit:converged`, the backend now compiles the plan and propagates it as memory over SLIM.

## What changed

**Backend**
- **`memory_sync.py`** — the L9 `knowledge` write path (bible §11). `build_knowledge_envelope` wraps a `KnowledgeWrite` (key + markdown + `version`/`base_version`) as a `knowledge:distillation` envelope that **carries the content** (push-with-content, not notify-then-pull). `apply_knowledge` writes it into a local store + reindexes the JSONL under the decided **last-write-wins** policy: a same-version arrival is an **idempotent** no-op (the same-machine loopback of a write the backend just made), a **stale-base** write **fails with details** (current content + `updated_by` + `updated_at`) and moves on, **no merge**.
- **`plan_sync.py`** — the `on_converged` consumer. Compiles `plan/tasks.md` via `plan_compiler` (**fail-soft** to the raw agreement on an LLM outage) and broadcasts the plan as a `knowledge` push, which doubles as the **"plan ready"** signal now that the verdict is already on the wire. A distinct consumer stage across the seam, **not** a CE step (CLAUDE.md plan-compiler decision).
- **`room_channels.py`** — `RoomConvergedHook` + `_converged_adapter`, room-aware exactly like Step 7's summon seam; wired in `main.py` to `PlanSyncEngine.handle_converged`.
- **`plan.py`** — `write_plan_file` gains a `version` param so re-compiles bump the version the `knowledge` stream orders writes by.

**CLI**
- **`daemon/connector.py`** — `apply_knowledge_message` acts on a `knowledge` arrival (before `should_wake`): writes the carried markdown into the local store; never wakes a turn. **`filesystem.apply_knowledge`** mirrors the conflict policy CLI-side.
- **`slim/l9.py`** — `KNOWLEDGE_KIND` + `payload_data_of`.

## Definition of Done ✅
`@`-summon the aligner over a seeded exchange → it emits `commit:converged` → the backend compiles `plan/tasks.md` → a `knowledge` message writes the markdown + updates the JSONL on a second local store. The same-machine mini-demo of the hero flow.

## Tests
[...]

PR #428, opened 2026-08-05
Debt D2: extend golden guard for the knowledge envelope

juliarvalenti:
## What & why

Debt **D2** (bible Part IV, Known debt register): the repo-root `slim-l9-golden.json` freezes shared SLIM+L9 wire constants that both the FastAPI backend and the CLI daemon must reproduce byte-for-byte, and two golden tests assert against it. It covered the `exchange` envelope + `mint_shared_secret` + URN forms — but **not** the new shared shape Step 8 (PR #427) added: the L9 **`knowledge:distillation`** envelope + **`KnowledgeWrite`** payload used to sync memory over SLIM.

The live integration slice guards that end-to-end, but the **fast unit gate did not** — a drift between what the backend produces and what the CLI parses could merge green. Step 9 (cross-machine) relies entirely on this knowledge stream, so a silent drift would break cross-machine memory sync. This closes that gap in the fast gate.

## What I froze

New `knowledge` section in `slim-l9-golden.json` (existing sections untouched):
- a known `KnowledgeWrite` input — `key`, `content`, `version`, `base_version`, `created_by`, `updated_by`, `updated_at`
- the golden serialized `knowledge:distillation` **envelope dict** the backend's `build_knowledge_envelope` produces for it (mirrors the existing `envelope` section)

`build_knowledge_envelope` mints a fresh random `message.id` per call (not part of the shared contract), so the golden's `message.id` is a `MESSAGE_ID_PLACEHOLDER` sentinel and the backend test normalizes the produced id to it before comparing. Every other field is frozen byte-for-byte.

## The two new assertions

The produce↔parse contract is asymmetric (backend **produces**, CLI **parses**), so both directions are guarded:

1. **Backend** (`fastapi-backend/tests/test_slim_l9_golden.py::test_knowledge_envelope_serializes_to_golden`): `build_knowledge_envelope(...)` + `KnowledgeWrite.to_payload()`, serialized via `l9.envelope_to_dict`, equals the golden envelope dict (id normalized to the placeholder; produced id separately asserted to be a non-empty string).
2. **CLI** (`mycelium-cli/tests/test_slim_l9_golden.py::test_knowledge_envelope_parses_to_golden_write`): feeding that **same** golden envelope through the CLI's actual parse path — `l9.kind_of` / `l9.payload_data_of`, the two `connector.apply_knowledge_message` reads through — recovers every carried write field. This guards that the CLI reads exactly what the backend writes.

## Contract match

**No drift found.** Verified the backend-produces and CLI-parses shapes agree: the exact envelope `build_knowledge_envelope` emits, parsed by the CLI's `payload_data_of`, yields back the same `key`/`content`/`version`/`base_version`/`created_by`/`updated_by`/`updated_at`.

## Gate results

Both new tests are fast (no SLIM node, no embedding model):
- Backend `test_slim_l9_golden.py`: **7 passed**. `ruff check` / `ruff format --check` / `ty check`: **all pass**.
- CLI `test_slim_l9_golden.py`: **7 passed**. `ruff check` / `ruff format --check` / `ty check`: **all pass**.
- Full CLI suite: **380 passed, 2 skipped**.
- Full backend suite: 261 passed, 7 skipped, **1 pre-existing environment-only failure** unrelated to this change — `test_memory_sync.py::...::test_writes_markdown_and_reindexes_second_store` fails on `PermissionError: /opt/fastembed` (the fastembed model download path is unwritable in this sandbox). It touches none of the files here and exercises the real reindex/embedding path, not the golden guard.

Production code unchanged — test/golden-data addition only.

### DIFF
--- a/fastapi-backend/tests/test_slim_l9_golden.py
+++ b/fastapi-backend/tests/test_slim_l9_golden.py
@@ -25,6 +25,7 @@ import pytest
 
 from app.services import l9
 from app.services.l9_models import Kind
+from app.services.memory_sync import KnowledgeWrite, build_knowledge_envelope
 from app.services.slim_client import (
     _DEV_MASTER_SECRET,
     DEFAULT_CHANNEL_TOPIC,
@@ -89,6 +90,41 @@ def test_exchange_envelope_serializes_to_golden():
     )
     content = {"content": i["text"], "l9": l9.envelope_to_dict(envelope)}
     assert content == g["expected_content"]
+
+
+def test_knowledge_envelope_serializes_to_golden():
+    """The backend's serialized ``knowledge:distillation`` envelope is the golden.
+
+    Step 8 (bible §11) added the memory-sync wire shape: ``build_knowledge_envelope``
+    wraps a :class:`KnowledgeWrite` as a ``knowledge:distillation`` envelope the CLI
+    daemon parses and applies. ``build_knowledge_envelope`` mints a fresh random
+    ``message.id`` per call (it is not part of the shared contract), so the produced
+    id is normalized to the golden's placeholder before the byte-for-byte compare;
+    every other field must match exactly.
+    """
+    g = _golden()["knowledge"]
+    w = g["write"]
+    write = KnowledgeWrite(
+        key=w["key"],
+        content=w["content"],
+        version=w["version"],
+        base_version=w["base_version"],
+        created_by=w["created_by"],
+        updated_by=w["updated_by"],
+        updated_at=w["updated_at"],
+    )
+    envelope = build_knowledge_envelope(
+        room=g["room"],
+        write=write,
+        recipients=g["recipients"],
+    )
+    produced = l9.envelope_to_dict(envelope)
+    # The message id is a fresh UUID per build — freeze everything else, and check
+    # the id is a non-empty string separately (not a shared-contract value).
+    assert isinstance(produced["header"]["message"]["id"], str)
+    assert produced["header"]["message"]["id"]
+    produced["header"]["message"]["id"] = g["message_id_placeholder"]
+    assert produced == g["expected_envelope"]
 
 
 def test_channel_name_topic_matches_golden():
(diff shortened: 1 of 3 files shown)


juliarvalenti:
## Review — golden-knowledge guard ✅ (closes the Step 8 fast-gate gap)

Extends the D2 golden drift guard to the `knowledge:distillation` / `KnowledgeWrite` memory-sync shape that Step 8 added. **Test/data only** (+115/−0, 3 files, no production code).

**Verified:**
- Guards **both directions** of the produce↔parse contract: backend test asserts `build_knowledge_envelope` + `KnowledgeWrite.to_payload()` serialized == golden; CLI test parses the *same* golden via `l9.payload_data_of` (what `connector.apply_knowledge_message` reads) and recovers every write field. So a future backend/CLI drift on the knowledge shape now goes red at the **fast gate** — closing the gap before Step 9 leans on this stream.
- **No drift found** — backend-produces and CLI-parses shapes match exactly (verified live before freezing). That's the key reassurance.
- `MESSAGE_ID_PLACEHOLDER` sentinel + backend-side id normalization is the right approach (the random `message.id` isn't part of the shared contract).

**Notes (non-blocking):**
1. **New D10 item, wire-visible:** the frozen knowledge envelope's sender actor is `"CognitiveEngine"` (role `coordinator`) — the system-actor id is still a **CFN-era name**, and it's on the wire (an agent inspecting a knowledge message sees sender `CognitiveEngine`). Worth renaming (e.g. `system`/`backend`) in the D10 comment/naming audit; the golden will re-lock to the new value.
2. The one pre-existing backend failure the agent hit (`test_writes_markdown_and_reindexes_second_store`, `PermissionError: /opt/fastembed`) is a **sandbox/environment** issue (the Step 8 test hits the real fastembed download path), unrelated to this PR — but flag it as test-hygiene (mock embeddings / mark as needing the model) so the suite is runnable in constrained envs.

**Verdict:** closes the knowledge fast-gate drift gap, no drift, green, no prod code touched. Merging — Step 9 unblocked.

PR #429, opened 2026-08-05
Step 9: cross-machine — the hero flow across two hosts

juliarvalenti:
Picks up **Step 9** from `docs/START_HERE_STEP_9.md` (Step 8 merged). Makes the `join → exchange → converge → plan → work` loop real across **two machines on a LAN** through one shared SLIM node — moderated by host A, with host B a daemon-only member.

Per the step's mandate ("this is plumbing + proof, not a new sync mechanism"), the one genuinely new piece of code is the **reindex-on-arrival** gap closer; the rest is proving the Step 8 transport-agnostic path works over a longer wire and documenting the ops story.

## What changed

**Reindex gap (the sharp one) — `mycelium/daemon/connector.py`**
- On host A the backend's file watcher re-embeds a connector's write; a member host that runs a daemon-only has **no watcher**, so pushed markdown lands but search never sees it.
- `apply_knowledge_message` now returns the `KnowledgeApplyResult`, and on a real write `handle_inbound` calls new **`reindex_after_knowledge(config, room)`** — a best-effort, room-scoped `POST /api/rooms/{room}/reindex` to the co-located backend (the CLI has no local embedder, so reindex is HTTP, exactly like `mycelium memory reindex`). Fires only on an applied write, never on a stale-base no-op.

**Remote endpoint plumbing** — verified + unit-pinned: `run_connector` dials `config.slim.node_endpoint` (a remote LAN node), not the `127.0.0.1` default; `mycelium connect` already persists a normalized endpoint.

**Cross-host consent** — the invite/consent state machine is node-free and addresses the invitee by `workspace/room/agent` **identity only** (no host/endpoint), which is why consent → invite → join needs no new cross-machine mechanism. New test pins it.

**Docs** — `docs/cross-machine.md` (self-hosted shared node default; connector-explicit reindex; open-internet path via hosted node / tunnel — **NAT not built**) and the **Step 10 handoff** (`docs/START_HERE_STEP_10.md`).

## Decisions (flagged, per the step)
- **Hub location:** self-hosted shared node (`mycelium hub host` on the owner); hosted rendezvous is post-MVP. `connect` already accepts `https://` so no code change needed later.
- **Reindex:** connector reindexes explicitly after `apply_knowledge` (option (a) from the trap), rather than relying on a watcher — deterministic and works on a daemon-only host.

## Tests
Fast gate (no node):
- `test_connector_knowledge_sync.py` — reindex fires on apply, **not** on stale-base conflict; helper POSTs the room-scoped endpoint and swallows errors.
- `test_daemon_connector.py` — connector dials the configured remote endpoint.
- `test_mentions_and_invites.py` — moderator invites a remote agent by identity only.

Guarded live-node slice (skipped without a node):
- `test_connector_knowledge_sync_over_slim.py` — a real connector applies a `knowledge` push over the node to a **genuinely separate data dir** (not a swapped `MYCELIUM_DATA_DIR`) and triggers reindex there.

## Verification
- Backend: ruff + format + ty clean; `263 passed, 7 skipped`.
- CLI: ruff + format + ty clean; `385 passed, 3 skipped`.
- All prior guarded slices unchanged.

## DoD note
The headless two-host loop (connect → consent → exchange → converge → plan → synced-and-reindexed memory) is covered by unit + guarded slices here. The full two-laptop / two-container browser demo is the manual acceptance; the **UI inspector + room view + consent prompt** and the legacy SSE/poller retirement are **Step 10** (handoff included).

### DIFF
--- a/docs/START_HERE_STEP_10.md
+++ b/docs/START_HERE_STEP_10.md
@@ -1,0 +1,188 @@
+# START HERE — Step 10 (UI: protocol inspector + room view + consent prompt; retire the legacy SSE poller)
+
+Companion to [`START_HERE.md`](./START_HERE.md). Step 9 is **done** (this PR into
+`slim-native-rewrite`); you are picking up **Step 10**, the **final** step. Read `START_HERE.md`,
+[`START_HERE_STEP_1.md`](./START_HERE_STEP_1.md) → [`START_HERE_STEP_9.md`](./START_HERE_STEP_9.md)
+first if you haven't — the same rules apply. This file gives you (a) the exact state Step 9 left
+behind, (b) your Step 10 marching orders, (c) the facts you must internalize, and (d) the traps
+specific to this step.
+
+**Step 9 made the loop real across two machines.** `mycelium connect` points a member host at a
+shared node host A runs; the connector joins A's moderated room channel; an `@`-invite raises a
+**consent prompt**, and on accept the moderator invites the remote agent **by identity**; the L9
+exchange → aligner converge → `commit:converged` → compiled `plan/tasks.md` → `knowledge` push
+all cross the wire, and the pushed markdown lands **and reindexes** on the member's own store even
+with no file watcher there (the connector reindexes explicitly). **Step 10 makes it visible and
+finishes the demotically:** a UI that shows the room, the live protocol traffic (the L9
+inspector), and surfaces the consent prompt to the human — plus the long-promised retirement of
+the legacy SSE/poller transport now that everything rides SLIM.
+
+## Ground rules (unchanged from START_HERE.md)
+
+1. **The bible is authoritative:** [`slim-native-rebuild-bible.md`](./slim-native-rebuild-bible.md).
+   Your work is **Part V · Step 10**, with **§10 (UI / inspector)**, **§12 (consent)**, and the
+   **Acceptance** hero demo as the design reference.
+2. Leave the project **runnable and green** — backend + CLI **and frontend** quality gates pass at
+   the end.
+3. Code/config blocks in the bible are **reference, not paste**.
+4. **Verify before you edit** — the paths below were accurate at the end of Step 9; confirm shape
+   before changing a file.
+5. Fixed decisions are not up for debate. Retire the legacy SSE/poller **fully** — don't leave a
+   dead endpoint or a half-wired dispatcher behind (see the retirement checklist).
+
+## Where Step 9 left things (your starting state)
+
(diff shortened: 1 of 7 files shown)


juliarvalenti:
## Review — Step 9 gate ✅ (cross-machine; small + focused, exactly right)

Makes the hero flow real across two LAN hosts through one shared node. Correctly scoped as "plumbing + proof" — Step 8's transport-agnostic path already does the sync, so this is deliberately small (+684). Reviewed against bible §16 + Step 9 DoD. Green (backend 263/7, CLI 385/3).

**Verified — the one new piece is right:**
- **`reindex_after_knowledge` (the daemon-only-host gap closer)** — a member host runs no file watcher, so a pushed markdown write would land but never index. `handle_inbound` now reindexes **only on a real applied write** (`result.applied`; never on a stale-base no-op or idempotent loopback), via a best-effort room-scoped `POST /api/rooms/{room}/reindex` (the CLI has no local embedder, so reindex is HTTP — same as `mycelium memory reindex`). Failure is logged + swallowed (markdown is canonical + rebuildable); idempotent, so retry-on-reconnect is safe. Correct.
- **Consent-by-identity** — cross-host consent needs no new mechanism because the invite addresses `workspace/room/agent` (SLIM routes by name, not host). Elegant payoff of the name-based addressing; unit-pinned.
- **Remote endpoint** — `run_connector` dials `config.slim.node_endpoint` (remote LAN node), not the `127.0.0.1` default; pinned.
- **Integration slice** — `test_connector_knowledge_sync_over_slim.py` applies a `knowledge` push over the node into a **genuinely separate data dir** (not a swapped `MYCELIUM_DATA_DIR`) and reindexes there. That's an honest cross-machine sim, not an env trick. 👍
- **Decisions match the bible** — self-hosted shared node default (`mycelium hub host`), hosted rendezvous post-MVP (`connect` already accepts `https://`); explicit connector reindex over watcher-reliance. No golden change (no new wire shape). Correct.

**Non-blocking notes:**
1. **"Daemon-only host" still needs a co-located backend** for reindex/search — the CLI has no embedder, so on a host with *no* backend, memory syncs (markdown lands) but isn't **searchable** (reindex best-effort-fails). Worth a line in `docs/cross-machine.md` so the member-host requirement is explicit.
2. The full **two-laptop / two-container browser demo is manual acceptance** (can't run two physical hosts in CI) — the guarded slice + separate-data-dir sim is the automated coverage. Honest and reasonable; just noting the acceptance is partly manual.

**Verdict:** meets Step 9 DoD, the reindex gap closer is correct, cross-host consent is clean, integration slice is honest, decisions on-bible. Merging — one step left (UI).

PR #431, opened 2026-08-05
Debt D1 (minimal hardening) + D12 (CI safety net)

juliarvalenti:
Pays down two items from the bible's known-debt register. Per the scoping decision, **D1 is a minimal, low-risk hardening** (full JWT/SPIRE identity stays a dedicated post-MVP effort); **D12 is done in full**.

## D1 — env-configurable MLS master secret (partial)

The public `_DEV_MASTER_SECRET` seeds every room's MLS key, so anyone with the repo can join any channel. This makes the secret **configurable** without attempting the full identity overhaul:

- `mint_shared_secret` resolves the master secret from **`MYCELIUM_SLIM_MASTER_SECRET`** — implemented identically in `fastapi-backend/app/services/slim_client.py` and `mycelium-cli/src/mycelium/slim/naming.py` (byte-for-byte matched so moderator + members still derive the same value cross-host).
- Falling back to the dev literal logs a **one-time warning**; **`MYCELIUM_SLIM_REQUIRE_SECRET=1`** makes a host **fail closed** rather than derive keys anyone could reproduce.
- **Default behaviour is unchanged** (dev literal when unset), so the frozen SLIM+L9 golden stays green.
- Tests both sides (override changes the digest; require-secret refuses; env-var names frozen as the cross-package contract).

Still deferred (explicitly, per the bible): per-agent identity + revocation via **JWT/SPIRE** — the production path and the §16 hosted-rendezvous prerequisite.

## D12 — make CI run the safety net

- **New `integration-slim` job**: stands up a `ghcr.io/agntcy/slim:1.4.0` node via `docker run` (GitHub `services:` can't override the command / mount the node config) and runs the guarded **backend + CLI live-node slices** with `MYCELIUM_SLIM_ENDPOINT` set — previously these only ran manually, so CI proved nothing about the live path.
- **Frontend job** now runs the **vitest** component tests (infra landed in Step 10); **CLI job** now runs its **unit suite** (was lint-only).
- **Removed dead backend CI setup** — AgensGraph service, pgvector install, `DATABASE_URL`/`GRAPH_DB_URL`, and the `test_integration.py` step — all defunct since the DB came out in Step 1 (the job referenced a deleted file).
- **Regenerated `mycelium-frontend/package-lock.json`** so `npm ci` resolves the vitest devDeps Step 10 added — the frontend CI job uses `npm ci`, so the prior lockfile/`package.json` mismatch (my Step 10 PR) would have broken CI once the rewrite reached `main`. Validated end-to-end: `npm ci` + `tsc` + `vitest` + `next build` all pass under npm.
- **`test_writes_markdown_and_reindexes_second_store`** stubs embeddings (autouse fixture) so it no longer `PermissionError`s writing `/opt/fastembed` in sandboxes.
- **`test_connect_persists_endpoint`** `delenv`s `MYCELIUM_SLIM_ENDPOINT` (env outranks persisted config in pydantic-settings, so an exported endpoint — set for the guarded slices — shadowed the written value; the long-standing whole-suite wrinkle).

## Scope notes / follow-ups (flagged)

- **CI triggers stay `on: [main]`** (unchanged convention). The safety net therefore gates **per-PR once the rewrite is mainline**. Wiring CI onto the `slim-native-rewrite` branch — and refreshing the `openapi.json` snapshot that earlier steps (incl. Step 10's `/episodes` routes) left stale — is the rewrite **landing** step, out of scope here to keep this PR focused.
- `integration-slim` can't be exercised from my env (no node); the job config + generated slim YAML are validated programmatically, and the chosen slices are CI-designed (mocked agent spawns, no co-located backend). Its first real run will be on mainline.

## Verification
- **Backend**: ruff / format / ty pass; `pytest` → **273 passed, 8 skipped**.
- **CLI**: ruff / format / ty pass; `pytest` → **377 passed, 3 skipped** (incl. `test_slim_config` green with `MYCELIUM_SLIM_ENDPOINT` exported).
- **Frontend**: `npm ci` + `tsc` + `vitest` (8) + `next build` — all green.
- `ci.yml` parses; the heredoc-generated slim node config matches the compose shape.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,21 +11,9 @@ jobs:
     name: Backend tests
     runs-on: ubuntu-latest
 
-    services:
-      agensgraph:
-        image: skaiworldwide/agensgraph:v2.16.0-alpine3.22
-        ports:
-          - 5432:5432
-        env:
-          POSTGRES_USER: postgres
-          POSTGRES_PASSWORD: password
-          POSTGRES_DB: mycelium
-        options: >-
-          --health-cmd "pg_isready -U postgres -d mycelium"
-          --health-interval 10s
-          --health-timeout 5s
-          --health-retries 5
-
+    # No database service: the SLIM-native rebuild removed Postgres/AgensGraph
+    # (memory is markdown + a local JSONL index). The live-node SLIM slices run
+    # in the `integration-slim` job below.
     steps:
       - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
 
@@ -39,16 +27,6 @@ jobs:
         working-directory: fastapi-backend
         run: uv sync --group dev
 
-      - name: Install pgvector in AgensGraph
-        run: |
-          docker exec ${{ job.services.agensgraph.id }} sh -c "
-            apk add --no-cache build-base git &&
-            git clone --branch v0.8.0 https://github.com/pgvector/pgvector.git /tmp/pgvector &&
-            cd /tmp/pgvector && make OPTFLAGS='' && make install &&
-            rm -rf /tmp/pgvector && apk del build-base git
-          "
-          docker exec ${{ job.services.agensgraph.id }} psql -U postgres -d mycelium -c "CREATE EXTENSION IF NOT EXISTS vector;"
-
       - name: Lint
         working-directory: fastapi-backend
         run: uv run ruff check .
@@ -61,19 +39,11 @@ jobs:
         working-directory: fastapi-backend
         run: uv run ty check .
 
-      - name: Unit tests (SQLite)
+      - name: Unit tests
         working-directory: fastapi-backend
         env:
           MYCELIUM_STUB_EMBEDDINGS: "1"
-        run: uv run pytest tests/ -x -q --ignore=tests/test_integration.py
-
-      - name: Integration tests (AgensGraph + pgvector)
-        working-directory: fastapi-backend
-        env:
-          DATABASE_URL: postgresql+asyncpg://postgres:password@localhost:5432/mycelium
-          GRAPH_DB_URL: postgresql://postgres:password@localhost:5432/mycelium
-          MYCELIUM_STUB_EMBEDDINGS: "1"
-        run: uv run pytest tests/test_integration.py -x -q
+        run: uv run pytest tests/ -x -q
 
   openapi-staleness:
     name: OpenAPI snapshot staleness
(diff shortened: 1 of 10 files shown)


juliarvalenti:
## Review — D1 (minimal) + D12 ✅ merge, with two firm caveats

Net-positive and green (backend 273/8, CLI 377/3, frontend `npm ci`+tsc+vitest+build all pass). Several items fix **real latent bugs**, not just debt.

**D1 — minimal hardening, correctly matched:**
- `resolve_master_secret()` prefers `MYCELIUM_SLIM_MASTER_SECRET`, one-time warns on the dev fallback, and **fails closed** under `MYCELIUM_SLIM_REQUIRE_SECRET=1`. **Byte-for-byte mirrored** in `slim_client.py` and `naming.py` (same env-var literals, same precedence) so moderator + members still agree; default path unchanged so the golden stays green. Honestly scoped — full JWT/SPIRE identity remains deferred.
- ⚠️ **New follow-up (D2-class):** the **env-var-name contract is not golden-guarded.** If one package's env-var name later drifts, its own test still passes, the golden (dev-literal digest) still passes, but a host that sets the secret and one that reads a different name derive **different keys → silent join failure**. Freeze `MYCELIUM_SLIM_MASTER_SECRET`/`MYCELIUM_SLIM_REQUIRE_SECRET` in `slim-l9-golden.json` (asserted both sides) so this contract can't drift like the envelope/secret did.

**D12 — real fixes landed, but the headline item is unverified:**
- ✅ Genuinely fixed: frontend vitest + CLI unit suite now run in CI; removed the **dead AgensGraph/pgvector/`test_integration.py`** setup (CI referenced a deleted file); **regenerated `package-lock.json`** (Step 10's vitest devDeps would have broken `npm ci` on main — good catch); `test_writes_markdown_and_reindexes_second_store` stubs embeddings (no more `/opt` `PermissionError`); `test_slim_config` env-isolated.
- ⚠️ **`integration-slim` has never executed.** It's config-validated only (no node in the agent's env), and CI triggers on `main`, so it won't run until the rewrite is mainline. So **the safety net still does not actually run** — D12's core goal is *mechanism-in-place, first-run-pending*. Expect to **debug this job on the mainline landing** (the docker-run node bring-up, endpoint wiring, slice selection are all unexercised). Do not treat D12 as closed until it goes green on a real run.

**Landing checklist this surfaces** (the agent named it): when merging `slim-native-rewrite → main` — wire/confirm CI actually runs on the landing, watch `integration-slim`'s first real run, and **refresh the stale `openapi.json`** snapshot (Step 10's `/episodes` routes aren't in it).

**Verdict:** correct, green, fixes real bugs — merging. **D1 minimally hardened (not closed); D12 implemented but pending its first real CI run.** Both tracked accordingly.

PR #435, opened 2026-08-05
Rung 2 — Pi brain for the internal SAO mediator

juliarvalenti:
## What

Rung 2 of the SAO-mediator track, **plus the Rung-1 live validation and the hardening + correctness fixes it earned.** Gives mycelium's *internal* cognition agent (the SAO mediator) a Pi runtime, retires the `ALIGNER_MODE` flag so the mediator is unconditionally the aligner, validates the whole thing against **real cold-spawned agents**, and fixes the bugs that surfaced — including a real correctness bug where the mediator could misreport the agreement.

### Scope correction (Pi = internal agents only)
An earlier draft said Rung 2 was "swap the *worker* runtime to Pi for all participants." **That was wrong.** Pi is the runtime for mycelium's *internal* agents only (the mediator now). **User/participant agents keep their own framework** (claude_code, cursor, …) — no user-facing `pi` adapter, no change to `AGENT_ADAPTERS` or the daemon dispatch path.

## Commits (by theme)

**Rung 2 — Pi brain**
- `PiBrain` (`app/services/pi_brain.py`): synchronous `llm_sync`-compatible callable driving a persistent `pi -p --session <id> --mode json` process (session memory across SAO rounds), default-off OpenShell sandbox seam, `ALIGNER_BRAIN` selection in `aligner.py:_make_brain`. Default `litellm` → additive, stays green. Node-free tests.

**Rung 2.5 — retire `ALIGNER_MODE`**
- A summon now unconditionally drives a live NEGMAS SAO; no observer/driver/mediator flag. The mediator *is* the aligner. (observer/driver methods survive only for the live roundtrip scripts, pending Rung-3 deletion.)

**Reliability fixes (found by validating live)**
- **Single-wake**: the mediator addresses one agent per turn via L9 recipients, but its prompt text embedded the broker summary naming *both* agents, and the connector wakes on a raw `@handle` in text — so every turn woke *everyone*, piling up cold-spawns until real replies missed the round window. Fix: neutralize `@` in outgoing prompts.
- **Fail-closed**: a timed-out (empty) reply was fed to `interpret()`, which *hallucinated* a full offer. Now empty prose short-circuits to reject/hold — never the LLM.
- **Room-visibility**: the mediator now records its turn-prompts via `persister.ingest_local` so the negotiation is followable in the room, not just backend logs.

**Faithful interpretation (the correctness fix)**
- The worst bug: the mediator could **ship an agreement the agents never made** (they bargained to 30% live, it recorded 25%, which compiles into `plan/tasks.md`). Root cause: `discover_issues` fixed a discrete option grid without 30 on it, so `interpret` snapped the agreed value to a wrong option.
- `app/services/offer_snap.py` (new): stdlib fuzzy snap (exact → case-insensitive → normalized → token-containment → `difflib`) rescuing LLM near-misses (`"30%"→"30"`) onto canonical options, and **refusing to fabricate** an out-of-grid value (stays a real reject). No new deps.
- `mediator.py:to_outcome` snaps instead of rejecting near-misses; `discover_issues` now generates an evenly-spaced numeric grid spanning the agents' range *and the middle*, and is told not to invent phantom issues.
- Inspired by the clean part of `outshift-open/ioc-scale-cf-cognition-engines` (`offer_validation.py`) — **not** its SAV validator.

## Live validation (Gate A — real cold-spawned agents)

Proven end-to-end against two real `claude_code` agents:
- **Converged and terminated**: `{tech: 40%, cap: 25%}`, `agreement in 1 step`, single wakes (`@growth` then `@risk`, separate), replies in ~10s (no timeouts). The anti-theatre property, live.
- A forced multi-round bargain (`bargain2`) had the agents genuinely walk to a middle over several exchanges — and is what exposed the discretization correctness bug (now fixed) plus the remaining open items below.

## Still open (documented in `docs/START_HERE_MEDIATOR_RUNG2_VALIDATION.md`)
- Confirm the granular-grid `discover_issues` change on a live run (the snap logic is fully unit-covered; the grid is a prompt tweak).
- The mediator's own messages (prompts + verdict) reach the persister log but not the `/messages` API store — negotiation not yet followable in the room/UI.
- Agents cave below stated hard lines under the mediator's BATNA push (`mediator.py:_BATNA`) — wants a soften-able knob.
- Gate B (Pi brain live) + OpenShell sandbox.

## Tests
`uv run pytest tests/test_pi_brain.py tests/test_mediator.py tests/test_aligner.py tests/test_offer_snap.py -q` → **36 passed**. ruff + ty clean. (Unrelated pre-existing failure: `test_l9_over_slim_roundtrip` hits a `/opt/fastembed` permission error in the local sandbox.)

## Out of scope
No user-facing `pi` adapter; no change to `AGENT_ADAPTERS` / `daemon/dispatch.py` / `runner.py`; `claude -p` / cursor runtimes NOT retired. Wake-up/invocation UX stays a separate per-framework concern. No SAV / drift-validator engine.


juliarvalenti:
## Gaps found running the stack locally (rung2-pi-openshell)

Three things that blocked a clean first-run:

**1. `pi` binary not in the container image**
`ALIGNER_BRAIN=pi` is wired end-to-end but the Dockerfile doesn't install `pi` — the aligner wakes, hits `pi_brain.py`, and crashes with `` `pi` not found on PATH ``. The branch name promises Pi but you can't actually run it without a manual workaround.

**2. No CLI command to post a message / summon the aligner**
`mycelium agent invoke` only works for registered agents. The built-in `@aligner` handle has no equivalent CLI surface — you have to hit the REST API directly (`POST /api/rooms/{room}/messages`) with the right `message_type` enum value, which isn't documented anywhere obvious.

**3. `mycelium config set` doesn't expose aligner settings**
`ALIGNER_BRAIN`, `ALIGNER_PI_OPENSHELL`, etc. have to be set by editing `~/.mycelium/.env` manually. `mycelium config set aligner.brain pi` fails with `'MyceliumConfig' object has no attribute 'aligner'`.

juliarvalenti:
**Gap: openshell not installed in container**

`ALIGNER_PI_OPENSHELL=true` fails with `FileNotFoundError: No such file or directory: 'openshell'` — the binary isn't bundled in the Dockerfile. Disabling for now (`ALIGNER_PI_OPENSHELL=false`) to unblock Pi-brain runs.

juliarvalenti:
## Gaps found running rung2-pi-openshell locally

**1. `pi` binary not in the container image**
`ALIGNER_BRAIN=pi` wired end-to-end but Dockerfile doesn't install `pi` — aligner wakes, hits `pi_brain.py`, crashes with `` `pi` not found on PATH ``.

**2. pi requires its full distribution directory, not just the binary**
Copying only the `pi` binary isn't enough — Pi crashes at startup looking for theme assets relative to the binary (`ENOENT` on `dark.json`/`light.json`). The full release tarball unpacks a directory (binary + `assets/`, `theme/`, `node_modules/`, `photon_rs_bg.wasm`, etc.). The Dockerfile install step needs to extract the whole tree to e.g. `/opt/pi/` and symlink `/usr/local/bin/pi → /opt/pi/pi`.

**3. openshell not installed in container**
`ALIGNER_PI_OPENSHELL=true` fails with `FileNotFoundError: No such file or directory: 'openshell'` — binary not bundled.

**4. No CLI command to post a message / summon the aligner**
`mycelium agent invoke` only works for registered agents. The built-in `@aligner` handle has no CLI surface — you have to hit `POST /api/rooms/{room}/messages` directly with the right `message_type` enum, undocumented. (`mycelium engine invoke` is Stage B, not yet implemented.)

**5. `mycelium config set` doesn't expose aligner settings**
`ALIGNER_BRAIN`, `ALIGNER_PI_OPENSHELL`, etc. require manual `.env` edits. `mycelium config set aligner.brain pi` fails with `'MyceliumConfig' object has no attribute 'aligner'`.

PR #418, opened 2026-08-04
SLIM-native rewrite — one messaging node, no database, secure agent coordination (draft · do not merge)

juliarvalenti:
# SLIM-native rewrite — one messaging node, no database, agents that coordinate over a secure fabric

A ground-up rebuild of Mycelium's coordination + memory substrate. It replaces a 4-service closed-source "CFN" cluster and a Postgres/AgensGraph database with **a single `slim` messaging node, a thin FastAPI backend, and local markdown/JSONL files — no database at all.** Agents coordinate as members of secure, end-to-end-encrypted rooms; the human is just another participant; and a negotiation that reaches consensus compiles into a shared plan and syncs to memory.

> `main` was intentionally *not* the source of truth during the rebuild — `slim-native-rewrite` is. This is the cumulative diff vs `main`.

## How it works
- **Rooms are SLIM group channels.** The always-on backend is each room's moderator; agents (and the human, by proxy) are members. Messages are L9 envelopes — `exchange` / `commit` / `knowledge` — carried over MLS-encrypted multicast, so the node forwards only ciphertext.
- **Participation is a CLI primitive, not a daemon.** Being a room member — receiving coordination and replying — is `mycelium await` / `mycelium respond`: any already-awake caller (a Claude Code session, a subagent, a shell script) joins a room's channel as a genuine SLIM member and coordinates directly. The **daemon is now an optional auto-waker** for runtimes that can't wake themselves — it cold-spawns `claude -p` on a mention — and it is rebuilt on the *same* membership core, so the CLI path and the daemon path can't drift. Agents never speak SLIM or L9.
- **Consent is first-class.** Inviting a *foreign* agent into a room raises an accept/decline prompt (visible in the UI); a user's *own* registered agents are pre-authorized and join directly.
- **Negotiation runs in episodes.** Summoning the aligner opens an *episode* — a tagged, membership-scoped negotiation on the room channel with its own record. It scores the room's positions (MPC/GAR/SCR); on convergence it emits `commit:converged`, the backend compiles a shared `plan/tasks.md`, and that plan syncs as a `knowledge` memory. Each convening is a distinct, addressable episode (unique id — no clobber, its own transcript slice, live in-progress status), surfaced in the UI protocol inspector.
- **State is just files.** Rooms are folders, memories are markdown, the search index is JSONL — all under `~/.mycelium`. Sharing is git; cross-machine is the same channel over a shared node.

## How it's different
| | Before | After |
|---|---|---|
[...]

PR #438, opened 2026-08-12
feat(spoke): propagate room deletion to spoke nodes

sfph:
## Summary

- **Hub**: fires a `room_deleted` SSE tombstone after the DB commit (`rooms.py:_notify_room_deleted`), so spoke daemons and OpenClaw agents are notified the moment deletion is committed
- **Daemon (Python path)**: `subscribe_room` handles `room_deleted` events and bounded 404 retries (3 × 15 s) via `_handle_room_deleted`, which sets the tombstone, optionally removes the local directory (when `auto_gc_orphaned_rooms = true`), updates `daemon.toml`, and unregisters adapter configs — all in one canonical place
- **OpenClaw (TS path)**: `room-sse.ts` matches the Python 404-retry logic; `route.ts` dispatches a human-readable cleanup instruction to each agent on `room_deleted`; both delivery paths behave identically
- **Startup reconcile**: `reconcile_local_rooms` catches rooms deleted while the spoke was offline, seeds `state.rooms_deleted` before any SSE tasks are created, and filters `rooms_configured` so the health endpoint never reports orphaned rooms
- **`mycelium room gc`**: new CLI command to list or prune orphaned local room directories; uses paginated bulk GET to avoid N+1 hub requests; `--json` output is clean (all human-readable `secho` calls gated on `not json_output`)
- **`mycelium doctor`**: `check_orphaned_rooms` implemented with the same paginated bulk fetch; uses `get_mycelium_dir()` so `MYCELIUM_DATA_DIR` is respected
- **Tombstone invariant**: tombstones are set by 404-after-retries, `room_deleted` events, or startup reconcile — never cleared at hot-reload, only on daemon restart when a fresh reconcile can confirm hub status with a live request
- **Rolling-restart safety**: 404 retry counter resets on any non-404 ≥ 400 response so interleaved 5xx responses don't accumulate toward the deletion threshold (both Python and TS paths)

## Test plan

- [ ] Delete a hub room while spoke daemon is running — verify daemon logs `room_deleted received`, removes dir if `auto_gc_orphaned_rooms = true`, removes room from daemon.toml, daemon SIGHUP causes health to no longer list it
- [ ] Delete a hub room while spoke is offline — restart daemon, verify `reconcile_local_rooms` detects orphan, seeds tombstone, no 404-retry SSE task spawned
- [ ] Simulate rolling restart (hub returns 500 then 404) — verify 404 counter resets on 500, daemon does not tombstone prematurely
- [ ] Run `mycelium room gc` with and without `--prune-orphans` and `--json` — verify JSON output contains no stray text
- [ ] Run `mycelium doctor` — verify `check_orphaned_rooms` lists orphans without N+1 requests
- [ ] OpenClaw: delete a room — verify agents receive the `room_deleted` instruction string, SSE loop exits

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ pnpm-lock.yaml
 
 # Claude Code harness state (per-session worktrees, transcripts)
 .claude/worktrees/
+*.swp
--- a/fastapi-backend/app/routes/messages.py
+++ b/fastapi-backend/app/routes/messages.py
@@ -79,9 +79,7 @@ async def _notify_room(channel: str, payload: dict) -> None:
 async def fire_room_notify(room_name: str, payload: dict) -> None:
     """Fire a Postgres NOTIFY on ``room:{room_name}`` without persisting a DB row.
 
-    Used by routes that need to push a synthetic event (e.g. ``room_deleted``)
-    before the room row is torn down.  Non-fatal: the SSE delivery is
-    best-effort and must never block the delete path.
+    Non-fatal: the SSE delivery is best-effort and must never block the caller.
     """
     await _notify_room(room_channel(room_name), payload)
 
--- a/fastapi-backend/app/routes/messages.py
+++ b/fastapi-backend/app/routes/messages.py
@@ -76,6 +76,16 @@ async def _notify_room(channel: str, payload: dict) -> None:
                     logger.warning("NOTIFY failed for %s: %s", channel, e)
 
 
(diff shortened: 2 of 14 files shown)


juliarvalenti:
Superseded by the SLIM-native rewrite (#418): built on removed subsystems (see the diff). The salvageable idea is tracked in #443 — closing this stale PR; the branch stays in git history as the reference implementation.

Issue #472, opened 2026-08-13
docs: introduce "engines" as a concept (not just the aligner)


juliarvalenti:
> Part of #187 (Epic: User-configurable cognition engines).

## Problem

The docs present the **aligner** as *the* cognition engine. The `mycelium engine create … --kind aligner` / `engine invoke` surface only ever appears inside aligner content (`docs/aligner.md`, `architecture.md:112` "The aligner **is** the cognition engine"), so a reader has no way to learn that:

- **engines** are a general concept — first-party cognition citizens registered per room and summoned explicitly, and
- the aligner is just **one `--kind`**, with the command surface (`--kind …`, per-room registration, invoke-to-run) deliberately built for more kinds.

The `--kind aligner` flag is the only hint the plurality exists, and it's unexplained.

## Ask

- [ ] Add an **Engines** concept section/page: what an engine is (first-party cognition citizen, registered per room, summoned not auto-run), the shared `engine create --kind <kind>` / `engine invoke` lifecycle, and that the aligner is one kind among a growing set.
- [ ] Reframe `aligner.md` + `architecture.md` so the aligner is presented as *an engine kind* (the mediator/negotiation kind), not synonymous with "the cognition engine."
- [ ] Cross-link from `overview.md` / `index.md` so the concept is discoverable before the aligner specifics.
- [ ] Leave room for future kinds (e.g. the synthesizer in #471) without enumerating unbuilt ones.

Consumer-first, concept + why before mechanics (per the docs-site conventions). Regenerate via the docs skill after editing source.

Refs #187, #471.

PR #473, opened 2026-08-13
feat(engine): synthesizer engine kind (#471)

### LINKED ISSUE #471: Reintroduce synthesis as an invocable engine (not an auto-triggered system)
> Part of #187 (Epic: User-configurable cognition engines) — the synthesizer is a concrete second engine kind alongside the aligner.

## Intent

The old memory-synthesis system was removed (#333) because it was an implicit, auto/threshold-triggered backend service (`room synthesize` command, `mycelium catchup`, trigger configs). The remaining dead display residue is cleaned up in #458 / #470.

Rather than resurrect that model, **reintroduce synthesis as a first-party engine** — the same shape as the aligner: a room citizen you register and *explicitly invoke*, not something the backend fires on a threshold.

## Why an engine

The engine seam already exists and is the right home for room-scoped cognition:

- `mycelium engine create synthesizer --kind synthesizer` — register it as a room citizen.
- `mycelium engine invoke synthesizer "…"` (or `@synthesizer` mention) — run it on demand.
- Runs where the daemon runs; heavy deps opt-in via the `[engine]` extra; brain is a one-shot **Pi** turn like every other mycelium cognition call (`plan_compiler.py` is the closest template — a throwaway `pi` session off the event loop).

This keeps synthesis **explicit, inspectable, and pull-based** — consistent with the protocol's design principle that coordination is invoked, not auto-run on a join window (see the aligner rationale in CLAUDE.md).

## Sketch

- [ ] New engine kind `synthesizer` alongside `aligner` (engine registry + `--kind` validation).
- [ ] On invoke: read the room's memory namespaces (`decisions/`, `status/`, `context/`, `work/`, `procedures/`, `log/`, `plan/`) and materialize a **structured summary** as a `knowledge` memory (`memory set`, so it's upserted + indexed like everything else). No new storage concept.
- [ ] Pull-based only — no trigger configs, no `synthesis_complete` NOTIFY events, no backend threshold firing. The removal in #333 stays removed; this is a new, opt-in engine.
- [ ] Decide output surface: a dedicated `plan/`-adjacent summary file vs. a `context/` knowledge memory. (Prefer reusing an existing namespace over inventing one.)
- [ ] Docs: add to the engines guide (Stage B alongside the aligner), consumer-first.

## Non-goals

- Not reviving the auto/threshold trigger model or the `catchup` command.
- Not touching the aligner's **reply-envelope synthesis** (L9 `record_reply`) — unrelated, load-bearing, stays as-is.

Refs #333 (original removal), #458 / #470 (residue cleanup).

juliarvalenti:
**Draft for review.** Implements #471 — synthesis as an *invocable engine* (not the removed auto-triggered system).

A second cognition-engine `kind` alongside the aligner. On `@`-summon it reads the room's memory namespaces, compiles a structured markdown briefing via one throwaway `pi` turn, and upserts it as a `knowledge` memory at `context/synthesis` — searchable + version-tracked like any memory.

### What's here
- **`protocol.py`** — register `synthesizer` in `ENGINE_KINDS`. CLI `engine create --kind synthesizer` / `engine invoke` validation is already data-driven off this set, so no command changes needed.
- **`app/services/synthesizer.py`** — `SynthesizerEngine`:
  - dormant-by-default summon gate (reuses the aligner's `_registered_engine_kind` / `_norm` primitives — one source of truth for the manifest read),
  - one-shot Pi cognition mirroring `plan_compiler._pi_complete`,
  - **fail-soft**: a Pi outage logs and writes nothing (no half-baked summary),
  - writes through the canonical `create_memories` path (versioned, indexed, NOTIFY),
  - excludes `agents/` manifests and the prior summary from its own corpus.
- **`app/config.py`** — `SYNTHESIZER_HANDLE` + `SYNTHESIZER_PI_TIMEOUT_S`; reuses the shared `LLM_*` / `ALIGNER_PI_*` Pi runtime.
- **`app/main.py`** — a fan-out summon dispatcher so the aligner and synthesizer share the single `on_summon` seam; each self-selects by manifest kind (only one ever acts per summon).

### Tests
- CLI: `synthesizer` accepted as a kind.
- Backend: summon gate (fires only for a registered synthesizer, not a teammate or the aligner), host-runtime skip, and the read → summarize → write path incl. fail-soft + corpus exclusions.
- Full suites green: backend 376 passed / 1 skipped, CLI 417 passed. ruff + ty clean.

### Design notes / open questions for review
- **Output surface**: I chose `context/synthesis` (existing namespace, stable slug so it upserts). The issue floated `plan/`-adjacent vs `context/` — easy to change.
- **No SLIM broadcast yet**: the write goes through `create_memories` which already fires a NOTIFY; I did not add a `knowledge`-envelope broadcast like `plan_sync` does. Can add if we want live push to members.
- **Corpus scope**: currently all namespaces except `agents/` (includes `log/`). Could trim `log/episodes/` if summaries get noisy.

Still to add (per follow-up): seed the synthesizer into the UI mock data + the "Run a sample coordination" demo. Docs concept work is tracked in #472.

Closes #471

### DIFF
--- a/docs/adapters.html
+++ b/docs/adapters.html
@@ -311,8 +311,6 @@
 
     <hr class="divider">
 
-
-
 <section class="doc-section" id="adapter-api">
       <!-- keep -->
       <h1>REST API</h1>
--- a/docs/concepts.html
+++ b/docs/concepts.html
@@ -4,9 +4,9 @@
 <meta charset="UTF-8">
 <meta name="viewport" content="width=device-width, initial-scale=1.0">
 <title>Concepts · mycelium</title>
-<meta name="description" content="The core concepts behind Mycelium: rooms, episodes, memory, plan, the aligner, and the L9 protocol.">
+<meta name="description" content="The core concepts behind Mycelium: rooms, episodes, memory, plan, engines (the aligner and synthesizer), and the L9 protocol.">
 <meta property="og:title" content="Concepts · mycelium">
-<meta property="og:description" content="The core concepts behind Mycelium: rooms, episodes, memory, plan, the aligner, and the L9 protocol.">
+<meta property="og:description" content="The core concepts behind Mycelium: rooms, episodes, memory, plan, engines (the aligner and synthesizer), and the L9 protocol.">
 <meta property="og:image" content="https://mycelium-io.github.io/mycelium/og.png">
 <meta property="og:url" content="https://mycelium-io.github.io/mycelium/concepts.html">
 <meta property="og:type" content="website">
@@ -52,6 +52,7 @@
       <a href="#episodes" class="nav-link sub">Episodes</a>
       <a href="#memory" class="nav-link sub">Memory</a>
       <a href="#plan" class="nav-link sub">Plan</a>
+      <a href="#engines" class="nav-link sub">Engines</a>
       <a href="#aligner" class="nav-link sub">Aligner</a>
       <a href="#l9-protocol" class="nav-link sub">L9 Protocol</a>
     </div>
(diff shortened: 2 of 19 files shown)


juliarvalenti:
Added the synthesizer to the UI mock data + sample tour (commit 263eb3d7):
- **fixtures**: `agents/synthesizer` registered in both mock rooms; a `context/synthesis` briefing artifact seeded in atlas-migration so the memory inspector shows the synthesizer's output; synthesizer added to the collector metrics. (Also registered the previously-missing `agents/aligner` manifest in the pricing-model demo room.)
- **stream**: the pricing-model demo now summons the synthesizer after consensus+plan and posts its distilled briefing to the channel.
- **tour**: new 'Summon the synthesizer' step after the plan step (convergence-sync indices unchanged).

Frontend `tsc`, lint, and vitest all green.

juliarvalenti:
Added the **Engines** concept doc + live-verified against a real backend (commits 799fe354, 152fc542):

**Docs (closes #472):** New `engines.md` concept page — engines as first-party cognition citizens, summoned not auto-run, with the aligner and synthesizer as two `kind`s under one shared `engine create/invoke` lifecycle. Reframed the aligner page + architecture blurb so the aligner is *the negotiation engine kind*, not 'the cognition engine'. Synthesizer documented inline as a kind. Docs site regenerated (also picked up some pre-existing staleness: litellm→Pi row, `room delete` multi-arg).

**Real-backend verification:** Rebuilt the backend from source, ran `mycelium engine invoke synthesizer` end-to-end (CLI → SLIM summon → backend → `context/synthesis` memory). Real Pi/Anthropic call produced a faithful briefing. Also added a guarded live test (`MYCELIUM_LLM_TESTS=1`).

Side effect: found & fixed a config bug where `config apply` dropped `ENGINE_RUNTIME` from `.env` — shipped separately as #474 (merged).

juliarvalenti:
https://github.com/user-attachments/assets/5e7b41c8-f98a-4cf3-bb85-fe17d855de7a

Issue #493, opened 2026-08-14
Epic: Live coordination surface — the "Needs you" cockpit (CLI + GUI), tied to GitHub by reference


juliarvalenti:
## Thesis
The plan surface is mid because it's a flat, compiled markdown checklist (`plan/tasks.md`) frozen from a negotiation and disconnected from real work — a document, not a surface. Replace it with a **live projection of the event ledger** (concerns/actions/`source_event`s over L9): self-populating, transient, agent-native, steerable. It's the live execution layer that sits *beneath* GitHub Issues — not a competitor to it.

## Principles (what keeps it from being a todo app *or* a worse GH Projects)
- **Self-populating from reality** — herdr, PRs, CI, and agents fill it. The human doesn't groom a backlog; they steer exceptions.
- **Default lens = "Needs you"** (the steer-lens); the firehose ("All") is opt-in.
- **One-gesture triage**, and the *same verbs for humans and agents*: claim / resolve / block / promote / dismiss / capture.
- **Transient by design** — the `event` primitive's TTL means the board self-heals and can't rot into a backlog.
- **Agents are first-class rows** (peers), with live presence + jump-in (herdr).
- **Tied to GitHub by reference, not identity** (see below).

Design mechanics stolen from fast task apps, aimed at *triage* (capture is done by events/agents): Todoist/Superlist natural-language capture, Any.do one-gesture state changes, Linear keyboard-first, Microsoft To Do "My Day" = the steer-lens.

## CLI surface
```
$ mycelium board                      # default = the "needs you" steer-lens
  handshake · 3 need you · 4 in flight · 6 resolved today

  ⚠ NEEDS YOU
   d3f  decision  agent-y: JWT TTL — 15m or 60m?              ↳ reply
   a91  blocked   "thin-spoke join"  waiting on #502          40m
   7c2  review    agent-z opened PR #504, wants eyes          CI ● green

  ◐ IN FLIGHT
   b12  auth → JWT         🤖 agent-y   feat/jwt · CI● · PR#502   ⤷ jump
   e45  cache TTL sweep    🧑 julia     feat/cache · CI◐          ⤷ jump

  (c)laim (r)esolve (b)lock (p)romote→gh (x)dismiss  (/)capture  (w)atch

$ mycelium claim b12 --to agent-y --dispatch   # assign + herdr spawns its runtime
$ mycelium promote d3f                          # → files a GitHub issue, drops from live
[...]

Issue #476, opened 2026-08-13
SLIM channel identity: per-member signing credential for MLS (SignerJwt floor, SPIRE recommended — #579)


juliarvalenti:
## Move SLIM identity beyond the shared-secret PSK (JWT / JWKS / SPIRE)

**Type:** enhancement / hardening — hard prerequisite before anything hosted or multi-user.

### Current state

SLIM app identity today is a **deterministic shared-secret PSK**: a per-channel secret is `HMAC-SHA256(master_secret, "workspace/room")`, derived independently by every host so no key-exchange round-trip is needed. Passed to the fabric via `service.create_app_with_secret(name, secret)`.

- Master secret: `MYCELIUM_SLIM_MASTER_SECRET`, else a **public dev literal** committed in-repo (`mycelium-dev-shared-secret-v1-do-not-use-in-prod`) — anyone with the repo can derive it, so it authenticates nothing on its own.
- `MYCELIUM_SLIM_REQUIRE_SECRET=1` fails closed on a host that forgot to set a real secret.
- Source: `fastapi-backend/app/services/slim_client.py` (~L61-75, `mint_shared_secret`/`resolve_master_secret`) and the CLI mirror `mycelium-cli/src/mycelium/slim/naming.py`.
- CLAUDE.md already flags this: *"real identity (JWT/SPIRE) is a hard prerequisite before anything hosted / multi-user."*

### Gaps this creates

- **No per-agent identity.** The secret is scoped to `workspace/room` — the agent segment is ignored — so every member of a room is cryptographically indistinguishable.
- **No revocation.** Removing an agent's access means rotating the room/master secret for everyone.
- **Shared-secret parity trap** across hosts (documented in `docs/cross-machine.md`): every host must set the *same* private master secret out of band.

### What the fabric already supports (from the audit)

slim-bindings exposes a full identity stack we don't touch — verified in the upstream examples (`_slim-research/slim-bindings/python/examples/common.py`):

- **JWT + JWKS** — `IdentityProviderConfig.JWT` / `IdentityVerifierConfig.JWT`, `ClientJwtAuth`, `JwtKeyConfig` (`JwtAlgorithm.RS256`, `JwtKeyFormat.PEM|JWKS`, `JwtKeyData`), audience/issuer/subject/duration. Identity = the `sub` claim; per-agent identity falls out naturally.
[...]

Issue #450, opened 2026-08-13
Epic: Bake the framing into the product, top-down (humans + agents in your SDLC) — not a doc


juliarvalenti:
The framing has moved, and the way we land it is **through the product, top to bottom — not a positioning doc.** No `POSITIONING.md`, no manifesto page. The application itself — every surface a user touches — teaches what Mycelium is. If the framing only lives in a doc, it isn't real, and it's cringe.

## The framing every surface should express
Mycelium is where **humans and agents coordinate inside the SDLC you already use.** The impetus to install it: your engineers (and their agents) move too fast to stay in sync — every hand-off (human↔human, human↔agent, agent↔agent) is a context rebuild, and that tax scales with team speed. Mycelium is the shared feed that pays it down — no new editor, no new workflow.

- **Foot in the door:** team comms — GitHub/PR/CI state and agent breadcrumbs ("heads up, refactoring auth") flowing into the room feed (the `event` primitive). Rides the tools you already use.
- **Back pivot:** the same feed becomes where agents coordinate *themselves* — claim/resolve `concern`s. Human-in-the-loop → human-on-the-loop.
- **Substrate:** a typed, causal, encrypted ledger (L9 over SLIM) — same rails at every autonomy level. The IoC bet, earned not claimed.
- **Negotiation is not the product.** The aligner is optional cognition the feed *invokes*; anti-theater is a *guarantee* for when it runs, not the headline. (Kernel guarantees + the GAR/SCR theater-meter are subordinated here — tracked via #451.)

## The work: bake it in, top-down
Not a doc — every surface a user actually touches:
- [ ] **README** — hero + "how it works" lead with the human+agent SDLC-coordination feed, not negotiation.
- [ ] **Docs site / concepts** — reframe around the feed + event ledger; demote the aligner to invoked cognition.
- [ ] **CLI** — command descriptions, `--help`, and `mycelium docs` text speak the framing (feed, events, breadcrumbs — not "negotiation").
- [ ] **UI copy** — labels, empty states, onboarding, first-run: the feed is the center of gravity; presence + events are the story.
- [ ] **Naming** — concept names (rooms, feed, events, engines) reflect coordination-in-your-SDLC, not multi-agent negotiation.
- [ ] **First-run / install** — the moment someone installs, what they see teaches the framing (a live feed of humans + agents), never a negotiation demo.

**The test:** a new user learns what Mycelium is **by using it** — never by reading a manifesto.

[...]

Issue #446, opened 2026-08-12
Spike: mycelium × herdr — persistent agent runtimes as a coordination substrate


juliarvalenti:
## Spike: does herdr solve mycelium's persistent-runtime gap, and how would we fit in?

**Type:** exploration / de-risking spike (not a commitment to build)

### Why this matters

Mycelium's hardest unsolved problem is runtime persistence. Coding agents (Claude Code, Cursor) are **ephemeral**: they exist only while a session is live, the daemon's `claude -p` cold-spawn throws away all context every turn, and once you close the laptop there's nothing to wake — a tag "goes to the aether." We keep wanting "drop a bunch of agents in a room and coordinate," but that only feels real when the agents are **persistent and addressable**, not re-spawned from zero.

[herdr](https://herdr.dev) ("terminal workspace manager for AI coding agents") is exactly that missing layer: a headless server that keeps agent sessions alive across detach, organizes them into workspaces/tabs/panes, recognizes ~24 coding-agent kinds, and exposes a **socket API** to drive them. The clean division of labor:

- **herdr = where agents *live*** — persistent runtime, addressable, wakeable, survives detach.
- **mycelium = how agents *coordinate*** — rooms, the SLIM channel, shared memory, the aligner.

Non-overlapping. Complementary bets, not competing ones.

### What I verified live (from a herdr-managed pane)

Drove a real persistent Claude agent through herdr's API and confirmed the make-or-break motions:

- `herdr agent prompt <pane> "<text>" --wait` — atomically injects a prompt + Enter into a **persistent** agent and blocks until it settles (`idle`/`done`/`blocked`). Watched the agent's context grow `0% -> 3.9%` and settle to `idle`. This is the exact motion the daemon's wake path needs.
- `herdr agent list` / `get` — live lifecycle state as JSON (`idle`/`working`/`blocked`/`done`/`unknown`). Observed one agent `working` on a long task while others sat `idle`.
- Recognized kinds include **`pi`, `claude`, `cursor`, `hermes`** — literally the mycelium cast (Pi brain, our proven adapter, the untested one, a retired one).
- **Limitation found:** `herdr agent read` cannot scrape Claude Code's reply text — Claude runs a full-screen (alt-screen) TUI, so the response never lands in herdr's scrollback. All read sources return only the input box + status bar.

[...]

PR #494, opened 2026-08-14
Coordination board (Needs-you cockpit) — ideation

juliarvalenti:
## Summary

Adds a **coordination board** — a live projection over the room's event ledger — across the backend, CLI, and GUI. It replaces the flat `plan/tasks.md` surface (the Plan tab) with a lens-based board of concern/action rows (issue #493).

This is an **exploratory / ideation branch** (draft). It is wired end-to-end and verified live, but the design is still moving — see the pinned framing comment below for where this is headed.

## What's here

**Backend** (`fastapi-backend`)
- `app/services/board.py` + `app/routes/board.py`: `GET /rooms/{room}/board` projects three sources into one row list — the stateful `action`/`concern` event ledger (#392), the compiled plan checklist, and the live episode — with live presence, plus the triage verbs (`claim`/`block`/`resolve`/`promote`/`dismiss`) and NL `capture`. Richer board state + display fields ride in the event `payload`, so the shared #392 primitive is untouched.
- Verbs are the same vocabulary agents already drive over the ledger (agent parity for free).
- 10 route/projection tests.

**CLI** (`mycelium-cli`)
- `mycelium board` — default "Needs you" lens, `--all`, `--watch`; the five verbs + `capture`; short-id prefixes resolve to the full item. 7 tests.

**Frontend** (`mycelium-frontend`)
- `RoomBoard` replaces the Plan tab: three lenses (Needs you / In flight / Resolved), rows rendered as stateful objects with expand-on-demand detail, one-gesture triage, ⌘K capture. `board_updated` SSE nudges a live refresh; `dev:mock` has a board fixture.

## Status / honesty

- Ledger items are in-memory; plan rows persist on disk.
- `promote` stamps a GitHub back-link and resolves the row, but does **not** file the issue yet (needs `gh`).
- CI/PR status render when present in a row's payload, but nothing populates them yet (herdr/CI bridge, #446).

## Not for merge

Draft. The point of this branch is to think, not to ship. Framing/vision is tracked in a comment below rather than in the issue.

Related: #493, #446, #450, #301, #476


juliarvalenti:
## Framing / vision (tracked here, deliberately not in the issue)

This branch started as "replace the plan surface with a board" and drifted into a longer conversation about what this thing actually *is*. Capturing that here so it doesn't evaporate. This is the north star; the code above is only partway to it.

### The axis

It's a **coordination layer for teams with more agents in flight than any one person can track.** GitHub stays the source of truth for what needs doing — we don't replace or mirror it. What we add is the layer where agents pick up that work and coordinate with each other to get it done, over a shared message bus (the SLIM `await`/`respond` + L9 spine already exists). The shift for the human is the whole point: **you go from dispatching tasks and reviewing everything yourself to steering the exceptions the agents can't resolve on their own.**

Day to day it's a live board of work in progress, except the cooks are agents. You watch them claim tickets, hand off, hit blockers against real GitHub issues. When one does something that needs a human — changes an authz model it isn't sure about, hits a call above its pay grade — it **raises that through the fabric and the exception lands in front of the right person.** You're not digging through diffs or sitting in a review queue guessing; the coordination routes the judgment calls to you.

### What it is deliberately *not*

- **Not a memory engine.** Memory is underneath, feeding agents context — but we don't compete on retrieval (we'll lose that benchmark race and shouldn't play). Memory is fuel, not the product.
- **Not a wiki / Jira / board you curate.** The failure mode we've watched happen to a sibling project: the moment you show people a curated-looking store, they curate it; the moment they curate it, hallucination stops being "a summary" and becomes "a bug." Intent doesn't survive org gravity — only structure does.

### The anti-rot rules (structural, not aspirational)

1. **Nothing durable lives here without a hard-source backing (a real GitHub issue/PR) or a TTL.** The substrate physically can't accumulate into a backlog to groom.
2. **Read-through, never edit-through.** You can't mutate authoritative state from the board — no second master to clobber on sync.
3. **Provenance is visible.** If a row is soft, it looks soft. You can't accidentally present off a pointer.
4. **Self-populating.** It fills from reality (events, PRs, CI, agents); the human steers exceptions, never fills an empty box.

### Prior art: this is "Delta, for coordination"

Zed's Delta / DeltaDB does exactly this for **code**: Git stays authoritative, DeltaDB captures what happens *between commits* (edits + conversation) anchored so it can't rot, and you steer by asking the agent instead of reconstructing intent from a diff. We're one altitude up: capture what happens **between issues**, across teams and agents, over a shared fabric instead of one worktree. Same discipline, coordination granularity. (https://zed.dev/blog/introducing-delta)

### The three organs (keep them separate)

- **GitHub** — authoritative work orders. Never usurped.
- **TOME / the wiki** — curated structural + cultural knowledge (how we deploy, OpenFGA conventions, gotchas). A different organ; the board reaches into it but does not become it.
- **Mycelium** — the connective tissue: the derived cognition that accretes *while work happens* (decisions, deploy plans, episodes) + the fabric where agents coordinate. This is what we own and what nobody else has.

### The drift we're trying to close

What's built is closer to a human-driven triage board (Trello-ish). The vision wants:
- **"Needs you" = what agents escalated to you**, not human to-dos.
- **GitHub-backed rows are first-class** (down-link: an actively-worked issue projects its live execution state); bespoke items are visibly transient.
- **Agent coordination is visible** on the board (handoffs, who's addressing whom), not invisible plumbing.
- **Anti-rot enforced**, not just intended (backing-or-TTL as a real property).

The commits after this comment are a first pass at pulling the board toward that axis.

PR #529, opened 2026-08-14
refactor: drop the inert per-agent budget_usd_per_month cap

### LINKED ISSUE #525: Drop the per-agent budget system (`budget_usd_per_month`) — it was cold-spawn enforcement, now inert
## Why

The budget system was **cold-spawn enforcement**, and cold-spawn is being removed (daemon-removal PR). With the daemon gone, nothing enforces it — it's now a cosmetic cap on the manifest that no code path reads to gate anything.

### How it actually worked (all in the now-deleted daemon)
- `budget_usd_per_month` on the agent manifest.
- The daemon's `spawn_claude` parsed `total_cost_usd` out of `claude -p`'s JSON output and accumulated it in `DaemonState.budget_used_usd[(handle, month)]`.
- `gate_budget()` checked `used < budget_usd_per_month` **before each cold-spawn** and refused the dispatch over-cap.

That accumulation + gate lived entirely in `daemon/dispatch.py` + `daemon/state.py` + `claude_code/spawn.py` — **all deleted.** In the resident model the runtime is the user's own session (Claude Code / Cursor), so mycelium never sees per-call cost and can't meter spend. Enforcement belongs at the provider account, not here.

### Why drop it rather than keep the field
- It reports a **cap, not spend** — `user show` sums manifest caps and calls it the user's cost, which was always an approximation and is now not even a ceiling on anything.
- Keeping an unenforced money field invites false confidence ("this agent is capped at \$5/mo" — it isn't).
- It's honest-architecture debt: the whole reason it existed (meter cold-spawn token cost) is gone.

## Scope (what to remove)

**CLI**
- `protocol.py`: `AgentManifest.budget_usd_per_month` field
- `commands/agent.py`: `--budget` flag, `budget=` plumbing in `build_manifest` calls, the `agent show` / `agent ls` budget lines, the owner budget roll-up (`_owned_agents_*`)
- `commands/user.py`: summed-budget display
- `integrations/{base,claude_code,cursor,engine}/dispatch.py`: `budget` param on `build_manifest`
- `docs/principals.md`: the "summed budget" section

**Backend**
- `schemas.py`, `routes/agents.py`, `routes/users.py`, `services/principals.py`: `budget_usd_per_month` field + any summed-budget response

**Frontend**
- `lib/api.ts`: `AgentSummary.budget_usd_per_month`

## Open question
Do we want *any* cost surface later? If so it should be **observed spend** (from whatever runtime actually reports it), not a manifest cap — a separate, honest feature. This issue is just: remove the inert cap.

## Depends on
The daemon-removal PR (which deletes `gate_budget` + the accumulation). Best done right after it lands so we're not editing the same manifest surface twice.

juliarvalenti:
## Summary

Removes `budget_usd_per_month`, the per-agent monthly-spend cap that used to be enforced by the now-deleted daemon (`gate_budget()` in `daemon/dispatch.py`, checked before each `claude -p` cold-spawn). With cold-spawn gone, agents are resident sessions mycelium never meters, so the field had become a cosmetic manifest cap nothing reads — reporting a cap, not spend, and inviting false confidence ("this agent is capped at $5/mo" — it isn't).

## Changes

**CLI**
- `protocol.py`: drop `AgentManifest.budget_usd_per_month`
- `commands/agent.py`: drop the `--budget` flag, `budget=` plumbing through `build_manifest`, the Budget column in `agent ls`, the budget line in `agent show`, and the summed-budget return from `load_owned_agents`
- `commands/user.py`: drop the summed-budget display in `user show` / `whoami` (text + JSON)
- `commands/engine.py`, `integrations/{base,claude_code,cursor,engine}/dispatch.py`: drop the `budget` param on `build_manifest`
- `docs/principals.md`: drop the "Cost roll-up" section

**Backend**
- `schemas.py`, `routes/agents.py`, `routes/users.py`, `services/principals.py`: drop `budget_usd_per_month` field + the summed-budget roll-up on users/teams

**Frontend**
- `lib/api.ts`: drop `budget_usd_per_month` from `AgentSummary`, `OwnedAgent`, `User`, `Team`

**Docs**
- Regenerated `docs/concepts.html` + `docs/llms-full.txt` from the updated `principals.md` source (`docs/generate_docs.py`)

## Why stacked on `refactor/remove-daemon-cold-spawn` (#526)

Issue #525 explicitly depends on the daemon-removal PR landing first ("best done right after it lands so we're not editing the same manifest surface twice") — that PR deletes `gate_budget()` and the accumulation in `daemon/dispatch.py`/`daemon/state.py`. Since #526 was still open when this was picked up, this branch is based on `refactor/remove-daemon-cold-spawn` rather than `main` so the two don't collide. Retarget to `main` once #526 merges.

## Testing

- [x] Unit tests pass (`uv run pytest tests/ -x -q`) — CLI: 272 passed; backend: 407 passed, 6 skipped
- [x] Linting passes (`uv run ruff check .`) — both CLI and backend
- [x] Format check passes (`uv run ruff format --check .`) — both CLI and backend
- [x] `uv run ty check .` clean (CLI + backend; only a pre-existing unrelated `negmas` import note in the sandbox)
- [x] Frontend `tsc --noEmit` clean
- [x] CLI↔backend SLIM+L9 wire-contract test unaffected (`test_slim_l9_wire.py`, both sides)

## Related Issues

Closes #525

---
_Generated by [Claude Code](https://claude.ai/code/session_016Y5WvQCarzhi5DiNR84xWt)_


juliarvalenti:
Rebased onto main now that #526 has landed (used `--onto` to drop the daemon-removal commits this branch carried, which are already on main squashed) and retargeted the base main→. One reconciliation: `routes/engines.py` (native engine invite, merged in #523 after this branch was cut) still wrote `budget_usd_per_month` and passed it to `AgentRead` — dropped both so the route keeps working with the field gone. Gates green: backend ruff/ty + principals/engines tests, CLI ruff/format/ty + 296 tests, frontend tsc.

PR #532, opened 2026-08-14
Configurable audio ping for room activity when the tab isn't focused

### LINKED ISSUE #514: Configurable audio ping for room activity when the tab isn't focused
When you're not looking at the Mycelium tab, there's nothing to pull you back — you have to keep the window visible to notice a reply, a mention, or a consensus. Add an **audio ping** for room activity while the tab is unfocused/hidden, with user controls.

No audio/notification code exists in the frontend today.

## Behavior
- Play a short sound when a relevant event arrives **and** the tab isn't visible (`document.hidden` / `visibilitychange`, Page Visibility API). Never ping while the user is actively viewing.
- Source the events off the existing SSE stream (`event-stream.tsx`) — no new backend needed.

## Configurable
- **On/off** toggle (persisted locally, like the acting-as identity).
- **Scope:** what pings — everything, or only "needs me" (mentions of my acting-as handle / my agents, consensus, escalations). Default to the narrow scope so it isn't noisy.
- **Sound + volume** (a couple of built-in tones; mute).
- Debounce / rate-limit so a burst of messages is one ping, not twenty.

## Notes
- Browsers require a user gesture before audio can play — gate first playback behind the user enabling it (the toggle click counts).
- Optional stretch: also fire a Web `Notification` (OS-level) when hidden, gated on permission — but the audio ping is the ask here.
- Ties naturally to the principal work: "needs me" scoping keys off the acting-as handle + owned agents.

juliarvalenti:
## Summary

Adds an audio ping for room activity while the browser tab is hidden/unfocused, so a reply, mention, or consensus doesn't require keeping the tab visible to notice. Sourced entirely off the existing SSE stream in `event-stream.tsx` — no backend changes.

## Changes

- `src/lib/audio-ping.ts` — synthesizes three short tones (chime/ping/tone) via Web Audio oscillators, no audio assets to ship. `primeAudio()` creates/resumes the shared `AudioContext` from a user gesture, satisfying the browser's autoplay-gesture requirement.
- `src/components/notification-settings.tsx` — locally persisted preferences (on/off, scope, sound, volume) under `localStorage["mycelium.notify"]`, mirroring the existing acting-as identity pattern (`current-user.tsx`). Wired into the root layout as a context provider.
- `src/components/notification-bell.tsx` — a bell popover in the room header (next to the acting-as picker) to toggle pinging, pick scope, sound, and volume, with a click-to-preview on each sound option.
- `event-stream.tsx` — the SSE `onmessage` handler now checks `document.hidden` and the current scope before playing a tone:
  - **Scope "Needs me"** (default, narrow): consensus always qualifies; chat only qualifies when it's directed at me, directed at an agent I own, or @-mentions either — never for my own outgoing messages.
  - **Scope "Everything"**: any activity visible in the channel view.
  - Rate-limited to one ping per 4s so a burst of messages/negotiation ticks doesn't turn into a machine-gun.
- Existing `event-stream.*.test.tsx` files now render through a small `renderWithProviders` test helper (`src/test/render-with-providers.tsx`) since `EventStream` now depends on the acting-as/notification-settings contexts.
- New `event-stream.audio-ping.test.tsx` covers: no ping while visible, no ping when disabled, scope "all" pings on a broadcast, scope "needs-me" ignores unrelated chat and pings on an @-mention, scope "needs-me" always pings on consensus, and a burst of activity debounces to a single ping.

## Testing

- [x] Unit tests pass (`pnpm test` — 19 passed, including the 7 new audio-ping tests)
- [x] Type check passes (`pnpm lint` / `tsc --noEmit`)
- [x] Production build passes (`pnpm build`)
- [x] Manual check in `MYCELIUM_UI_MOCK=1` dev mode via headless Chromium: bell renders in the room header, popover opens with the enable toggle/scope/sound/volume controls, toggling on persists `{"enabled":true,"scope":"needs-me","sound":"ping","volume":0.8}` to `localStorage` with no console errors from the new code

## Related Issues

Closes #514

---
_Generated by [Claude Code](https://claude.ai/code/session_014WJsTa5ev5ZaPqMbYf3iVA)_

### DIFF
--- a/mycelium-frontend/AGENTS.md
+++ b/mycelium-frontend/AGENTS.md
@@ -1,0 +1,9 @@
+<!-- BEGIN:nextjs-agent-rules -->
+
+# This is NOT the Next.js you know
+
+This version has breaking changes — APIs, conventions, and file structure may all differ from your training data. Read the relevant guide in `node_modules/next/dist/docs/` (resolved from this file's directory; in monorepos the `next` package may not be visible from the repo root) before writing any code. Heed deprecation notices.
+
+This block is written and re-added by `next dev` — verify at `node_modules/next/dist/server/lib/generate-agent-files.js`. Removing it from a diff only re-creates the uncommitted change; committing it with your work keeps the tree clean.
+
+<!-- END:nextjs-agent-rules -->
--- a/mycelium-frontend/CLAUDE.md
+++ b/mycelium-frontend/CLAUDE.md
@@ -1,0 +1,1 @@
+@AGENTS.md
--- a/mycelium-frontend/src/app/layout.tsx
+++ b/mycelium-frontend/src/app/layout.tsx
@@ -5,6 +5,7 @@ import type { Metadata } from "next";
 import "./globals.css";
 import { ThemeProvider } from "@/components/theme-provider";
 import { CurrentUserProvider } from "@/components/current-user";
+import { NotificationSettingsProvider } from "@/components/notification-settings";
 
 export const metadata: Metadata = {
   title: "mycelium",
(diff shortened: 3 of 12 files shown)


juliarvalenti:
Superseded by #537 (the full notifications system per #536). The per-room ping design is replaced by a cross-room aggregate stream + inbox, so closing without merging.

PR #535, opened 2026-08-15
fix(frontend): switch main chat window to the shared ScrollArea

### LINKED ISSUE #521: Main chat window uses raw overflow-y; switch to the shared ScrollArea
The main chat/event window (`mycelium-frontend/src/components/event-stream.tsx`) scrolls with raw `overflow-y-auto` (lines ~462 and ~466), while the rest of the app uses the shared `ScrollArea` primitive (`components/ui/scroll-area.tsx`, already used in `rooms-sidebar.tsx`). Inconsistent scrollbar styling/behavior, and the chat is the most-looked-at surface.

## Ask
Wrap the event/chat list in `ScrollArea` instead of a bare `overflow-y-auto` div, for consistent styled scrollbars across the app.

## Watch out for (the one real wrinkle)
Auto-scroll-to-bottom currently does `scrollRef.current?.scrollTo(...)` on the overflow div (event-stream.tsx ~381). With `ScrollArea` (base-ui) the scrollable element is a nested **viewport**, not the outer node — so the ref needs to point at the ScrollArea viewport, or the "stick to bottom on new message" behavior breaks. Verify:
- new message auto-scrolls to bottom,
- and it doesn't yank you down when you've scrolled up to read history (if we have that behavior; if not, fine).

juliarvalenti:
## Summary

The main chat/event window (`event-stream.tsx`) scrolled with a raw `overflow-y-auto` div while the rest of the app (e.g. `rooms-sidebar.tsx`) uses the shared `ScrollArea` primitive (`components/ui/scroll-area.tsx`). This swaps both `overflow-y-auto` usages in `event-stream.tsx` — the channel/chat list and the SLIM tab — to `ScrollArea`, for consistent styled scrollbars across the app.

## Changes

- `components/ui/scroll-area.tsx`: add an optional `viewportRef` prop to `ScrollArea` so callers can get a ref to the actual scrollable **viewport** node (base-ui nests the scrollable element inside the root, it's not the outer wrapper).
- `event-stream.tsx`: wrap the channel/chat list and the SLIM view in `ScrollArea`; the existing auto-scroll-to-bottom `scrollRef` now points at `viewportRef` so `scrollRef.current?.scrollTo(...)` keeps targeting the real scrollable element.
- `vitest.setup.ts`: stub `Element.prototype.getAnimations` (jsdom doesn't implement it) — base-ui's `ScrollArea` viewport calls it internally, and it was throwing uncaught exceptions in component tests that render `ScrollArea`.
- `event-stream.render.test.tsx`: add a regression test asserting that a new SSE message calls `scrollTo` on the element with `data-slot="scroll-area-viewport"`, not the outer wrapper — this was the one real wrinkle called out in the issue.

## Testing

- [x] `npx tsc --noEmit` passes
- [x] `npx vitest run` passes (13/13, including the new auto-scroll regression test)
- [x] `npm run build` (Next.js production build) succeeds

## Related Issues

Closes #521

---
_Generated by [Claude Code](https://claude.ai/code/session_01GEyj1kwCSqyy1SHR7T42UP)_

### DIFF
--- a/mycelium-frontend/src/components/event-stream.render.test.tsx
+++ b/mycelium-frontend/src/components/event-stream.render.test.tsx
@@ -57,6 +57,24 @@ describe("<EventStream /> live message rendering", () => {
     expect(await screen.findByText("hello over the live channel")).toBeInTheDocument();
   });
 
+  it("auto-scrolls the ScrollArea viewport, not the outer wrapper, on new messages", async () => {
+    const scrollTo = vi.spyOn(Element.prototype, "scrollTo");
+    render(<EventStream roomName="sprint" />);
+    await act(async () => {});
+    const es = FakeEventSource.latest();
+
+    await act(async () => {
+      es.open();
+      es.emit(l9Exchange("stick to bottom"));
+    });
+    await screen.findByText("stick to bottom");
+
+    expect(scrollTo).toHaveBeenCalled();
+    const scrolledEl = scrollTo.mock.contexts[scrollTo.mock.contexts.length - 1] as Element;
+    expect(scrolledEl.getAttribute("data-slot")).toBe("scroll-area-viewport");
+    scrollTo.mockRestore();
+  });
+
   it("renders an l9_commit streamed over SSE as a consensus notice, not the unhandled fallback", async () => {
     const warn = vi.spyOn(console, "warn").mockImplementation(() => {});
     render(<EventStream roomName="sprint" />);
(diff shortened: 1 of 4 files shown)


juliarvalenti:
Rebased onto main (resolved an additive test conflict with #537's l9_commit/l9_knowledge render tests — kept all three). Also fixed a scroll regression: the swap had dropped `min-h-0` from the channel container, and base-ui's `ScrollArea` (size-full viewport) needs a height-bounded root, so the chat wouldn't scroll at all. Restored `min-h-0` to match the SLIM ScrollArea. Verified scrolling + auto-stick-to-bottom in the app.

Issue #549, opened 2026-08-15
Apply inbound knowledge envelopes on receive (close cross-store memory convergence)


juliarvalenti:
## Problem

As of #544/#548, the backend **emits** `knowledge` envelopes on memory writes (`extraction`) and on consensus plan-sync (`distillation`), and `ingest_local`s them so the *sender's* own transcript/UI bus sees them. But **nothing on the receiving backend applies an inbound `knowledge` envelope** to its local memory store. So a `knowledge` broadcast reaching another host updates that host's transcript/UI at best — it does not write the carried markdown into that host's `.mycelium/` store or reindex it.

This gap is symmetric: it affects both `distillation` (compiled plans) and `extraction` (direct writes) equally. The historical consumer — the CLI daemon's `apply_knowledge_message` (referenced in `contracts/slim-l9-wire.json`'s knowledge comment) — was removed along with the cold-spawn daemon, so the apply half is currently unwired.

The producer + applier primitives already exist:
- **Producer:** `memory_sync.build_knowledge_envelope` (done, #544/#548).
- **Parser:** `memory_sync.knowledge_write_from_envelope` — subkind-agnostic, already parses both distillation and extraction.
- **Applier:** `memory_sync.apply_knowledge` / `apply_knowledge_to_dir` — writes the markdown + reindexes. Present but not invoked on inbound channel traffic.

## Proposal

Wire the receive path so a backend that receives a `knowledge` envelope on a room channel calls `apply_knowledge` (subkind-agnostic) to converge its local store, then surfaces it on its bus.

Design points:
- **Loop guard is critical.** Applying an inbound knowledge write must **not** re-trigger the outbound `_broadcast_memory_write` producer (#548) — otherwise two hosts ping-pong a write forever. Apply-on-receive must be a distinct path from the API write path, and should not itself emit.
- **Idempotency / versioning.** Use `base_version`/`version` (already carried in the payload) to make re-delivery and out-of-order arrival safe; last-writer / version-aware reconcile (ties into the Drive-style sync design, #543).
- **Scope:** both `distillation` and `extraction` apply the same way — the applier already ignores subkind.
- Consider whether this belongs on the moderator backend, every connected member, or both.

## Related
- #544 / #548 (the producer half — now merged)
- #543 (Drive-style transparent memory sync — the conflict/version model lives here)
- `contracts/slim-l9-wire.json` (the knowledge wire shape both sides must agree on)

Issue #544, opened 2026-08-15
Emit an L9 knowledge packet when shared room memories are updated


juliarvalenti:
## Problem

When an agent updates a memory in a shared room, nothing tells the other members. Today the update is silent, so a human or agent has to manually announce "I updated the memories" in the room — which is exactly the kind of coordination overhead L9 exists to remove.

## Proposal

On `memory set` (and direct-file-edit reconciliation) in a **shared** room, the backend emits an additive **L9 `knowledge`** envelope on the room's SLIM channel describing the change (key, version, author, room, maybe a short summary).

Consequences:
- Resident agents receive the update pushed to them — this is the natural producer for the existing `mycelium memory subscribe` consumer.
- The knowledge packet **is** the announcement: machine-readable, additive, no manual "I updated stuff" message required.
- Feeds the UI's need for server-pushed refresh events on top-level state changes.

## Design notes

- L9 envelopes are additive and never required of agents (see `CLAUDE.md` / `app/services/l9.py`). `knowledge` is already a defined kind — this is a new producer, not a new subkind.
- Fire-and-forget / fail-soft: a memory write must still succeed if the envelope can't be sent.
- Scope to shared rooms; private/local-only writes need not broadcast.
- Payload should carry the causal context (episode/topic URNs) consistent with other L9 envelopes.

## Related
- #543 (Drive-style transparent sync — this is the push-notification half of keeping everyone current)
- #542 (the read-side split-brain bug)

juliarvalenti:
This paket should also emit to the chat as well as the l9 message set

juliarvalenti:
Implemented by #548 (merged): `memory set` now broadcasts a `knowledge:extraction` L9 envelope on the room channel — the producer this issue asked for. It surfaces in the CLI/UI `l9_knowledge` feed (raise-up whitelist, #546). Note: under the thin-client spoke model (#557), the cross-machine *apply* side (#550) is dormant, but the emit + in-room feed this issue scoped is done. Closing as implemented.

Issue #543, opened 2026-08-15
Drive-style transparent local/remote memory sync (retire the manual sync ritual)


juliarvalenti:
## Problem

Memory today has a split-brain: writes/search/subscribe go to the **backend**, while `ls`/`get` read **local files**, and a manual `mycelium memory sync` is the only thing that reconciles them. The `sync` ritual is unintuitive — you don't know you need it until a read silently lies (#542). The `.mycelium/`-is-truth model only holds when the CLI and backend share a filesystem; with a remote backend they don't.

## Target design: local mirror, backend is truth (like Google Drive)

**The room's memory store (backend, SLIM-synced) is the source of truth and is always up to date. Local `.mycelium/` is a transparent, editable mirror** — you read it, you can edit files directly, and changes reconcile automatically. You should never *have* to think about sync.

1. **Reads become read-through.** `ls`/`get` reflect authoritative state and opportunistically refresh the local mirror. Offline → fall back to the mirror, clearly labeled stale (never silently empty — see #542).
2. **`sync` demotes to an escape hatch.** Keep it for "force full materialize for offline work / git sharing" (cf. Drive's *Available offline*), but remove it from the happy path.
3. **Writes reconcile both ways.** `memory set` writes backend **and** refreshes the mirror. Direct local file edits are detected and pushed up (watcher or push-on-detect) — this is the "edit via your filesystem" half of the Drive model and the harder part.
4. **Conflict handling.** Define last-writer-wins vs. version-aware merge when both sides changed (memories already carry a `version`).

## Why

Removes the footgun in #542 by construction, and makes "state is files" true again even when the backend is remote. Sharing-via-git still works because the mirror is real files.

## Related
- #542 (the silent split-brain bug + minimal stopgap)
- Emit L9 `knowledge` on shared-memory update (the push-notification half of keeping everyone current)

juliarvalenti:
Superseded by #557. Drive-style two-way local/remote sync is unnecessary once spokes keep no local file replica and read through the hub — there's a single store (the hub's), so there's nothing to reconcile. The conflict/version semantics discussion is preserved here for reference if ever needed.

PR #553, opened 2026-08-15
Add `mycelium join`, the thin-spoke member client

### LINKED ISSUE #551: Build the `mycelium join <hub>` thin-spoke member client (make the knowledge emit→apply loop actually cross machines)
## Goal

Build `mycelium join <hub>`: a thin-spoke **member client** that joins an existing room's channel (moderated by one always-on backend), receives the room's traffic, and **applies inbound `knowledge` envelopes to its own local store**. This is the missing piece that makes the already-merged knowledge emit→apply loop actually cross machines.

## Background (why this is needed)

The emit and apply halves already exist:
- **Emit** (#544/#548): `memory set` on the hub broadcasts a `knowledge:extraction` L9 envelope on the room channel.
- **Apply** (#549/#550): a persister applies an inbound `knowledge` envelope to its local store.

But there is currently **no process that receives an envelope it didn't emit itself**, so the loop is dormant. Verified live: two independent backends on the same SLIM node + same room + same PSK do **not** share a group (each is its own moderator with its own MLS session), so nothing crosses — not even a plain chat broadcast. One-moderator-per-channel is the intended topology; the fix is a **member**, not a second moderator.

## What already exists to reuse

- **Moderator + invite flow:** `fastapi-backend/app/services/room_channels.py` — the hub creates the group session and invites members (`manager.invite(room, agent)`). Joining as a member is the supported path.
- **The applier:** `fastapi-backend/app/services/persister.py` — `_ingest` → `_schedule_knowledge_apply` → `_apply_knowledge`, calling `memory_sync.apply_knowledge` (subkind-agnostic, version-idempotent, conflict-aware). This is what a spoke needs to run on inbound traffic.
- **Cross-machine transport:** `mycelium hub host` / `mycelium connect` already point a client at a shared SLIM node.

## What to build

1. **`mycelium join <hub>` command** (CLI): enroll this machine as a **member** of a room on a remote hub — resolve the hub + room, obtain membership (invite/accept), and start a long-lived receiver.
2. **Spoke receiver runtime:** a lightweight process that holds channel membership and runs the persister's apply path on inbound `knowledge` envelopes, writing into the spoke's **local** `.mycelium/` store. Reuse `_apply_knowledge` / `memory_sync.apply_knowledge` rather than reimplementing.
3. **Identity gating:** membership must be gated on per-agent identity within the trust domain (JWT/SPIRE — see #446). PSK-only is dev-tier; don't ship spoke join without an identity story (coordinate with the identity work).
4. **(Follow-up, not this ticket)** bidirectional: a spoke's *own* writes reaching the hub. Land the receive direction first.

## Acceptance criteria

- With a hub backend running room R and a spoke that ran `mycelium join <hub>` for R:
  - a `memory set` on the hub results in the memory appearing in the **spoke's local store**, and
  - the spoke's `knowledge_applied` health counter increments (`knowledge_conflicts` stays 0 for a clean apply).
- A re-delivery / same-version envelope is a silent no-op (idempotent), and a stale-base write is recorded as a conflict, not a clobber (the #550 semantics, now over real transport).
- The spoke is a **member**, not a moderator — no second group session is created for R.

## How to verify

Same shape as the #550 live test, but the second store joins as a **member** instead of standing up its own moderator: run one hub backend, `mycelium join` from a second store, `memory set` on the hub, assert convergence into the spoke + its health counter. (The old two-moderator setup can't work — that's the bug this closes.)

## Non-goal (do not build; do not reopen)

**Moderator federation / multi-master is explicitly out of scope.** The reason is *wrong topology*, not difficulty: single-writer-through-the-hub is a property we want. Two co-equal moderators would need MLS group-merge, dedup, multi-master conflict resolution, and leader election — the peer-mesh/public-distributed shape ruled out by **one-instance-per-trust-domain**. Do not reopen this as a scaling optimization.

## Related
- #544 / #548 — emit knowledge on memory write
- #549 / #550 — apply inbound knowledge on receive (the path the spoke runs)
- #543 — Drive-style transparent memory sync (conflict/version model)
- #446 — herdr integration spike (thin-spoke identity / wake path)

juliarvalenti:
## Summary

The knowledge emit→apply loop had both halves but no middle. `memory set` on a hub
broadcasts a `knowledge:extraction` envelope (#544/#548) and an applier for inbound
`knowledge` exists (#549/#550) — but nothing ever *received* an envelope it hadn't
emitted itself, so the loop never crossed machines.

Standing up a second backend doesn't close it: each becomes its own moderator with its
own MLS session, so nothing crosses — not even a plain broadcast. One-moderator-per-room
is the intended topology, so the fix is a **member**.

`mycelium join <room>` is that member. It runs the same handshake the `l9 send` / `slim
send` plumbing already used — connect, ask the hub's moderator to invite us, wait to be
admitted — but holds the session open instead of publishing once, and applies every
inbound `knowledge` write into its own local store. A `memory set` on the hub now lands
on the spoke.

Verified against a live SLIM node: a moderator broadcasts, an independent member client
receives and applies it.

## Changes

- **`mycelium/spoke.py`** (new) — the receiver runtime: decode inbound frames, filter to
  `knowledge`, apply the carried write to this machine's `.mycelium/`, tracking
  `applied` / `conflicts` / `ignored` counters that mirror the backend persister's.
- **`mycelium/commands/join.py`** (new) — `mycelium join [--room R] [--handle H]`, a
  resident foreground member; per-write output, counters on exit, Ctrl-C to stop.
  Registered top-level beside `await` / `respond` / `connect`.
- **`slim/member.py`** — the join handshake moves out of `publish_once` into a
  `joined_member` context manager, so the publish-once and long-lived cases share one
  path. `publish_once` keeps its exact behavior.
- The spoke borrows only the *apply* half of the persister. The transcript, delivery
  cursors, in-process bus and episode hooks are moderator-authoritative and have no
  meaning off the hub; the version check already makes a replay a no-op, so no
  message-level dedup is needed.
- **Docs** — a "Receiving memory on a spoke" section in `docs/cross-machine.md`
  (the page previously claimed the plan "syncs to every machine", which wasn't yet
  true); regenerated `reference.html`; `generate_docs.py` imports the new command so
  its `@doc_ref` registers; CLAUDE.md gains the one-moderator/spokes-are-members note.
- **Contract** — `contracts/slim-l9-wire.json`'s comments pointed at
  `daemon/connector.py`, removed long ago with the daemon; repointed at the module that
  now parses and applies that envelope. No frozen values changed.

### Behavior and scope

Semantics carry over unchanged onto real transport: last-write-wins by version, a
re-delivered version is a silent no-op, and a stale-base write is recorded as a conflict
rather than clobbering newer local content.

Membership is gated by the room's shared-secret PSK, exactly as every other member is — a
spoke is no more authenticated than a connector, and per-agent identity (JWT/SPIRE, #446)
remains a prerequisite for anything hosted or multi-user. This PR doesn't claim to solve
that.

Two deliberate limits: direction is **hub → spoke only** (a spoke's own writes reaching
the hub is the follow-up the issue scopes out), and received memories are on disk but not
searchable until `mycelium sync`, since the CLI has no local embedding index. Moderator
federation / multi-master stays out of scope.

## Testing

- `mycelium-cli/tests/test_spoke.py` (new, node-free) — 15 tests driving the **frozen
  contract envelope** the backend actually produces, so these assert against real hub
  traffic rather than a shape invented in the test: decode, apply, idempotent
  re-delivery, stale-base conflict, newer-version supersede, malformed/non-knowledge
  frames ignored, and the join path asserting membership came from asking the moderator
  to invite (never creating a group).
- `mycelium-cli/tests/test_spoke_roundtrip.py` (new, live-node) — one client moderates,
  another **joins as a member**, and the spoke applies a broadcast it did not emit; plus
  idempotent replay and stale-base conflict over the wire. Ran green against a real SLIM
  node. This is the first CLI path that needs a node, so CI's `integration-slim` job now
  runs it (the job carried a comment saying the CLI had no such slices — no longer true).

- [x] Unit tests pass — 313 passed (CLI), 434 passed / 2 skipped (backend)
- [x] Type check passes (`uv run ty check .`)

## Related Issues

Closes #551

Related: #544 / #548 (emit), #549 / #550 (apply), #543 (conflict/version model),
#446 (spoke identity)

https://claude.ai/code/session_01As6dWbqUhAArZ83G5aSdjq

---
_Generated by [Claude Code](https://claude.ai/code/session_01As6dWbqUhAArZ83G5aSdjq)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -233,8 +233,19 @@ jobs:
           tests/test_slim_roundtrip.py
           tests/test_l9_over_slim_roundtrip.py
 
-      # The CLI has no live-node slices: `await`/`respond` are pure HTTP (the
-      # SLIM member connector that needed a node was removed with the daemon).
+      - name: CLI deps
+        working-directory: mycelium-cli
+        run: uv sync --group dev --extra engine
+
+      # `await`/`respond` are pure HTTP, but the thin spoke (`mycelium join`) is a
+      # real SLIM member — the only CLI path that needs a node. Its node-free
+      # semantics are covered in the unit job; this proves it receives traffic it
+      # did not emit, which is the whole point of the spoke.
+      - name: CLI live-node slices
+        working-directory: mycelium-cli
+        env:
+          MYCELIUM_SLIM_ENDPOINT: http://127.0.0.1:46357
+        run: uv run pytest -q tests/test_spoke_roundtrip.py
 
       - name: SLIM node logs (on failure)
         if: failure()
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -71,7 +71,8 @@ backend is each room's **moderator**; agents (and the human, by proxy) are membe
 `.mycelium/rooms/{room}/{key}.md`. Search is a **local embedding index** (fastembed
 ONNX, `BAAI/bge-small-en-v1.5`, 384-dim, no external service) persisted as JSONL
 per room. `memory set` writes the markdown and updates the index; direct file
-writes work too; run `mycelium reindex` to resync. Sharing is git.
+writes work too; run `mycelium reindex` to resync. Git can version or back up the
+files, but it is **not** the sharing path — see **Sharing is the live channel** below.
 
 **L9 rides SLIM.** Coordination messages carry additive IOC **L9** JSON envelopes:
 `exchange` (ticks/replies), `commit:converged|resolved|rejected`, `knowledge`.
@@ -156,8 +157,11 @@ is no litellm dependency.
   `members()` is the union of live SLIM members and lease holders. This is why
   turn-based agents never miss a tick. The **durable inbox** (`persister.py`)
   re-serves missed point-to-point messages on rejoin (SLIM has no offline replay).
-- **Git for sharing.** Rooms can be shared via git push/pull; cross-machine is the
-  same channel over a shared SLIM node (`mycelium hub host` / `mycelium connect`).
+- **Sharing is the live channel.** Cross-machine sharing is the same SLIM channel over a
+  shared node (`mycelium hub host` / `mycelium connect` / `mycelium join`), plus
+  `mycelium room clone --from <api-url>` for a point-in-time HTTP snapshot. Git can back
+  up or version the `.mycelium/` files but is **not** a sharing mechanism — no room flow
+  pushes or pulls over git.
 - **One moderator per room; spokes are members.** Cross-machine memory sync runs
   hub → spoke over the room's channel: the hub broadcasts a `knowledge` envelope on
   every `memory set`, and `mycelium join` (`mycelium/spoke.py`) holds membership on a
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -158,6 +158,13 @@ is no litellm dependency.
   re-serves missed point-to-point messages on rejoin (SLIM has no offline replay).
 - **Git for sharing.** Rooms can be shared via git push/pull; cross-machine is the
   same channel over a shared SLIM node (`mycelium hub host` / `mycelium connect`).
+- **One moderator per room; spokes are members.** Cross-machine memory sync runs
+  hub → spoke over the room's channel: the hub broadcasts a `knowledge` envelope on
+  every `memory set`, and `mycelium join` (`mycelium/spoke.py`) holds membership on a
+  second machine and applies each one to its own local store. A second backend on the
+  same room is *not* the path — each would moderate its own MLS group and nothing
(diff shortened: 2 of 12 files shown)


juliarvalenti:
Closing in favor of the CLI thin-client spoke model (#557). This PR builds the spoke's local file replica + apply-on-receive; we've decided a spoke keeps no local files and accesses memory through the hub instead, so the local-store receiver is superseded. Memory itself is unchanged — this is about client access, not removing memory. The CLAUDE.md 'sharing is the live channel' doc fix from this branch is preserved in PR #556. The 'resident member to be woken' idea, if revived, belongs to the participation/wake track (#446), not memory sync.

Issue #575, opened 2026-08-16
Epic: keyboard-first & command-driven navigation


juliarvalenti:
## Goal

**Make the entire app operable by keyboard alone — vim-inspired.** Every primary surface reachable without a mouse; a modal, discoverable, command-driven UX.

## Tracks

- [ ] #572 — Keyboard-first navigation: app-wide keybinds (vim-inspired)
- [ ] #573 — Global command palette (⌘K)
- [ ] #574 — Command-style search

## Principles (shared across all three)
- **Keyboard-complete:** room switching, room panes, rail tabs, and chat focus are all reachable without a pointer.
- **One central keymap registry** — declared once, reused by the keybinds, the palette, and the `?` cheatsheet. Every binding is discoverable and conflict-checked.
- **Modal / vim feel:** command mode vs typing. `Esc` returns to command mode; **keybinds never fire inside text inputs** (chat, forms).
- **Scales past 10 rooms:** switching uses **Vimium-style hint labels** (home-row chars → 2-char hints), not digits-only.
- **Discoverable:** a `?` cheatsheet of the bindings active in the current context.
- **Accessible:** focus management, ARIA roles, `prefers-reduced-motion`.

## Grounding (frontend)
- `app-shell.tsx` — overall shell / keymap host
- `rooms-sidebar.tsx` — left-nav room list (hint switcher)
- `room-inspector.tsx` — right rail tabs (Members / Episodes / Memory)
- `src/app/room/[name]/page.tsx` — room panes (channel / negotiate / l9 / plan / slim)
- `room-chat-box.tsx` — chat input (focus target + the "don't fire while typing" guard)
- `command-center.tsx` — existing **home dashboard** (not a palette; note the naming collision for #573)

Issue #573, opened 2026-08-16
Global command palette (⌘K)


juliarvalenti:
## Framing

A global **⌘K / Ctrl-K command palette** to run any navigation or action from the keyboard — connective tissue for the keyboard-first push (see the epic).

## Scope
- Fuzzy-searchable **command registry**: switch room, open pane (channel / negotiate / l9 / plan / slim), switch rail tab (members / episodes / memory), focus chat, create room, invite engine, toggle theme, acting-as switch, go to metrics, etc.
- Sourced from the **same central keymap registry** as the keybinds issue, so every binding is also a palette command **and the palette displays its shortcut**.
- Keyboard-navigable results; grouped/scoped; recent/frequent surfaced first.
- **Extensible + context-aware:** components register commands; room commands only appear inside a room.

## Notes
- Distinct from the existing `command-center.tsx` (that's the **home dashboard landing**, not a palette). Consider renaming it to avoid the naming collision.
- Consider whether **command-style search** (separate issue) shares this surface as a mode (e.g. `>` prefix = actions, plain text = search).

## Open questions
- Build on the existing Base UI primitives (we already wrap Dialog/Popover/Tooltip) vs pull in a `cmdk`-style lib.
- One surface with modes, or palette + search as siblings.

juliarvalenti:
Part of the keyboard-first epic #575.

PR #576, opened 2026-08-16
feat(frontend): keyboard-first navigation (#572)

### LINKED ISSUE #572: Keyboard-first navigation: app-wide keybinds (vim-inspired)
## Framing

Make the **entire app operable by keyboard alone** — vim-inspired. This issue is the keybind/navigation layer of that push (see the epic).

A centralized, discoverable keymap; vim-y sequences; hint-based room switching that scales past 10; strict input-focus guarding.

### Room switching — left nav (`rooms-sidebar.tsx`)
- **Hold-to-reveal quick switch:** hold a leader/modifier → each room avatar overlays a **hint label**; press it to jump.
- **Scales past 10 (the open question):** don't rely on digits 1–9 alone. Use **Vimium-style hint labels** — single home-row chars (`a s d f g h j k l …`) for the first N rooms, auto-expanding to 2-char hints once rooms exceed the single-char alphabet. Digits `1–9` can stay as a fast-path alias for the first nine.
- **Sequential cycle:** next/prev room (e.g. `[` / `]`, or `J` / `K`).

### Room panes (`src/app/room/[name]/page.tsx`)
- Jump directly to each pane: **channel / negotiate / l9 / plan / slim** (direct keys or a leader sequence).

### Right-rail tabs (`room-inspector.tsx` → Members / Episodes / Memory)
- Keys to switch among the agents / episodes / memory tabs.

### Focus targets
- Focus the **chat input** (`room-chat-box.tsx`).
- Focus / open the **Memory** tab (and the other rails).
- `Esc` blurs the current input back to "command mode" (the modal, vim-y feel).

## Guards & robustness
- **Keybinds must NOT fire while a text input / textarea / contenteditable is focused** (especially the chat box). Typing is typing; `Esc` is the escape hatch back to command mode.
- Cross-platform modifiers (⌘ on macOS, Ctrl elsewhere).
- **Central keymap registry** = single source of truth: bindings declared once, conflict-checked, and reused by the command palette and the cheatsheet.
- **Discoverability:** a `?` overlay / cheatsheet listing the bindings active in the current context.
- **Accessibility:** visible focus rings, correct roles/ARIA, respect `prefers-reduced-motion` for the hint overlay.

## Open questions
- Leader key choice (`g`? space? a dedicated modifier?).
- Hint alphabet + tie-break ordering (nav order vs recency).
- Pane/tab keys direct vs leader-prefixed.

juliarvalenti:
## Summary

Makes the workspace operable from the keyboard alone, vim-style — the keybind/navigation track of the keyboard-first epic.

Everything hangs off **one central keymap** (`mycelium-frontend/src/lib/keymap.ts`): each binding is declared once with an id, its key sequences, a label, a group and a scope. The dispatcher and the `?` cheatsheet both read that table, so a binding can't exist in one surface and be missing from the other — and `findConflicts` (duplicate sequences, prefix shadowing, reused ids) is asserted empty by a unit test, so adding a colliding key is a red gate rather than a key that silently stops working. #573's palette can read the same table.

### Answers to the issue's open questions

- **Leader:** `g` ("go"), one leader covering both panes and rails without collision.
- **Hint alphabet + ordering:** home row `asdfghjkl`, labels assigned in nav order (the rooms as listed, so a filtered sidebar switches among what it shows). Labels are uniform width and therefore prefix-free — single chars while the list fits the alphabet, widening to 2 (then 3) chars past it. `1`–`9` stay as the fast path for the first nine.
- **Panes/tabs:** leader-prefixed, so single letters stay free for the modal verbs (`f`, `i`).

## Changes

**New**
- `src/lib/keymap.ts` — the binding table plus chord/sequence matching, cross-platform chord formatting (`mod+` → ⌘ on macOS, Ctrl elsewhere), conflict detection, and the hint-label generator.
- `src/components/keymap-provider.tsx` — hosts one keydown listener in the `AppShell`; resolves chords into leader sequences and dispatches to whichever component registered the action (`useKeyAction` / `useKeyScope` / `useKeyCapture`). Shows a pending-leader indicator that lapses after 1.5s, and a status-bar `? keys` affordance so `?` is findable without knowing `?`.
- `src/components/keymap-cheatsheet.tsx` — the `?` overlay, listing the bindings live in the current context.

**Bindings**

| Keys | Action |
| --- | --- |
| `f` | Reveal hint labels over the room avatars — hold to peek, tap to latch, press a label to jump |
| `]` / `[` (or `J` / `K`) | Next / previous room |
| `1`–`9` | Jump to one of the first nine rooms |
| `g h` | Command center |
| `g c` `g n` `g l` `g p` `g s` | Channel / Negotiate / L9 / Plan / SLIM pane |
| `g a` `g e` `g m` | Members / Episodes / Memory rail (opens it if collapsed) |
| `\` | Collapse / expand the inspector rail |
| `i` | Focus the composer |
| `Esc` | Leave the input — back to command mode |
| `?` | Keyboard shortcuts |

**Guards**
- Nothing fires while an `input`, `textarea`, `select` or contenteditable has focus; `Esc` is the only key honoured there, and it blurs. The chat box's own `Esc` (dismissing its mention popover) wins first — a second `Esc` blurs.
- Keys are inert behind a modal, which owns its own `Esc`.
- The platform's *other* command modifier maps to an inert prefix, so ⌃C on macOS never fires the binding on `c`.
- IME composition and bare modifier keydowns are ignored.

**Accessibility**
- A global `:focus-visible` ring — the app had none, which matters once it's driven by keyboard.
- `role="dialog"` / `aria-modal` / focus capture + restore on the cheatsheet, `role="status"` live regions on the hint overlay and the pending-leader indicator.
- `motion-safe:` on both overlay animations.

## Testing

- `src/lib/keymap.test.ts` (11) — no conflicts in the shipped keymap; conflict detection catches duplicates, prefix shadowing and reused ids; scope gating; multi-key bindings; platform chord formatting; hint labels stay unique, prefix-free and correctly widened.
- `src/components/keymap-provider.test.tsx` (7) — leader sequences fire their action, an unknown second chord drops the sequence, room bindings stay inert outside a room, nothing fires while typing and `Esc` returns to command mode, `?` opens a cheatsheet scoped to the live context, pending leader is displayed.
- `src/components/rooms-sidebar.test.tsx` (8) — hint reveal and jump, 2-char widening past nine rooms with partial-then-commit typing, hold-to-peek vs tap-to-latch, `Esc` exits without navigating, next/prev cycling and wrap-around, digit fast path, and switching restricted to what the filter leaves on screen (with the filter box swallowing keys while focused).

- [x] Unit tests pass (`pnpm test` — 68 passed, 9 files)
- [x] Linting passes (`pnpm lint` — `tsc --noEmit` clean)
- [x] `next build` succeeds

Backend untouched, so the Python gates in the template don't apply.

## Related Issues

Closes #572 — part of the keyboard-first epic #575.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JaHZwZXHTCAgJAhHbdzzpg)_

### DIFF
--- a/mycelium-frontend/src/app/globals.css
+++ b/mycelium-frontend/src/app/globals.css
@@ -285,15 +285,6 @@ body {
   transform: translateY(0.5px);
 }
 
-/* ── Keyboard focus ──────────────────────────────────────────────────────
- * The app is meant to be driven from the keyboard, so where focus sits has to
- * be visible everywhere by default — pointer clicks stay unringed. */
-:focus-visible {
-  outline: 2px solid var(--accent);
-  outline-offset: 2px;
-  border-radius: 4px;
-}
-
 /* ── Onboarding tour (driver.js popover, themed to the app tokens) ── */
 .driver-popover.mycelium-tour {
   background: var(--elevated);
--- a/mycelium-frontend/src/app/globals.css
+++ b/mycelium-frontend/src/app/globals.css
@@ -285,6 +285,15 @@ body {
   transform: translateY(0.5px);
 }
 
+/* ── Keyboard focus ──────────────────────────────────────────────────────
+ * The app is meant to be driven from the keyboard, so where focus sits has to
+ * be visible everywhere by default — pointer clicks stay unringed. */
+:focus-visible {
+  outline: 2px solid var(--accent);
+  outline-offset: 2px;
+  border-radius: 4px;
+}
+
 /* ── Onboarding tour (driver.js popover, themed to the app tokens) ── */
 .driver-popover.mycelium-tour {
   background: var(--elevated);
--- a/mycelium-frontend/src/app/layout.tsx
+++ b/mycelium-frontend/src/app/layout.tsx
@@ -5,6 +5,7 @@ import type { Metadata } from "next";
 import "./globals.css";
 import { ThemeProvider } from "@/components/theme-provider";
 import { CurrentUserProvider } from "@/components/current-user";
+import { KeymapProvider } from "@/components/keymap-provider";
 import { NotificationsProvider } from "@/components/notifications-provider";
 
 export const metadata: Metadata = {
@@ -25,7 +26,11 @@ export default function RootLayout({ children }: { children: React.ReactNode }) 
       <body className="min-h-screen bg-bg text-text antialiased">
         <ThemeProvider attribute="class" defaultTheme="dark" enableSystem disableTransitionOnChange>
           <CurrentUserProvider>
-            <NotificationsProvider>{children}</NotificationsProvider>
+            <NotificationsProvider>
+              {/* Above the pages, not inside a page's shell: a page registers
+                  its own bindings, so it has to sit under the provider. */}
+              <KeymapProvider>{children}</KeymapProvider>
+            </NotificationsProvider>
           </CurrentUserProvider>
         </ThemeProvider>
       </body>
--- a/mycelium-frontend/src/app/room/[name]/page.tsx
+++ b/mycelium-frontend/src/app/room/[name]/page.tsx
@@ -13,6 +13,7 @@ import { RoomInspector, type Tab } from "@/components/room-inspector";
 import { RoomTour } from "@/components/room-tour";
 import { GlobalStatusItems, StatusButton } from "@/components/status-items";
 import { ActingAsPicker } from "@/components/acting-as-picker";
+import { useKeyAction, useKeyScope } from "@/components/keymap-provider";
 import { useRoomStatus } from "@/lib/use-status";
 
 function episodeSummaryLabel(episodes: EpisodeSummary[] | null): { text: string; color: string } | null {
@@ -72,6 +73,21 @@ export default function RoomPage() {
     setInspectorTab(tab);
     setInspectorOpen(true);
   }, []);
+
+  // Room-scoped keybinds: the panes, the inspector rails, and the composer are
+  // all reachable without a pointer. The chat box focuses the textarea itself;
+  // this only makes sure the pane holding it is the one on screen.
+  useKeyScope("room");
+  useKeyAction("pane.channel", () => setEditorView("channel"));
+  useKeyAction("pane.negotiate", () => setEditorView("negotiate"));
+  useKeyAction("pane.l9", () => setEditorView("l9"));
+  useKeyAction("pane.plan", () => setEditorView("plan"));
+  useKeyAction("pane.slim", () => setEditorView("slim"));
+  useKeyAction("rail.agents", () => openTab("agents"));
+  useKeyAction("rail.episodes", () => openTab("episodes"));
+  useKeyAction("rail.memory", () => openTab("memory"));
+  useKeyAction("rail.toggle", () => setInspectorOpen(open => !open));
+  useKeyAction("focus.chat", () => setEditorView("channel"));
 
   const episodeLabel = useMemo(() => episodeSummaryLabel(episodes), [episodes]);
 
--- a/mycelium-frontend/src/components/app-shell.tsx
+++ b/mycelium-frontend/src/components/app-shell.tsx
@@ -5,6 +5,7 @@
 
 import type { ReactNode } from "react";
 import { RoomsSidebar } from "@/components/rooms-sidebar";
+import { KeymapHelpButton, KeymapProvider } from "@/components/keymap-provider";
 
 interface Props {
   /** The open room (highlights its sidebar row); null on the home view. */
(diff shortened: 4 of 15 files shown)


juliarvalenti:
Tested this locally (mock frontend) — the keyboard nav needs rework before it lands. Three problems, roughly in priority order:

### 1. The bindings don't actually fire
Pressing the pane shortcuts (`g c`, `g n`, `g l`, …) does nothing in a room — the panes never switch. So the core feature isn't working as shipped; please debug why the room-scoped binds aren't triggering.

### 2. They're completely obfuscated — no discoverability
Nothing on the UI hints that a shortcut exists. The tabs (Channel / Negotiate / L9 / Plan / SLIM) and the rail items give no visual cue, so unless you happen to open `?` you'd never know the binds are there — and even then, two-key `g`-sequences are a lot to memorize for the *primary* navigation.

### 3. Desired direction — VSCode-style modifier binds + hold-to-reveal
Rather than gmail-style `g`-chords, make these feel like an editor:
- **Use a modifier (Cmd/Ctrl + key)** for the pane/rail switches, so they're single, instant, muscle-memory chords instead of leader sequences.
- **Hold-to-reveal overlay:** while the modifier is held, show a tiny badge over each selectable item with its key — e.g. hold ⌘ and `[L]` appears over **L9**, `[C]` over Channel, etc. Release → badges disappear. That makes the whole scheme self-teaching: you never have to look anything up.

**If you keep a chord scheme instead of modifiers**, then at minimum the binds must be **visible on the targets** (the key shown on/near each tab and rail item), because right now it's totally hidden.

Net: fix the dead binds (#1), and pick either (a) modifier binds + hold-to-reveal, or (b) chords with the keys shown on the items. The current hidden two-key sequences are the worst of both.

juliarvalenti:
All three addressed in `1f3ce37`. Went with option (a) — modifier binds + hold-to-reveal.

## 1. The dead binds — root cause found

`KeymapProvider` lived in `AppShell`. But the room page calls `useKeyAction`/`useKeyScope` in its own component body and *renders* `<AppShell>` — so the page sat **above** the context it was registering into. Every room-scoped binding went into a null provider and the `room` scope never went live. The sidebar and chat box are part of `children`, hence rendered *inside* the provider — which is exactly why room switching and `Esc` worked for you and the panes didn't.

Two changes: the provider moves to the root layout, above every page; and `useKeyAction`/`useKeyScope` now **throw** without a provider instead of silently no-oping, the same way `useNotifications` does. That's the real fix — this class of mistake is now a crash in dev, not a dead key. There's a test pinning it.

## 2 + 3. Modifier binds, and the keys live on the targets

Hold <kbd>⌥</kbd> and every navigable target wears its own key — pane tabs, rail tabs, the first nine room avatars, the brand. Release and they're gone. `KeyBadge` reads the chord straight from the keymap, so a badge can't disagree with what the key actually does. Both badge variants are positioned out of flow, so holding the modifier never reflows a tab strip (the first pass widened the rail and clipped Memory).

The `g`-sequences are gone entirely, along with the sequence machinery that served them.

| | |
| --- | --- |
| <kbd>⌥C</kbd> <kbd>⌥N</kbd> <kbd>⌥L</kbd> <kbd>⌥P</kbd> <kbd>⌥S</kbd> | Channel / Negotiate / L9 / Plan / SLIM |
| <kbd>⌥A</kbd> <kbd>⌥I</kbd> <kbd>⌥M</kbd> | Members / Episodes / Memory |
| <kbd>⌥1</kbd>…<kbd>⌥9</kbd> | first nine rooms |
| <kbd>⌥H</kbd> | command center |
| <kbd>]</kbd> <kbd>[</kbd> (or <kbd>J</kbd> <kbd>K</kbd>) | cycle rooms |
| <kbd>f</kbd> | label *every* room — the answer past nine |
| <kbd>i</kbd> / <kbd>Esc</kbd> / <kbd>\</kbd> / <kbd>?</kbd> | composer / command mode / collapse rail / cheatsheet |

### Two things worth flagging on the modifier choice

**It can't be ⌘/Ctrl.** The mnemonic chords there are all taken by the browser and not interceptable by a page: ⌘C copy, ⌘L address bar, ⌘N new window, ⌘P print, ⌘S save. ⌥ alone is the one single modifier a page can actually hold, so that's what reveal and the chords use. There's a unit test asserting no binding claims `mod+`.

**⌥E is skipped** (Episodes is ⌥I) — Alt+E opens the browser menu on Windows and Linux.

Also: a modified chord now resolves through `event.code` rather than `event.key`, because ⌥C reports `"ç"` on macOS and any non-QWERTY layout moves the letters around.

## One more thing your report surfaced

The modal guard tested whether *focus* was inside a dialog. Clicking a dialog's backdrop moves focus out without closing it, which would have leaked keys to the surface behind an open consent prompt. It now tests for an open dialog's presence. Test added.

## Verification

Since unit tests missed a functional bug, I drove the real app in a browser this time — Playwright against `next start` with `MYCELIUM_UI_MOCK=1`, 22 checks, all passing: every pane switch, every rail switch, reveal-on-hold showing the right badges and clearing on release, `i` focusing the composer, typing not firing binds, `Esc` returning to command mode, binds working again after, `⌥2` switching rooms, and the cheatsheet opening and closing.

Unit suite is 75 passing across 9 files; `tsc --noEmit` clean; `next build` succeeds.


---

Issue #579, opened 2026-08-16
SLIM channel identity via SPIRE/SPIFFE — workload identity on the MLS group (follows #476)


juliarvalenti:
The flagship AGNTCY/SPIFFE story for the SLIM channel, **following the JWT slice (#476)**. Where #476 replaces the shared-secret PSK with slim-bindings' native **JWT** identity provider (fed by our OIDC tokens), this adds **SPIRE/SPIFFE** as a second provider behind the same seam: **JWT-SVIDs from the Workload API** for zero-secret bootstrap, workload attestation, and automatic rotation — per-member MLS identity with real revocation.

## Why it's a great SLIM showcase
slim-bindings exposes this natively (per the #476 audit): `IdentityProviderConfig.SPIRE` / `IdentityVerifierConfig.SPIRE`, `SpireConfig(socket_path, target_spiffe_id, jwt_audiences, trust_domains)`. It's upstream's **recommended production setup**. Demonstrating it end-to-end (agent workload → SVID → MLS group member, no shared secret, auto-rotating) is the "enterprise-grade SLIM" story.

## Scope
- Wire `SpireConfig` into `slim_client.py`'s identity seam (the same one #476 introduces for JWT), selectable by config.
- SLIM **node** SPIRE verifier config + a SPIRE server/agent quickstart (dev): the heavier infra this path implies.
- Reuse the #564 agent-credential seam: `MYCELIUM_AGENT_AUTH_TOKEN` already accepts a JWT-SVID for the HTTP plane; extend the same SVID to the SLIM channel so **one workload identity covers both planes**.

## Constraints
- **Optional, off by default** (#567): the dev PSK stays the default; JWT (#476) is the light opt-in; SPIRE is the heavier opt-in. Its server+agent+attestation weight must never land on the try-it path.
- CLI/backend SLIM primitive duplication stays contract-tested (`contracts/slim-l9-wire.json`).

## Depends on
- **#476** (JWT slice — establishes the pluggable identity seam SPIRE drops into)
- #564 (agent credentials — the SVID source) · #560 (identity epic) · design `docs/design/identity-and-auth.md`

juliarvalenti:
## Elevated: SPIRE is now the *recommended* path, not a later tier

[...]

Issue #564, opened 2026-08-16
Identity: per-agent credentials — OIDC service-account (default) + optional SPIRE upgrade


juliarvalenti:
Part of #560. Related: #476. Design: `docs/design/identity-and-auth.md` (PR #559).

Agents are **workloads** — their identity should bootstrap without a human, rotate automatically, and attest what they are. Primary path is **SPIRE/SPIFFE** (JWT-SVIDs from the Workload API), which slim-bindings supports natively and upstream recommends (#476).

## Scope
- Per-agent identity via **SPIRE JWT-SVID** (Workload API): zero-secret bootstrap, auto-rotation. The SVID `sub` (SPIFFE ID) → the agent's canonical handle.
- **OIDC service-account** (client-credentials) fallback for environments without SPIRE.
- Agent presents this credential to the HTTP-API gate and (per #476) to the SLIM channel — ideally one token both planes accept (see design open question).
- Wire into `agent`/`engine` registration so a registered agent has an issued identity, not a self-asserted handle.

## Acceptance
- A registered agent authenticates to the API with its own credential; its writes attribute to its verified handle.
- No shared secret needed for a new agent; revoking one agent doesn't affect others.

juliarvalenti:
Reframed per the design update (#567): **agents default to an OIDC service-account** (same seam as humans — nothing extra to run an agent). **SPIRE is an *optional* hardening upgrade** for zero-secret bootstrap / auto-rotation / workload attestation, never the default and never required — its server+agent+attestation weight must not land on someone just trying the app. Build the OIDC service-account path first; SPIRE is additive behind a config switch.

juliarvalenti:
## Handoff / build brief — per-agent credentials

**Mission:** give each agent its **own** verified identity. Default path is an **OIDC service-account** (client-credentials) token; **SPIRE JWT-SVID** is an *optional* hardening upgrade (never required — see the off-by-default principle in `docs/design/identity-and-auth.md`). The agent presents its credential to the gate (#561) so its writes attribute to its verified handle (#562). This also **unblocks #571** (await queue authorization).

### Default: OIDC service-account
- Each agent = its own OIDC client (`client_id` + secret). Token `sub` = `client_id` = the agent's handle — exactly the shape verified against the #566 dev issuer (`client_credentials` → `sub: <client_id>`).
- No shared secret; revoking one agent's client doesn't touch others (the property the D1 PSK lacked).

[...]

Issue #563, opened 2026-08-16
Identity: CLI human login flow (mycelium login → OIDC token)


juliarvalenti:
Part of #560. Design: `docs/design/identity-and-auth.md` (PR #559).

Humans need to obtain and refresh an OIDC token for the authenticated HTTP API. `mycelium iam <name>` currently just self-asserts a handle (`identity.py`); replace/augment it with a real login.

## Scope
- `mycelium login` — OIDC **Authorization Code + PKCE** (browser), with a **device-code** fallback for headless/CI.
- Securely store the token + refresh token; auto-refresh on expiry; attach `Authorization: Bearer` to backend calls.
- `mycelium iam` / `whoami` reports the identity from the token.
- Issuer-agnostic config (works against Keycloak or any OIDC issuer).

## Out of scope
- Agent credentials (separate issue — agents use SPIRE/service-accounts, not interactive login).

## Acceptance
- `mycelium login` against a configured issuer yields a working token; subsequent `mycelium memory`/`room` calls authenticate.
- Expired access token refreshes transparently.

juliarvalenti:
## Handoff / build brief — CLI human login

**Mission:** `mycelium login` → obtain an OIDC token for a human, store + refresh it, and attach it as `Authorization: Bearer` on backend calls so a human authenticates against the enabled gate (#561, merged). Then `mycelium iam` / `whoami` reflects "who the token says I am" — satisfying the acceptance line carried over from #562.

### The hard rule (same as the whole epic): opt-in, off by default
No token, no change. If the user hasn't logged in — or the hub's gate is off — the CLI behaves exactly as today (no header, self-asserted handle). Login is purely additive; the try-it path is untouched.

### Flows
- **Authorization Code + PKCE** (browser) — primary, for interactive humans.
- **Device-code** — fallback for headless/SSH/CI.

### Token storage (not config.toml)
Tokens are secrets — do **not** put them in `config.toml`. Store access + refresh + expiry in a dedicated cache (`~/.mycelium/token.json`, mode 0600) or the OS keyring. **Auto-refresh** on expiry.

### Inject the Bearer — centralize the client factory (important)
There are currently **two** backend-client builders plus several raw httpx clients, and none attaches auth:
- `commands/memory.py:42 _get_client()` → `Client(base_url=cfg.server.api_url, ...)`
- `commands/room.py:50 _typed_client()` → same
[...]

PR #581, opened 2026-08-16
feat(slim): per-member JWT channel identity behind a config seam (#476)

juliarvalenti:
## Summary

Replaces "every member of a room shares one derived secret" with an **opt-in tier where each member presents its own bearer token** as its SLIM channel identity — the *same* token it already sends to the hub's HTTP API. One `mycelium login`, or one agent service-account credential (#563/#564), now authenticates both planes. Identity on the channel is the token's `sub`, which is the handle, so members are cryptographically distinct MLS participants and revoking one client leaves every other member untouched.

**The shared-secret PSK stays the default.** An install that configures nothing calls `create_app_with_secret(name, mint_shared_secret(...))` exactly as before — the try-it path is byte-for-byte unchanged, per #567.

Three things the audit's shape didn't predict, worth reading before review:

1. **Per-member identity needs no node change at all.** `IdentityVerifierConfig` is app-level — the second argument to `service.create_app(name, provider, verifier)` — and it is how a member verifies *its peers*. The node forwards ciphertext and never inspects the MLS identity, so this works against the stock `ghcr.io/agntcy/slim:2.0.0` with zero node configuration.
2. **The node's JWT axis is a different one.** `ServerConfig.auth: ServerAuthenticationConfig{BASIC | JWT}` gates whether a client may open a *dataplane connection* — transport, not membership. Optional and orthogonal. The client half is wired (same token, as `ClientConfig.auth`); the node half is **proposed and unverified** in `docker/compose-slim-jwt.yml`.
3. **The provider is `STATIC_JWT`, not `JWT`.** `IdentityProviderConfig.JWT(ClientJwtAuth)` makes the client *sign its own* JWT from a private key — a second credential, the opposite of the goal. `STATIC_JWT(StaticJwtAuth(token_file, duration))` presents an externally-issued bearer token and auto-reloads on file change, so a refreshed OIDC token is picked up without a reconnect.

## Changes

**The seam** — `fastapi-backend/app/services/slim_identity.py` and its CLI mirror `mycelium-cli/src/mycelium/slim/identity.py`. Tier selection, token resolution, and the provider / verifier / connection-auth trio handed to slim-bindings. Resolution is **env-only on both sides**, the same shape `MYCELIUM_SLIM_MASTER_SECRET` already has, so the two copies stay byte-for-byte comparable — they diff only in their module docstrings.

**Contract** — `contracts/slim-l9-wire.json` gains an `identity` block freezing the mode strings, defaults and env names; both twin wire tests assert against it, so neither copy can drift without a red gate.

**Token materialization** — `STATIC_JWT` reads from a file, so an in-memory token lands at `<data-dir>/slim-identity/<handle>.jwt` written `0600` via `os.open` (never momentarily default-mode), and is rewritten only when the contents actually change. A caller-owned `token_file` — a SPIRE JWT-SVID, or CI's — is passed through untouched, which is the shape #579 drops into beside `STATIC_JWT`.

**Client factory** — `mycelium/client.py` grows `access_token(config, handle=…)`; `auth_headers` is now a thin wrapper over it, so the SLIM plane and the HTTP plane resolve the credential through the identical path (agent credential wins over human session, unchanged).

**Config** — `[slim]` gains `identity`, `identity_issuer`, `identity_audience`, `identity_jwks`, `identity_jwks_file`, `identity_token_file`, `identity_algorithm`, `identity_duration_s`. Issuer and audience default to the HTTP plane's (`agent_auth`, then `login`), so the common case configures neither. `mycelium config apply` renders them into the container env. A bearer token is deliberately *not* read from `config.toml`.

**Degradation** — `jwt` with no resolvable token joins on the PSK with a one-time warning (the "gate off / no token → PSK behavior" acceptance). `MYCELIUM_SLIM_IDENTITY_REQUIRE=1` fails closed instead, mirroring `MYCELIUM_SLIM_REQUIRE_SECRET`.

**Docs** — Authentication guide gains a "The same token joins the SLIM channel" section; `docs/design/identity-and-auth.md` rollout step 5 marked landed with the two findings, and the "one token or two?" open question closed; `docs/cross-machine.md` notes JWT identity sidesteps the shared-secret parity trap; `CLAUDE.md` D1 bullet rewritten. Docs site regenerated.

## Testing

- [x] Unit tests pass (`uv run pytest tests/ -x -q`) — backend **505 passed, 6 skipped**; CLI **402 passed**
- [x] Linting passes (`uv run ruff check .`) — both packages
- [x] Format check passes (`uv run ruff format --check .`) — both packages
- [x] `uv run ty check .` clean on both packages

New suites `fastapi-backend/tests/test_slim_identity.py` + `mycelium-cli/tests/test_slim_identity.py` (twins, plus CLI-only coverage of the `[slim]`→env adapter): tier selection and rejection of an unknown tier, caller-token-beats-env precedence, the `0600` token file and its no-op rewrite on an unchanged token, handle path-escape folding, provider / verifier / connection-auth construction against the real bindings, unknown-algorithm and unusable-duration refusals, PSK-default-builds-nothing, and both degradation paths.

**Not verifiable offline** — this is a live-node slice, so the payoff needs your rig:

- an MLS join with JWT identity against a node configured for it, and per-member distinctness
- an invalid / expired / revoked token being refused
- one token passing the HTTP gate *and* joining the channel
- whether `JwtKeyType.AUTORESOLVE()` (the default when no JWKS is pinned) resolves keys from the token's issuer — `slim.identity_jwks` / `_jwks_file` is the escape hatch if not
- **`docker/compose-slim-jwt.yml`'s node config spelling.** Faithfully transcribed from the `ServerConfig.auth` → `JwtAuth{key, audience, issuer, subject, duration}` struct, but the node is a separate Rust binary and its serde field/tag names are not confirmed. Flagged in-file rather than guessed silently.

## Related Issues

Refs #476 · depends on #564 (token source, merged) · SPIRE follow-up #579 · design `docs/design/identity-and-auth.md`

https://claude.ai/code/session_017XZeVrcR54pYCg7fQJ1vcq

---
_Generated by [Claude Code](https://claude.ai/code/session_017XZeVrcR54pYCg7fQJ1vcq)_

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -193,11 +193,19 @@ is no litellm dependency.
   try-it path. The localhost bypass reads the request's peer address, so it does
   **not** fire for a containerized backend (published-port traffic looks like LAN
   traffic) — the local tier is served by leaving auth off.
-- **SLIM security is a shared-secret PSK today (D1).** The group key derives from
-  `MYCELIUM_SLIM_MASTER_SECRET` (set the same on every host that shares rooms);
-  `MYCELIUM_SLIM_REQUIRE_SECRET=1` makes a host fail closed rather than fall back to
-  the public dev literal. There is no per-agent identity/revocation yet: **real
-  identity (JWT/SPIRE) is a hard prerequisite before anything hosted / multi-user**.
+- **SLIM security is a shared-secret PSK by default, per-member JWT opt-in.** The
+  default group credential derives from `MYCELIUM_SLIM_MASTER_SECRET` (set the same
+  on every host that shares rooms); `MYCELIUM_SLIM_REQUIRE_SECRET=1` makes a host
+  fail closed rather than fall back to the public dev literal. It is scoped to
+  `workspace/room`, so members are indistinguishable and revocation means rotating
+  for everyone. `slim.identity = jwt` switches the plane to per-member identity:
+  each member presents **the token it already sends to the HTTP API**
+  (`app/services/slim_identity.py` + the CLI mirror `mycelium/slim/identity.py`,
+  frozen together in `contracts/slim-l9-wire.json`). Identity is the token's `sub`,
+  so members are distinct and individually revocable. Per-member identity is
+  client-side only — the node forwards ciphertext and needs no configuration;
+  the node's own `auth.jwt` is a separate transport gate. SPIRE JWT-SVIDs drop
+  into the same seam (#579).
 - **Adapter capability (be honest).** `claude_code` is proven; `cursor` is untested;
   `openclaw` and `hermes` are deprecated (they rode the removed SSE/coordination-tick
   model and have not been migrated to SLIM).
--- a/contracts/slim-l9-wire.json
+++ b/contracts/slim-l9-wire.json
@@ -32,6 +32,27 @@
     "contingency": ["negotiation"]
   },
 
+  "identity": {
+    "_comment": "Channel-identity tier selection, resolved from the environment on BOTH sides (app.services.slim_identity / mycelium.slim.identity). The two copies must agree on the variable names, the mode strings and the defaults, or a member reads a different tier than the moderator and its app presents a credential the group can't verify — a silent join failure, exactly like a diverging shared secret. Env-only for the same reason the master secret is: the backend has no config.toml and the CLI must not import the backend's Settings. `mycelium config apply` renders the authored [slim] block into these names.",
+    "modes": ["psk", "jwt"],
+    "default_mode": "psk",
+    "default_algorithm": "RS256",
+    "default_duration_s": 3600.0,
+    "token_dir_name": "slim-identity",
+    "env": {
+      "mode": "MYCELIUM_SLIM_IDENTITY",
+      "token": "MYCELIUM_SLIM_IDENTITY_TOKEN",
+      "token_file": "MYCELIUM_SLIM_IDENTITY_TOKEN_FILE",
+      "issuer": "MYCELIUM_SLIM_IDENTITY_ISSUER",
+      "audience": "MYCELIUM_SLIM_IDENTITY_AUDIENCE",
+      "jwks": "MYCELIUM_SLIM_IDENTITY_JWKS",
+      "jwks_file": "MYCELIUM_SLIM_IDENTITY_JWKS_FILE",
+      "algorithm": "MYCELIUM_SLIM_IDENTITY_ALGORITHM",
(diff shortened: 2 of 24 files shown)


juliarvalenti:
## Do not merge — live testing + deepwiki found the approach is unsupported for MLS

**Live result (this rig):** enabling `slim.identity=jwt` **panics the SLIM session** and permanently breaks the room:
```
thread 'slim' panicked at agntcy-slim-session-0.7.9/src/session_moderator.rs:150:
failed to create MLS state: MlsOp(IdentityProviderError(MlsNotSupported))
```
→ then endless `session closed` churn until the persister gives up ("channel appears permanently closed"). Reproduced on **both** a stock `slim:2.0.0` node **and** a `slim:2.1.0` node, so it is **not** node config/version. PSK default provisions cleanly (control).

**Root cause (confirmed via deepwiki on `agntcy/slim`):** the panic is provider-specific.
- **`STATIC_JWT` (StaticTokenProvider) cannot do MLS** — it *explicitly returns* `Err(AuthError::MlsNotSupported)` for the MLS signature-key methods. This PR uses `STATIC_JWT` (claim #3), so it can never work with an MLS group.
- **`IdentityProviderConfig.JWT` (SignerJwt) is the MLS-capable provider** — it manages MLS signature keys.

**Why this invalidates the PR's premise, not just the provider choice:** `STATIC_JWT` was chosen *deliberately* to present an **externally-issued** OIDC bearer token (the "same token, both planes" goal). But MLS fundamentally needs a **per-member signing key**, not a presented bearer token — which is exactly why `StaticTokenProvider` returns `MlsNotSupported`. So **"one `mycelium login` authenticates both the HTTP API and the SLIM channel" is not achievable for MLS.** The HTTP plane takes a bearer token; the MLS channel needs a signing credential. Different shapes.

**Path forward (see reframed #476 / #579):** SLIM channel identity must be a **per-member signing credential** — `SignerJwt` (local key, DIY trust-anchoring) as the floor, **SPIRE** (managed trust + attestation) as the recommended path. A de-risking SPIRE spike on a version-matched **2.2.x** node comes first (offline agents can't validate this — that's the lesson here).

**Also:** bindings resolved to `slim-bindings 2.1.0` while the node was `2.0.0` — a version skew to fix regardless (our version-pin rule).

The reusable parts (the `[slim] identity` config seam, the wire-contract `identity` block, off-by-default degradation, `0600` token materialization) are good reference for the reworked `SignerJwt`/SPIRE approach — but this PR as-is ships a tier that breaks the room when enabled. **Closing recommended in favor of the SPIRE-based rework.**

Issue #590, opened 2026-08-16
SLIM identity: member revocation / deprovisioning


juliarvalenti:
Part of #585. Depends on #476/#579.

The whole point of per-member identity over the shared PSK is **revocation without rotating everyone**. Make that real.

## Scope
- Removing an agent (`agent rm` / deprovision) revokes its channel identity: remove it from the MLS group, and revoke/expire its credential (SPIRE registration entry deletion, or drop its key from the room JWKS for the SignerJwt floor).
- Confirm remaining members keep working (no group-wide re-key required) — the property the PSK lacked.
- Short SVID/token lifetimes + auto-rotation (SPIRE gives this natively) so a leaked credential ages out.

## Acceptance
- Revoking one member removes it from the MLS group and blocks re-join with the old credential; other members are unaffected; tested.

juliarvalenti:
## Handoff — kickoff (the payoff ticket: revocation without group-wide rotation)

All the identity plumbing is merged + rig-proven: #476/#593 (SignerJwt floor), #579/#601 (SPIRE), #589/#603 (`@handle` spine + reconciliation). This ticket delivers **the reason per-member identity beats the PSK**: revoke one member without re-keying everyone.

### The model
Revocation is the inverse of the registration surface #589 built, and it differs by tier:
- **`signerjwt` floor:** revoke = **drop the member's JWK from the room roster** (`register_public_jwk`'s inverse — delete `roster/{handle}.jwk.json` / the roster entry). Once it's off the roster, peers' verifiers (static JWKS) reject that member's self-signed tokens. Per-agent, no room-wide rotation.
- **`spire`:** revoke = **delete the SPIRE registration entry** (`spire-server entry delete`) so no new SVID is minted; short SVID TTLs (already 300s in the quickstart) age out the leaked one fast. Entry deletion is external to the CLI (needs the SPIRE server) — surface the operator step, like #603's `spire` provisioning does.
- **`psk`:** no per-agent revocation exists (that's the whole limitation being fixed) — `agent rm` under psk can only note that rotating `MYCELIUM_SLIM_MASTER_SECRET` is the blunt instrument.

### Scope / acceptance
- Wire revocation into **`agent rm` / deprovision**: remove the member from the MLS group *and* revoke its credential per the active tier (drop JWK / delete SPIRE entry / operator step).
- **Prove the key property:** after revoking @bob, **@alice and the others keep working with no re-key** — the thing the PSK couldn't do. This is the acceptance test.
[...]

Issue #588, opened 2026-08-16
SLIM identity: SPIRE deployment for the appliance (optional compose profile)


juliarvalenti:
Part of #585. Depends on #579 (SPIRE mode) + #589 (handle↔SPIFFE) + #590 (revocation).

Ship SPIRE as an **optional, off-by-default** appliance tier — a **real product surface in the shipped compose**, not `docs/design` scaffolding.

## The UX bar (this is the actual point)
Turning on attested identity should be **one switch**, not a runbook. A user must never touch a SPIFFE socket path, a trust domain, or type `spire-server entry create` by hand. Setup friction is the enemy.

- **One control.** `mycelium config set slim.identity spire` → `mycelium up` brings up SPIRE server+agent automatically. The compose `spire` profile is an implementation detail the **config drives** — the user does *not* separately have to remember `--profile spire`. One source of truth, not two. Default (`psk`) brings up nothing extra; the try-it path is byte-for-byte unchanged.
- **Registration is automatic.** `mycelium agent create @alice` (spire on) registers the SVID entry against the running SPIRE server *itself* — no printed copy-paste operator step. (That printed step in #589/#603 was the honest interim; the clean end-state does it for you.)
- **Revocation is automatic.** `mycelium agent rm @alice` deletes the entry (ties to #590).
- **Legible status.** `mycelium doctor` / `/health` shows "SPIRE up, @alice attested" — so a misconfig is a clear message, not a member silently hanging at "Initializing spire identity manager."

## Scope
- A `spire` profile in the **shipped** `mycelium-cli/src/mycelium/docker/compose.yml` (SPIRE server+agent, named-volume Workload API socket, shared PID namespace, wired to the SLIM 2.1.0 node) — brought up by `mycelium up` / `hub host` when `slim.identity=spire`, **not** a loose docs/design compose.
- `mycelium up` reads the identity mode and adds the profile when it's `spire` (the user never passes `--profile` by hand).
- CLI-driven registration/revocation against the SPIRE server — replaces the printed operator step from #589/#603.
- Carry the #584 gotchas into the shipped config (named-volume socket not host bind; shared PID namespace; SLIM 2.1.0 `dataplane.servers`).
- Honest weight labeling — SPIRE (server+agent+attestation) never lands on the default path.

## Acceptance
[...]

PR #602, opened 2026-08-16
Migrate from daemon cold-spawn to resident agent loops

juliarvalenti:
## Summary

This PR completes the migration from the old cold-spawn daemon model to resident agent loops. Agents now stay awake in their own runtimes (Claude Code sessions, Cursor workspaces) and coordinate via `mycelium await --loop --exec <cmd>` / `mycelium respond`, eliminating context loss between turns. The aligner shifts from a scoring judge to an active mediator that brokers offers using NEGMAS SAO.

## Changes

**Architecture & Participation Model**
- Agents are now **resident** runtimes kept woken by looping `await` → reason → `respond`, not cold-spawned per mention
- Removed cold-spawn daemon and its associated infrastructure; participation is now a pure CLI primitive
- `mycelium await` long-polls with a durable transcript cursor; `mycelium respond` posts replies
- Presence leases ensure the backend knows which handles are resident; mentions to absent handles wait on the transcript cursor

**Aligner Engine**
- Aligner shifts from a dormant judge that scores convergence to an active **mediator** that brokers offers
- Implements NEGMAS Stacked Alternating Offers (SAO) mechanism with a persistent Pi brain session
- Addresses agents one at a time, interprets replies into SAO moves, and stops at unanimity (NEGMAS owns termination)
- Removed tunable thresholds (`ALIGNER_THRESHOLD`, `ALIGNER_MAX_ROUNDS`) — convergence is what the mechanism decides
- Kept `ALIGNER_ROUND_TIMEOUT_S` (per-agent reply timeout) and `ALIGNER_MEDIATOR_MAX_STEPS` (SAO safety bound)

**Adapter Changes**
- **Claude Code**: Installs a `/mycelium` skill (not lifecycle hooks); sessions stay resident and coordinate via CLI
- **Cursor**: Drops per-agent instructions (not daemon-managed); Cursor sessions loop `await --loop` in their workspace
- Removed OpenClaw and Hermes adapters (they rode the old SSE/coordination-tick model and are not migrated)
- Cursor adapter is now untested but expected to work; Claude Code remains the proven adapter

**Documentation & Configuration**
- Updated adapter docs to reflect resident loop model and per-agent instruction drops
- Clarified that memory is hub-resident, resolved over HTTP (no local cache)
- Updated SKILL.md and cursor rules to show `await --loop --exec` pattern
- Removed references to daemon, cold-spawn, and the old scoring judge model
- Added README files to major packages (fastapi-backend, mycelium-cli, mycelium-frontend, docs, contracts, scripts) documenting boundaries and working conventions

**Identity & Security**
- SLIM security defaults to shared-secret PSK; per-member identity (SignerJwt floor) is opt-in via `slim.identity` / `SLIM_IDENTITY`
- Identity remains off by default; full identity (SPIRE/SVID) is a prerequisite before hosted/multi-user deployment

**Cleanup**
- Removed litellm dependency; Pi is the sole LLM runtime
- Updated CONTRIBUTORS.md to reflect actual dependencies (SLIM, L9 models, NegMAS, FastAPI, sentence-transformers)
- Clarified that the backend is the hub and owns the store; other machines are thin clients

## Testing

- Existing unit tests pass (backend, CLI, frontend)
- Linting and format checks pass
- No new test coverage needed for this refactor; the resident loop model is exercised by the end-to-end suite

https://claude.ai/code/session_01Fibj6DRcwUd1EL2mi6WwWy


juliarvalenti (commented):
Reviewed the SLIM/identity prose via gh (didn't check out — you've got someone on the tree). The migration content looks right; three suggestions, all in the identity area:

### 1. The docs undersell what already shipped — name both identity tiers
The rewritten CLAUDE.md bullet says `slim.identity` "switches to the **SignerJwt floor**" and calls "full identity (SPIRE/SVID)" a *future* "hard prerequisite." But since this branch was cut, **`slim.identity=spire` merged as a working mode** (#579/#601 — SPIRE JWT-SVID → MLS, rig-verified). So the accurate framing is a 3-tier model:

- `psk` (default, off) → `signerjwt` (self-minted keypair floor, #476) → `spire` (attested SVID, #579)

Suggest rewording the "prerequisite" line: SPIRE the *mode* exists and works; what's still missing **before hosted/multi-user** is (a) turning identity on, (b) revocation/deprovisioning (#590), and (c) the appliance SPIRE deploy profile (#588). As written it reads like SPIRE is unbuilt.

### 2. That bullet has a mangled sentence
The fail-closed clause lost its subject — it currently reads roughly:
> `...per-member identity is opt-in.** / closed rather than fall back to the public dev literal.`
The `MYCELIUM_SLIM_REQUIRE_SECRET` subject dropped out. Worth a cleanup pass so the bullet parses.

### 3. `contracts/slim-l9-wire.json` is behind main — update-branch before merge
This branch's copy predates #593/#601, which added the `identity` block (+ `spire` sub-block) to this exact file. This PR only edits the `_comment` header, so a 3-way merge *should* keep main's blocks (and the contract tests would catch a regression) — but please `gh pr update-branch` and let CI re-run against the current contract before merging, so we confirm the identity/spire constants survive. (We hit an analogous "merged against stale main" situation recently; cheap to verify.)

Nothing blocking the migration itself — it's the identity descriptions catching up to the last few merges. Off-by-default framing is correct and well-stated in the outward `concepts.html` copy.

juliarvalenti:
Thanks — two of the three were right and are fixed in `fd85c14`. Point 2 I couldn't reproduce; details below.

### 1. Three tiers ✅ fixed

You're right, and I'd cut this branch before `b08456a` (#601) landed. Verified `spire` is a real mode, not a plan — `contracts/slim-l9-wire.json` on main carries `modes: ["psk", "signerjwt", "spire"]` plus the full `spire` sub-block (Workload API socket, `spiffe://{trust_domain}/agent/{handle}`, SVID audience bound to the shared MLS `audience`).

CLAUDE.md now spells out the ladder with the deploy cost of each rung, and the closing line says what you asked it to: what gates hosted/multi-user is turning identity on at all, revocation/deprovisioning (#590), and the appliance SPIRE profile (#588) — not a missing mode. The README paragraph names both opt-in tiers instead of one unnamed "signed credential".

Worth noting the generated config reference already had all three tiers correct, because it renders from the pydantic schema — the drift was confined to the hand-written prose, which is the recurring theme of this whole PR.

### 2. Mangled sentence — I don't think it is one

The bullet reads, unwrapped:

> …`MYCELIUM_SLIM_REQUIRE_SECRET=1` makes a host fail closed rather than fall back to the public dev literal.

The subject is present; "fail" and "closed" just land on either side of a line wrap. I suspect the fragment you quoted is a diff-hunk artifact — the old bullet's opening line was replaced, so `gh` shows the surviving `closed rather than fall back…` line with its subject above the hunk boundary. The rewrite moved that text into the `psk` sub-bullet anyway, so the wrap is gone either way. Happy to look again if it still reads wrong to you on the current head.

### 3. Stale contract — merged and verified ✅

Merged `origin/main` into the branch rather than update-branch, same effect. The 3-way merge behaved as you predicted: main's `identity`/`spire` blocks survived intact alongside my `_comment` edit, which only touched the header. Confirmed by test rather than by eye — `test_slim_l9_wire.py` and `test_slim_identity.py` pass on both sides (42 each), and the full CLI suite is 402 green.

While in there I fixed one more comment in the same file for the same reason the header needed it: the `knowledge` block still pointed at `daemon/connector.apply_knowledge_message` (removed module) and "bible §11" (not in the tree).

### Unrelated, but you'll want to know

`openapi.json` at the repo root has drifted from the backend independently of this PR. It still describes `/api/rooms/{room_name}/sync-mas` and the workspace/MAS memory-operations endpoints, which don't exist anywhere in `fastapi-backend/` — plus 18 CFN mentions and litellm-era health-probe text. Last refreshed Aug 13, before the auth and identity work. I left it alone since regenerating needs a live backend and drags `mycelium-client/` along, but it's worth its own PR. Same for `scripts/gen-cfn-client.sh` (targets a deleted service) and `scripts/update-pricing.py` (imports litellm).

---

PR #624, opened 2026-08-17
Rooms sidebar: activity signal (unread badges + recency order)

juliarvalenti:
The sidebar was the most polished of the three rails already — so this isn't a restyle, it's **signal**. It looked clean but told you nothing: every room identical, ordered alphabetically regardless of where anything was happening.

## Changes
- **Per-room unread badge.** Derived from the same client-side notification store the footer bell reads (`useNotifications()`), so it's basically free. A small accent count (9+ capped); unread also draws the room name brighter. Respects mutes (a muted room never badges) and **doesn't badge the room you're currently viewing** (being there is reading it).
- **Recency ordering.** Rooms sort by `last_activity` (falling back to `created_at`) so fresh ones float to the top — the same ordering the command palette already uses. Stable sort, so nothing reshuffles when timestamps tie.
- **Nit:** aligned the "Rooms" header count color with the members header (`muted`, was `faint`).

Avatars, hint/digit keyboard nav, search, and the footer are untouched.

## Tests
Added: a badge appears and counts on a room with unread activity while clean rooms stay bare; the badge stays off the room you're actively viewing (scoped to the row — global unread still reaches the bell). Existing keyboard-nav tests confirm the recency sort leaves their order intact.

`pnpm lint`, full `vitest` (149), and `pnpm build` all green.

### DIFF
--- a/mycelium-frontend/src/components/rooms-sidebar.test.tsx
+++ b/mycelium-frontend/src/components/rooms-sidebar.test.tsx
@@ -2,7 +2,7 @@
 // Copyright 2026 Mycelium Contributors
 
 import { act } from "react";
-import { render, screen } from "@testing-library/react";
+import { render, screen, within } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 import { FakeEventSource } from "@/test/fake-event-source";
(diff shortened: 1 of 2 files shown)


juliarvalenti:
Superseded by #625, which was branched from this and includes the sidebar activity-signal commit (87d3f084). Merged there.

Issue #664, opened 2026-08-17
Investigation: could a Signal-style Double Ratchet (blind-relay E2E) work for Mycelium?


juliarvalenti:
> **Investigation / research spike — not a design commitment.** Answer "could this even work, and at what cost," then decide whether it's ever worth pursuing. Explicitly optional; the current backend-moderator model + the server-side twins in #662 are the near-term path. This issue is the *far* end of the ladder.

## The question

Could Mycelium offer a **Signal-style, blind-relay end-to-end mode** — where the backend is a **dumb relay that only ever sees ciphertext** and never holds group keys — built on a **Double Ratchet**-style scheme? Investigate whether it's technically feasible here at all, and what it would cost.

## The honest starting point (why this is an investigation, not a flag)

- **SLIM's crypto is MLS (RFC 9420)** (`agntcy-slim-mls`, mls-rs, Curve25519 ciphersuite) — **not** the Signal Double Ratchet. The Double Ratchet is a **1:1** protocol; Signal's *group* messaging uses **Sender Keys**, and MLS is the modern group alternative to Sender Keys. So "add the Double Ratchet" is not a drop-in — it likely means a **parallel crypto path**, not reusing SLIM's group machinery.
- Two distinct things get conflated under "Signal-style" and the investigation must separate them:
  1. **The blind-relay *property*** (server sees only ciphertext) — potentially reachable *with MLS* by having clients hold the memberships and the backend never join the group. This is the "P2P over SLIM" shape.
  2. **The Double Ratchet *protocol* itself** — a different primitive that may or may not fit a multi-party agent room at all.
  The real question is probably (1) "is a blind-relay E2E mode feasible on SLIM," with the Double Ratchet as one candidate mechanism to evaluate against MLS-in-blind-mode.

## What to investigate

1. **Fit:** does a Double Ratchet even make sense for a *multi-party* room (N agents + humans), or does its 1:1 nature force Sender-Keys / MLS-in-blind-mode anyway? If the answer is "you'd use MLS blind, not the Double Ratchet," say so plainly.
2. **Blind-relay on SLIM:** can the backend act as a pure ciphertext forwarder — clients hold all group state, the node routes, the backend never joins? What in SLIM's session/channel model assumes a server-side member, and what would have to change?
[...]

Issue #663, opened 2026-08-17
Group invite accept/decline has no owner/allow_from authorization check


juliarvalenti:
## What's missing

Mycelium already has an owner/delegation model for handles — `mycelium agent create --owner`/`--allow-from` grants a principal the right to act as an agent (enforced via `principals.delegates_to`, checked by `may_act_as`/`authorize_handle` for `await` and, as of #659, `respond`). But **who is allowed to bring a new participant into a room's actual SLIM/MLS group is a completely separate, unauthorized surface**:

`fastapi-backend/app/routes/invites.py` — `accept_invite`/`decline_invite`:

```python
@router.post("/{invite_id}/accept")
async def accept_invite(room_name: str, invite_id: str):
    """Accept a consent prompt — invite the agent (or queue it mid-episode)."""
    _require_room(room_name)
    invite = await room_channels.manager.accept_invite(invite_id)
    ...
```

No principal check at all. Any caller who can reach this endpoint (any authenticated request under a gated hub, or anyone at all with the gate off) can accept or decline a pending invite for **any** agent into **any** room's channel — regardless of whether they own that agent, are a member of that room, or have any relationship to either.

## Why this matters

This is the actual admission-control point for the room's cryptographic group (SLIM/MLS channel membership) — it decides who gets a real seat in the room, which is a meaningfully bigger deal than a chat message. Right now it has *weaker* authorization than posting a chat message does. If mycelium is going to have a coherent identity/ownership story (the owner/allow_from model, human OIDC, SPIRE workload attestation, and — per #662 — potentially durable per-member MLS membership), invite admission needs to be a first-class part of that story, not the one gap where anyone can let anyone in.

## Scope

- Add an authorization check to `accept_invite`/`decline_invite`: at minimum, the caller should need to be a member of the room already (or the room's owner/admin concept, if one exists) to approve/deny a pending invite — not an unrelated arbitrary caller.
- Decide whether an invited agent's own `owner` should also get a say (e.g. can decline on their agent's behalf even if not yet a room member) — ties into the same delegation model `may_act_as` already implements elsewhere.
- Related to #662 (durable per-member SLIM state): if agents/humans eventually hold real MLS memberships of their own rather than routing everything through the backend's one membership, "who can invite/remove a group member" becomes an even sharper question — worth designing these together rather than fixing invite auth in isolation now and re-doing it later.
[...]

Issue #662, opened 2026-08-17
Spike: server-side per-actor "twin" SLIM/MLS sessions — real per-agent membership without blind-relay (#664)


juliarvalenti:
> **Spike** (not a commitment to rearchitect). Prove the twin model works end-to-end, measure the cost, then decide. Off-by-default and additive to the current server-held-membership model throughout.

## The decided direction: server-side per-actor "twins"

Today the backend is a **single SLIM/MLS member per room that impersonates every actor** — a service-account model. Attribution ("@alice said this") is stamped by the backend at the L9/transcript layer *after the fact*: application-level, forgeable by the backend, invisible on the wire.

The spike: move from that to the backend as a **custodian of N per-actor SLIM sessions** — one **"twin"** App per principal, each a *genuine* MLS member with **its own per-actor identity** (the SignerJwt/SPIRE credential the identity epic #585 already ships), all hosted **server-side**. The backend still dials out access to SLIM, but which actor can touch which room's data is now enforced by **MLS group membership**, not by mycelium's app logic.

This is the honest middle ground between today's service-account shortcut and #664 Mode 2 (blind relay) — and it **keeps all backend cognition** (twins live in the backend process, so the aligner / plan compiler / memory-sync / L9 still read plaintext). It does **not** require, and is **orthogonal to**, #664 Mode 2.

## Feasibility — already demonstrated
The two-distinct-members-in-one-process shape is proven on the pinned 2.1.0 stack: the #601/#603/#590 rig harnesses ran `alice` + `bob` as **separate MLS members, each with its own SignerJwt/SPIRE identity, in one process, in one GROUP session**. A twin fleet is that, scaled to one App per actor.

## What SLIM already provides (verified in `agntcy/slim`)
- **`crates/persistence`** (`agntcy-slim-persistence`): `SlimGroupStateStorage` — an MLS group-state store, `Sqlite` backend (native-only), AES-256-GCM at rest. Doc: *"State survives a restart, so a session can rejoin its MLS group without repeating the invite/welcome handshake."*
- **`crates/slim-bindings/src/persistence.rs`**: `Service::create_app_with_persistence(...)` + `App::restore_sessions(conn_id)` — the reconnect/resume call. `channel-manager` already uses `create_app_with_direction_and_persistence`.
[...]

Issue #657, opened 2026-08-17
respond (/reply) doesn't honor owner/allow_from delegation, unlike await


juliarvalenti:
## What happened

Created an agent owned by the authenticated principal:

```
mycelium agent create julia-agent --as julvalen@cisco.com --owner julvalen@cisco.com --room mycelium
```

`mycelium agent show julia-agent --room mycelium` confirms `owner: @julvalen@cisco.com`, matching the signed-in token exactly. Per the CLI's own docs (`agent create --help`, `--owner`: *"the owner may act on the agent's behalf under an enabled auth gate"*), this should let julvalen@cisco.com act as julia-agent.

`await` works as documented:

```
mycelium await --room mycelium --handle julia-agent   # succeeds — owner delegation honored
```

But `respond` (posting the reply for that same handle) is refused:

```
mycelium respond "hello, this is julia-agent" --room mycelium --handle julia-agent
Error: handle 'julia-agent' does not match the authenticated principal '@julvalen@cisco.com'; a request may only act as its own token.
```

## Root cause

`fastapi-backend/app/routes/participate.py`:

- `await_message` (line 153) calls `actor.authorize_handle(request, room_name, handle)` — this is the delegation-aware check (`may_act_as` → `principals.delegates_to`, which honors `owner`/`allow_from`).
- `post_reply` (line 200) calls `actor.bind_actor(request, body.handle, field="handle")` instead — this goes through `_authoritative`, which does **strict equality only** against the token's own principal, with zero delegation awareness.

So the two halves of the same participation protocol (`await` then `respond`) use two different authorization checks with different (and here, contradictory) semantics. An owner can watch for turns addressed to their agent but can never actually post the reply as it — the owned-agent flow is broken end to end under a gated hub, even though it's set up correctly.

## Suggested fix

`post_reply` should authorize the claimed handle the same way `await_message` does (`authorize_handle`/`may_act_as`) before binding it as the reply's author, rather than requiring strict token-handle equality.

## Repro

1. Turn the auth gate on, sign in as principal `P`.
2. `mycelium agent create <handle> --as P --owner P --room <room>`
3. `mycelium await --room <room> --handle <handle>` — succeeds.
4. `mycelium respond "..." --room <room> --handle <handle>` — 403s, contradicting the delegation `await` just honored.

Found while validating the Cisco SSO / OIDC integration on mycelium.outshift.io — tried to have an owned agent (`julia-agent`) actually post a message and hit this.

PR #665, opened 2026-08-18
Spike: server-side per-actor twin SLIM/MLS sessions (#662)

juliarvalenti:
Closes the spike ask in #662 (spike, not a rearchitecture commitment).

## What this is
Proves the decided direction end-to-end: the backend as a **custodian of N per-actor "twin" SLIM Apps**, each a genuine MLS member with its **own** SignerJwt identity, all hosted server-side in one process and joined to one room GROUP. Access becomes MLS-group-membership-enforced instead of app-logic-enforced; attribution moves onto the wire instead of being backend-stamped after the fact. All backend cognition still reads plaintext.

Additive and off-by-default. Leaves the current server-held-membership model and `persister.py` untouched.

## Validated live (stock `ghcr.io/agntcy/slim:2.1.0` node, `slim-bindings==2.1.0`)

| Q | Question | Result |
|---|----------|--------|
| Q1 | N twins in one process, distinct wire attribution | **YES** -- 3 twins, one GROUP, `distinct cryptographic wire senders: ['alice','bob','carol']` |
| Q2 | persistence API reachable from the 2.1.0 wheel | **YES** -- `Service.create_app_with_persistence_async`, `App.restore_sessions_async`, `PersistenceConfig(path, passphrase)` (methods, not free functions) |
| Q3 | resume without a re-invite | **YES** -- `restore_sessions` recovers the group from disk; revived twin transacts at the current MLS epoch, no welcome/commit replay |
| Q4 | store location + passphrase boundary | per-twin server-side SQLite, AES-256-GCM; passphrase = `HMAC(server session secret, handle)`, **not** the OIDC token |
| Q6 | cost | whole fleet multiplexes **one** dataplane connection (one socket) |

Q5 (lifecycle: ties to #663 / #590 / #656 / #657) is noted, not built.

## Honest scope boundary
Twins are **server-side**, so the hub still holds every twin's key + plaintext. This hardens the **wire + attribution** and makes **per-agent identity true at the MLS layer** (today the identity epic only identifies the backend's single App). It is **NOT** E2E-from-the-hub -- that is a different axis (#664 Mode 2), deliberately out of scope and not required by twins.

## Findings worth reading
- `create_app_with_persistence` is a **`Service` method**, not the free function the Rust source reading suggested.
- SLIM persistence resumes **crypto state only**, never missed messages -- so the durable transcript/inbox stays as the offline-replay mechanism.
- A graceful `delete_session` purges persisted state (it is a real leave, not a crash); the faithful crash sim just drops the App and lets `restore_sessions` recover from disk.
- Single-process resume has a stale-subscription confounder on inbound routing; the spike proves resume via the revived twin **sending**. A true two-process kill/restart is the clean follow-up.

## Contents
`docs/design/twin-sessions-spike/`: `README.md` (full report), `twin_sessions_spike.py`, `run.sh` (one-command repro), `build_roster_jwks.py`.

Run: `cd docs/design/twin-sessions-spike && ./run.sh` (needs docker + openssl + the `fastapi-backend` uv env).

### DIFF
--- a/docs/design/twin-sessions-spike/README.md
+++ b/docs/design/twin-sessions-spike/README.md
@@ -6,6 +6,16 @@ end-to-end on the pinned stack, measures its shape, and leaves the current
 server-held-membership model untouched so a follow-up can adopt it off-by-default
 and additively.
 
+Two rounds: **v1** (`twin_sessions_spike.py`) proved the primitive in one process;
+**v2** (`spike_v2.py`) then stress-tested the cases v1 could not, with real
+kill-able subprocesses -- restart-and-receive, epoch change while offline,
+moderator restart, wrong-passphrase rejection, multi-room resume, and the
+PSK-cannot-persist constraint. **All six v2 scenarios pass** (see the v2 matrix
+below). One apparent v1/early-v2 limitation (a restored member could send but not
+receive) was chased down, cross-checked against the SLIM source via deepwiki, and
+resolved: it is correct SLIM behavior, satisfied by mycelium's existing
+always-draining moderator loop.
+
 ## The problem it addresses
 
 Today the backend is a **single SLIM/MLS member per room that impersonates every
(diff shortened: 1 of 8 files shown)


juliarvalenti:
📓 **Findings gist (report + live run transcript):** https://gist.github.com/juliarvalenti/e051e1367ac48088b4118b4ac216f830

Same content as `docs/design/twin-sessions-spike/README.md` in this PR, plus the raw `./run.sh` output. Secret gist (unlisted).

> ⚠️ **Scope reminder for reviewers:** this is a standalone spike under `docs/design/` — it proves the primitive but is **not** wired into any product code. `create_app_with_persistence`/`restore_sessions` are referenced nowhere in `app/` or `mycelium-cli/`; `slim_client.py`, `room_channels.py`, `participate.py`, and the CLI commands are untouched. See the follow-up gaps below.

juliarvalenti:
## v2: tested everything before the final assessment

v1 proved the primitive but left the load-bearing questions open. v2 (`spike_v2.py`, run via `./run_v2.sh`) closes them with **real SIGKILL-able subprocesses** (`twin_runner.py`) — a faithful backend/twin bounce, no single-process confound — with per-test isolation so the node's per-Name multicast queue can't leak between scenarios.

**All six pass, live on `slim-bindings==2.1.0` + `slim:2.1.0`:**

| # | Scenario | Verdict | Establishes |
|---|----------|---------|-------------|
| C | PSK cannot persist | **PASS** | No `create_app_with_secret_and_persistence`; persistence needs the identity pair → **twins mandate `signerjwt`/`spire`**, not the default PSK. |
| D1 | Clean restart, receive NEW msg | **PASS** | Restored twin sends **and receives** new broadcasts; missed-while-down msg is **not** replayed (no offline replay). |
| D2 | Epoch change while offline | **PASS** | Twin resumes across a moderator rekey and still transacts — `rejoin()` heals the epoch, not stranded. |
| D3 | Moderator restart | **PASS** | The group creator restores from disk; room stays live both directions. |
| A | Wrong passphrase | **PASS** | Re-open fails `decryption failed (wrong key)` — the at-rest passphrase **is load-bearing**. |
| B | Multi-room resume | **PASS** | One `restore_sessions` returns **both** of a twin's rooms. |

**The one thing that looked like a limitation, run to ground:** in v1 a restored member could *send* but not *receive*. I cross-checked against the SLIM source via **deepwiki** (thanks for the pointer): `restore_sessions` re-establishes routes + emits an online `rejoin()`, but the **moderator only re-adds the member while it is actively draining `get_message`** — confirmed present in **2.1.0**, not a newer-branch feature. My failing runs had an idle moderator. That condition is already met in production: the backend moderator's `RoomPersister.run()` is an always-on drain loop. With a draining moderator, D1 passes in **one round**. So: correct SLIM behavior, already satisfied — not a bug, not a version gap.

**Still genuinely out of scope (honest follow-ups, not silently skipped):** scale (tested 3–4 twins, not a room's worth); lifecycle Q5 (create/teardown/revoke tied to #663/#590/#656/#657 — noted, not built); and wiring any of this into `slim_client.py`/`room_channels.py`/the CLI (still zero product-code changes — this remains a `docs/design/` spike).

Gist updated with the v2 report + raw transcript: https://gist.github.com/juliarvalenti/e051e1367ac48088b4118b4ac216f830

juliarvalenti:
## Review — spike succeeds, findings are sound and honestly scoped

Vibe-checked against independent verification (I read `agntcy-slim` `crates/persistence` + the session layer directly, and my own #601/#603/#590 rig harnesses already ran N distinct MLS members with per-member identity in one process / one GROUP). The results here are consistent with all of that:

- **Q1 (3 twins, distinct wire senders)** — matches what I saw running 2 identified members in one process; N-per-actor is the same shape scaled. ✅
- **Q2 (persistence reachable from the 2.1.0 wheel; `create_app_with_persistence` is a `Service` *method*)** — consistent with `crates/slim-bindings/src/persistence.rs`. Good catch that it's a method, not the free function the Rust reading implied.
- **Q3 (resume via `restore_sessions`, no re-invite)** — plausible and, importantly, **not overclaimed**: the spike is honest that single-process resume has a stale-subscription confounder and that it proved resume via the revived twin *sending*. The two-process kill/restart is correctly named as the clean follow-up.
- **Q6 (one dataplane connection multiplexes the fleet)** — matches my harnesses (2 apps, one conn). ✅
- **"Resumes crypto state only, never missed messages → transcript stays"** — independently confirmed: SLIM's session layer only tracks *heartbeats* as "missed," there is no message replay. So `persister.py` remains the offline-replay mechanism. This is the single most important correctness note in the report.

**Scope boundary is stated correctly** — server-side twins harden *wire + attribution + access-by-membership* and make per-agent identity true at the MLS layer, but are **not** E2E-from-the-hub (hub still holds keys/plaintext). That's the honest framing we agreed on, not oversold.

### What's proven vs. what's left
- **Proven:** the twin direction is feasible on the pinned stack, and it's the honest middle ground (cognition preserved, attribution cryptographic). The spike did its job.
- **Before productizing:** (1) the **two-process kill/restart resume** (removes the single-process stale-subscription confounder — the one real gap the report itself flags); (2) **Q5 lifecycle** — create/revoke a twin with the actor, wiring into #663 (invite authz), #590 (revocation), #656/#657 (real-IdP handles). That lifecycle is the actual implementation ticket this spike unblocks.

### Housekeeping
`docs/design/twin-sessions-spike/` is throwaway per our "no design docs / spikes in the tree" rule — captured this report + review as a gist for durable tracking so the learnings survive the eventual `docs/design/` cleanup. Recommend the spike code be dropped (or kept only until the productization ticket lands), with the rationale living in this issue + the gist.

Net: **sound spike, honest results, direction validated.** The productization follow-up is Q5 lifecycle + the two-process resume.

juliarvalenti:
Productization ticket filed: **#666** — migrate the backend from the service-account model to per-actor server-side twins. Carries forward the two open items from this spike (two-process resume + Q5 lifecycle) plus the off-by-default + honest-scope framing.

juliarvalenti:
Superseded by #667, which productizes this spike (renamed twins → **custodial sessions**), carries these spike commits forward, and **deletes** `docs/design/twin-sessions-spike/`. #667 also closes the two open items from this spike — the **two-process SIGKILL/restart resume** (verified live, no re-invite) and the Q5 lifecycle. Not merging this branch to avoid double-applying the spike commits; the report + review live in the tracking gist. Closing.

PR #667, opened 2026-08-18
feat(backend): productize server-side per-actor custodial MLS sessions (#666)

juliarvalenti:
Productizes the twin model proven in the #662/#665 spike. Migrates the backend off the **service-account model** (one SLIM/MLS App per room impersonating every actor; attribution stamped app-side and forgeable) to **per-actor server-side twin sessions**: one genuine MLS-member App per `(room, actor)`, each authenticating with the actor's own SignerJwt/SPIRE credential.

Closes #666.

> Note: this branch was cut from the spike branch, so it also carries the three `spike(slim)` commits (#662) that are not yet on `main`. This PR supersedes them and deletes the throwaway `docs/design/twin-sessions-spike/` per the no-spikes-in-tree rule.

## The change
- `respond(@alice, …)` sends through **@alice's twin** — cryptographically @alice on the wire (verifiable off the inbound `MessageContext`, not a backend-stamped L9 field).
- Room access is **MLS group membership**, not app logic (a twin is only in rooms its actor belongs to).
- All twins live in the backend process, so the moderator App + aligner / plan compiler / memory-sync / L9 still read plaintext. **Cognition is preserved.**

## Off by default (#567), non-negotiable
Twins engage **only** under an identity tier (`signerjwt`/`spire`). Under the PSK default nothing changes, byte-for-byte. `MYCELIUM_TWINS_DISABLE=1` forces the single-moderator path even under identity. The gate is structural too: `create_app_with_persistence_async` requires the identity provider/verifier pair, so a twin cannot run on PSK (spike finding C).

## What's here
- **`app/services/twins.py`** — custodian primitive: `twins_enabled()` gate, per-twin encrypted SQLite store, at-rest passphrase `HMAC(MYCELIUM_TWIN_STORE_SECRET, ws/room/handle)`, and the persistence create/join/restore calls.
- **`room_channels.py`** — custody the fleet: `invite()` stands up a real twin (concurrent listen + moderator invite) instead of a bare Name; `remove()` gracefully leaves + deletes the store (revocation, **no room-wide re-key**); `send_as_twin()`; `restore_twins()`; SignerJwt roster pre-build before the moderator verifier is snapshotted.
- **`participate.py`** — route `respond` through the actor's twin; the moderator still `ingest_local`s and dedups its own echo by L9 message id (single-copy transcript).
- **`main.py`** — `restore_sessions` every persisted twin on startup, no re-invite.

## Honest scope boundary (in code + `docs/design/twin-sessions.md`)
[...]

PR #671, opened 2026-08-18
Document two-plane security; persist hub SLIM PSK in config.toml

sfph:
## Summary

Clarifies **what PSK actually protects** in Mycelium’s hub-and-spoke model, fixes misleading hub-and-spoke docs, and makes the hub SLIM master secret **durable across `config apply`** by moving it into `config.toml`.

**Core insight:** Mycelium has **two enforcement planes**. Spokes participate over **HTTP (`:8000`)**; the **SLIM/MLS fabric (`:46357`)** is hub-side (moderator + native clients). PSK gates **native SLIM group join** on the hub — it does **not** secure spokes. Shared-hub security is **`auth.enabled` (JWT)**, not a private master secret.

### Two planes

| Plane | Port | Who uses it | Default protection |
|-------|------|-------------|-------------------|
| **HTTP API** | 8000 | Spokes, `memory`, `await`/`respond` | Open (opt-in JWT) |
| **SLIM / MLS** | 46357 | Hub backend moderator; dev `slim send` | PSK on hub |

Spokes are **thin HTTP clients**. The hub holds server-side presence (lease + durable transcript); agents never join MLS on the spoke in the default path.

### What PSK protects (and does not)

- **Does:** authenticates **native SLIM apps** joining a room’s MLS group (`MYCELIUM_SLIM_MASTER_SECRET` → HMAC per `workspace/room` → `create_app_with_secret`).
- **Does not:** HTTP memory/participation, `@handle` impersonation, spoke access, or privacy from the hub operator (backend reads plaintext for transcript/aligner/memory).
- **MLS** encrypts on the fabric; PSK is the room **gate**, not the message cipher.

### Why spokes work without the master secret

`await`/`respond` hit the hub API. PSK is irrelevant to that path — which is why setting a private PSK alone does **not** secure a shared LAN; **`auth.enabled`** does.

### Prior doc problems fixed

- Hub-and-spoke implied spokes “coordinate over SLIM” and need `:46357` + PSK sync.
- Troubleshooting pointed spokes at the SLIM node instead of `:8000`.
- Manual `.env` edits for PSK were unsafe: **`config apply` regenerates `.env` from `config.toml`** and only preserved `MYCELIUM_IMAGE_TAG`.

### Docs
- New **[Security Planes](mycelium-cli/src/mycelium/docs/guides/security-planes.md)** guide (two-plane model, deployment profiles, PSK scope).
- Rewritten **[Hub & Spoke](mycelium-cli/src/mycelium/docs/guides/hub-and-spoke.md)**: spokes → `server.api_url` `:8000`; SLIM optional; hub-only PSK via config.
- Updates to **architecture**, **troubleshooting**, **auth** (HTTP plane callout).

### Hub SLIM PSK — Option A (`config.toml` source of truth)
- `[slim].master_secret` in `MyceliumConfig`.
- **`ensure_slim_master_secret()`** on `config apply` / install: import from legacy `.env` if present, else generate `secrets.token_hex(32)`.
- Rendered to `MYCELIUM_SLIM_MASTER_SECRET` in `.env` for the backend container.
- **`config apply --migrate-env`** imports existing `.env` secret into config.
- **`config show`** masks SLIM PSK; rotation via `config set slim.master_secret`.
- Public dev literal **unchanged** as runtime fallback for tests / host `uvicorn` without Docker.

### CLI / doctor
- **`mycelium connect`**: notes that default participation uses HTTP, not SLIM.
- **`config set slim.identity signerjwt|spire`**: reminds that HTTP auth is separate.
- **`doctor`**: HTTP JWT gate status from `/health`; hub-only SLIM PSK check from `config.toml`.

## Deliberately out of scope

- Spoke-side SLIM connectors / custodial MLS (#667) — documented as future/alternate paths, not changed here.
- `MYCELIUM_SLIM_REQUIRE_SECRET` compose defaults, TLS to SLIM node, per-room PSK epoch rotation.
- Syncing PSK to spokes (they don’t use it).

## Test plan

- [x] `pytest tests/test_slim_master_secret_config.py tests/test_doctor_checks.py tests/test_docker_env_image_tag_pin.py tests/test_docker_env_auth.py`
- [ ] Fresh install → `config apply` → `config show` shows masked SLIM PSK; `.env` contains `MYCELIUM_SLIM_MASTER_SECRET`
- [ ] Re-run `config apply` → secret unchanged in `config.toml` and `.env`
- [ ] Spoke with remote `server.api_url` → `doctor` warns if HTTP auth off; no SLIM PSK check
- [ ] Hub with dev literal in config → `doctor` warns

### DIFF
--- a/docs/generate_docs.py
+++ b/docs/generate_docs.py
@@ -69,6 +69,7 @@ SECTION_CONFIG: list[tuple[str | None, str, str, str, str]] = [
     # CLI + Config blocks injected after architecture, before guides/troubleshooting.
     ("guides/structured-memory.md", "structured-memory",  "reference", "Guides",       "Structured Memory"),
     ("guides/hub-and-spoke.md",     "hub-and-spoke",      "reference", "Guides",       "Hub & Spoke"),
+    ("guides/security-planes.md",   "security-planes",    "reference", "Guides",       "Security Planes"),
     ("guides/auth.md",              "auth",               "reference", "Guides",       "Authentication"),
     ("guides/keycloak-oidc.md",     "keycloak-oidc",      "reference", "Guides",       "Keycloak / OIDC Setup"),
     ("guides/spire-identity.md",    "spire-identity",     "reference", "Guides",       "Attested Identity (SPIRE)"),
--- a/mycelium-cli/src/mycelium/commands/config.py
+++ b/mycelium-cli/src/mycelium/commands/config.py
@@ -58,6 +58,13 @@ def show(ctx: typer.Context) -> None:
             if config.llm.api_key:
                 masked = config.llm.api_key[:8] + "..." if len(config.llm.api_key) > 8 else "***"
                 typer.echo(f"  LLM API Key:  {masked}")
+            if config.slim.master_secret:
+                masked = (
+                    config.slim.master_secret[:8] + "..."
+                    if len(config.slim.master_secret) > 8
+                    else "***"
+                )
+                typer.echo(f"  SLIM PSK:     {masked}  (hub-only)")
             typer.echo(f"  Backend Port: {config.runtime.backend_port}")
 
     except Exception as e:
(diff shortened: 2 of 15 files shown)


juliarvalenti:
Pushed a small follow-up to this branch (`867e682b`): `chmod 0o600` on the secret-bearing config writes.

Since this PR moves the SLIM master secret into `config.toml`, that file (and the `config.json` snapshot, which dumps the full model) now carries two secrets: `slim.master_secret` and the pre-existing `llm.api_key`. `save()` was writing them at the default umask (usually 0644), so on a shared host they'd be world-readable. The change locks all three secret-bearing writes to owner-only (global `config.toml`, explicit-path `config.toml`, and the `config.json` snapshot), with a test asserting `0600` after `save()`.

Also fixed the red CI gate while here: `ruff format` on `commands/hub.py` (a one-line `typer.echo` collapse, formatting only).

The docs themselves are accurate and a real improvement, so no changes there.

Issue #684, opened 2026-08-19
Epic: sharpen aligner negotiation quality


juliarvalenti:
Tracking issue for a set of incremental improvements to how the aligner runs and audits negotiations — better visibility into what agents believed going in, catching terminology mismatches early, a clearer wire format for negotiation moves, and knowing how satisfied (or not) each agent actually was with the outcome.

- [ ] #679 Log opening positions as a structured snapshot before negotiation starts
- [ ] #680 Detect term-mismatch between agents before they start negotiating
- [ ] #681 Give coordination exchanges a small closed vocabulary of move types
- [ ] #682 Track how satisfied the least-happy agent was, not just whether everyone agreed
- [ ] #683 Address the least-satisfied agent next, not just the next agent in sequence
- [ ] #699 Explore a scored-candidate-pool alternative to single-offer negotiation (bigger, exploratory — tracks a structural gap, not a quick patch)

Issue #679, opened 2026-08-19
Log opening positions as a structured snapshot before negotiation starts


juliarvalenti:
Right now the aligner parses agents' opening positions straight into SAO negotiation; there's no durable, structured record of what the room believed *before* mediation began. When debugging a negotiation that went sideways, or auditing why a plan came out the way it did, it'd help to see each agent's restated understanding of the objective at the start, not just the final consensus.

**Proposal:** have the aligner synthesize a short "Opening Positions" snapshot (who's participating + what each agent believes the objective is) from the parsed positions, and record it in the episode log before mediation runs. Pure observability, no change to negotiation behavior.

**Acceptance:** every episode record gains an Opening Positions section derived from existing position text.

PR #686, opened 2026-08-19
feat(aligner): closed move vocabulary for coordination exchanges (#681)

### LINKED ISSUE #681: Give coordination exchanges a small closed vocabulary of move types
Coordination ticks currently carry mostly freeform offer text. That makes it hard to reason about a negotiation programmatically (e.g. count how many counter-offers happened, or replay one) without re-parsing prose.

**Proposal:** extend the exchange envelope with the move types the mediator already emits internally — `counter` (an offer, including the opening one) / `accept` / `reject` — just make that distinction explicit in the wire format instead of inferred from text.

**Acceptance:** the shared wire contract gets the new move types, both CLI and backend tests pass, existing negotiations round-trip unchanged.

juliarvalenti:
Closes #681. Part of #684 (sharpen aligner negotiation quality).

## What

Coordination replies carried the move type only as an unvalidated free string in `payload.data`, and `aligner._fold_reading` collapsed `counter` → `reject` before the envelope was built. So a negotiation move could not be counted or replayed without re-parsing prose.

This promotes the move type the mediator already emits (`counter` / `accept` / `reject`) to a closed vocabulary that rides the `exchange` reply envelope's `header.subkind`.

## How

- **`l9.py`** — new `EXCHANGE_MOVE_SUBKINDS = {counter, accept, reject}`, folded into `VALID_SUBKINDS[Kind.exchange]` alongside the pre-existing `team-formation`.
- **`l9_episode.record_reply`** — reads `reply["move"]` and stamps it as the envelope subkind (only when recognized; unknown/absent → no subkind).
- **`aligner._fold_reading`** — derives the real move (a proposer's offer is a `counter`, a responder's verdict is `accept`/`reject`) **without touching** the metric-folding `action`, so MPC/GAR/SCR stay byte-identical.
- **`contracts/slim-l9-wire.json`** + **CLI `slim/l9.py`** — the frozen wire contract and its byte-for-byte CLI mirror both gain the three move subkinds; both contract suites re-lock to the JSON.

Ticks are left unchanged — a tick is the mediator prompting an agent, not an agent move.

## Acceptance (from #681)

- [x] Shared wire contract gets the new move types
- [x] Both CLI and backend tests pass
- [x] Existing negotiations round-trip unchanged (subkind is optional; the metric binary is untouched)

## Design notes

- Move type lives on `header.subkind`, reusing the existing `validate_subkind` mechanism rather than a new validated payload field.
- The wire `move` is deliberately separate from the metric `action`: a proposer's offer folds to `action=accept` for the belief metrics but is a `counter` on the wire. Overloading one field would have shifted the metrics.

## Tests

7 new tests (move stamping, move derivation, unknown-move dropped, no-move round-trip). Full L9/aligner/episode/contract surface: `108 passed, 3 skipped` (backend), `10 passed` (CLI). ruff + ty clean on both sides.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,7 @@ pnpm-lock.yaml
 
 # Claude Code harness state (per-session worktrees, transcripts)
 .claude/worktrees/
+
+# portman per-worktree dev-server port leases (machine-local)
+.ports.lock
+.ports.toml
--- a/contracts/slim-l9-wire.json
+++ b/contracts/slim-l9-wire.json
@@ -45,11 +45,11 @@
   },
 
   "valid_subkinds": {
-    "_comment": "Kind -> allowed subkinds. Backend: app.services.l9.VALID_SUBKINDS (keyed by the Kind enum). CLI: mycelium.slim.l9.VALID_SUBKINDS (keyed by the plain string, since the hidden `l9 send`/`slim send` plumbing takes a raw --kind string). An empty/None subkind is always valid for any kind.",
+    "_comment": "Kind -> allowed subkinds. Backend: app.services.l9.VALID_SUBKINDS (keyed by the Kind enum). CLI: mycelium.slim.l9.VALID_SUBKINDS (keyed by the plain string, since the hidden `l9 send`/`slim send` plumbing takes a raw --kind string). An empty/None subkind is always valid for any kind. The exchange move subkinds counter/accept/reject are the closed negotiation-move vocabulary a coordination reply carries on the wire (counter = any offer incl. the opening one; accept/reject respond to the standing offer); team-formation predates them.",
     "knowledge": ["distillation", "extraction", "feedback", "query"],
     "commit": ["converged", "rejected", "resolved"],
     "intent": ["coordinator-assignment", "mission"],
-    "exchange": ["team-formation"],
+    "exchange": ["accept", "counter", "reject", "team-formation"],
     "contingency": ["negotiation"]
   },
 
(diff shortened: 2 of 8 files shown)


juliarvalenti:
Nice, this closes the gap cleanly. One ask before/after merge: could we also carry an explicit `phase` on the exchange envelope, even if for now it's just a single constant value (e.g. `"negotiate"`)? Right now the envelope only says *what* move happened (counter/accept/reject), not *what stage* of the interaction it belongs to. We don't need real phase logic yet — even a stagnant, always-the-same-value field would give us a place to hang future distinctions (e.g. a pre-negotiation clarification round vs. the negotiation itself, tying into #679/#680) without another wire migration later.

juliarvalenti:
## Self-review (medium effort)

Reviewed the diff across correctness (line-by-line, removed-behavior, cross-file caller/callee) and cleanup/altitude angles.

**Code: clean.** The move-vs-action split holds: `record_reply` leaves the metric-folding `action` untouched and only adds an optional `subkind`, so MPC/GAR/SCR stay byte-identical (all 61 prior episode tests still pass). `subkind=None` is dropped by `envelope_to_dict(exclude_none)`, so pre-vocabulary replies carry no subkind and round-trip unchanged. `_fold_reading`'s derivation is exhaustive over the three interpreter verbs plus the bare-offer case, and `move in EXCHANGE_MOVE_SUBKINDS` guards against fabricating an invalid subkind (covered by `test_reply_ignores_unknown_move`).

**One finding, fixed in the amend:** `git add -A` had swept `.ports.lock` / `.ports.toml` (portman's machine-local per-worktree port leases — they embed an absolute path + slot) into the commit. Removed them from the branch, added a `.gitignore` entry, and force-pushed. The diff is now code-only.

No further action needed.

PR #690, opened 2026-08-20
feat(aligner): catch term mismatch before the negotiation starts (#680)

### LINKED ISSUE #680: Detect term-mismatch between agents before they start negotiating
Agents occasionally use the same word to mean different things (e.g. "priority", "done", "blocked") and the aligner has no way to catch that before it starts affecting the negotiation. We've seen negotiations converge that later turn out to be talking past each other on terminology.

**Proposal:** add a cheap pre-negotiation check — given the opening positions, ask whether any agents define a shared key term differently. If so, inject one clarifying round via the existing `@`-mention mechanism before SAO offers begin.

**Acceptance:** a test scenario with an injected term mismatch reliably triggers one extra clarifying round; scenarios without a mismatch behave exactly as before.

juliarvalenti:
## Summary

Agents use the same word for different things — "done", "priority", "blocked" — and nothing caught it until the mismatch was already shaping offers. A negotiation could converge on prose each side read differently, which is worse than not converging: the plan compiles and the disagreement surfaces later.

This adds a **stage 0** to the mediator: one cheap brain call over the opening positions, before any offer exists, asking whether two participants are using a shared term in materially different senses. When it finds one, the aligner runs **exactly one** clarifying round through the existing `@`-mention seam — one turn per participant, asking only for a definition — and folds the answers into the prose issue discovery then reads. When it finds nothing (the common case) the negotiation is what it was before: no prompt, no reply wait, no extra round.

Builds directly on the opening-positions snapshot from #688: that captured what the room believed going in, this checks whether the room *means* the same thing by it.

## Changes

- **`mediator.detect_term_mismatch(positions, llm)`** — stage 0. Returns `[{"term", "readings": {handle: meaning}}]`. The prompt draws the line explicitly: wanting a different *value* for a term is the negotiation, not a mismatch, and an empty list is the expected answer. Faithful, never fabricated — a reported mismatch must name a real participant on both sides with a quoted reading each, or it is dropped rather than repaired; unknown handles, one-sided readings and malformed entries never reach the room. Capped at `MAX_TERM_MISMATCHES = 3` (a longer list reads as over-reading, not a room with five broken words) and fail-soft on an LLM error or garbage output.
- **`mediator.clarification_prompt(handle, mismatches)`** — deterministic prose, no LLM. Shows the agent every reading including its own, so a misread is correctable in-band the way the mediator already restates its SAO readings, and says plainly not to make or accept an offer yet.
- **`aligner._clarify_terms`** — runs the check after the episode opens and before `discover_issues`, then the one round via `_slim_turn` (`round=0`, `action="clarify"` on the tick payload, so a clarifying turn is distinguishable on the wire from an SAO position). A silent agent keeps its opening prose as stated. One round, never a loop: the point is to make the vocabulary visible, not to negotiate definitions.
- **`l9_episode.record_term_check`** + a `## Term Clarifications` section in `log/episodes/{id}.md`. Deliberately kept **out** of MPC/GAR/SCR — a definition is not a concession, and folding it in would score it as one. The opening-positions snapshot keeps the untouched prose, so the record still shows what each agent said before it was asked to define anything.
- **`ALIGNER_TERM_CHECK`** (default on) — an operator kill switch that skips the stage entirely, not even the one call.
- Threaded through `scripts/aligner_driver.py` (+ `result.json`) so the in-process rig from #687 exercises stage 0, and documented in the aligner docs (`How it negotiates` gains an "Align vocabulary" step; new tunable row).

## Testing

Acceptance from the issue is covered directly: a scenario with an injected mismatch triggers **one** extra clarifying round (asserted per-participant, ordered before every SAO tick, with the answers reaching discovery); a scenario without one is unchanged (every tick is a `position`).

12 new tests — mismatch parsing/filtering (real clash kept, one-sided/unknown-handle/malformed dropped), the cap, fail-soft on garbage and on a raising brain, the prompt contract, the injected round, the episode record, the no-mismatch path, the kill switch, and a broken check not costing the room its negotiation.

Backend suite: **681 passed, 9 skipped**. `ruff check`, `ruff format --check` and `ty check` all clean.

Not validated against a live model this round — no `pi` binary or LLM credentials in this environment, so the driver rig change is threaded but unexercised. Worth one live run through `scripts/aligner_driver.py` with a deliberately mismatched pair of positions before relying on the detector's precision in the wild.

## Related Issues

Closes #680. Part of #684.

---
_Generated by [Claude Code](https://claude.ai/code/session_019rU1wFabXpfJBMumsnSB2t)_

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -107,10 +107,11 @@ see issue #446.)
 engine, the **aligner** (`app/services/aligner.py`), summoned by `@`-mention. It
 runs a real **NEGMAS Stacked Alternating Offers** mechanism (`mediator.py`); its
 brain is a persistent **Pi** coding-agent session (`pi_brain.py`) so it keeps memory
-across rounds. It discovers issues from the agents' opening positions, brokers each
-round, `@`-addresses one agent at a time over SLIM, interprets each reply into an
-SAO move (`offer_snap.py` snaps near-misses / nearest numeric grid point), and
-**NEGMAS owns termination**: it stops the instant the agents agree.
+across rounds. It checks the opening positions for a term the agents are using in
+different senses (one clarifying round if so, none otherwise), discovers issues from
+those positions, brokers each round, `@`-addresses one agent at a time over SLIM,
+interprets each reply into an SAO move (`offer_snap.py` snaps near-misses / nearest
+numeric grid point), and **NEGMAS owns termination**: it stops the instant the agents agree.
 
 **Episodes.** A summon opens an **episode**, a tagged, membership-scoped negotiation
 on the room's channel (a tag over the existing channel, not a separate one), 1:1 with
--- a/docs/index.html
+++ b/docs/index.html
@@ -678,6 +678,7 @@ We chose Postgres because of myc://context/stack.</code></pre>
       <h2 id="aligner-how-it-negotiates">How it negotiates</h2>
       <p>The aligner drives a real <strong>NEGMAS Stacked Alternating Offers</strong> negotiation, so consensus comes out of the mechanism rather than an LLM improvising one. The mechanism owns proposer rotation and the unanimity stop; the aligner only supplies each agent&#x27;s move when NEGMAS asks for one.</p>
       <ol class="steps">
+        <li><strong>Align vocabulary.</strong> Before any offer exists, it reads the opening positions for a term the participants are using in <em>different senses</em> — &quot;done&quot;, &quot;priority&quot;, &quot;blocked&quot;. A word two agents read differently is a disagreement an agreement would hide rather than settle, so when it finds one the aligner runs a single clarifying round: one <code>@</code>-addressed turn each, asking only for a definition, folded into the prose the next step reads. Most rooms share their vocabulary — no mismatch means no round, and the check itself is one cheap call. <code>ALIGNER_TERM_CHECK=0</code> skips it.</li>
         <li><strong>Discover.</strong> It reads the participants&#x27; opening positions (posted with <code>mycelium respond</code>) and derives the negotiable issues and their options.</li>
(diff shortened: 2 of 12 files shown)


juliarvalenti:
## Reviewed + validated live against a real model

Ran the term-mismatch detector through the harness against `claude-haiku-4-5` (the PR flagged it hadn't been). It works end to end:

- Seeded a deliberate clash — both agents open with "ship the MVP by Q3" but **growth** means the full platform rebuild and **risk** means a thin one-module prototype.
- Real `detect_term_mismatch` caught **MVP**, and `_clarify_terms` ran **exactly one clarifying round per participant** (`action=clarify`, round 0, before any offer).
- The disambiguated prose then made `discover_issues` surface **`mvp_scope`** as an explicit 4-option ordinal issue instead of both sides "agreeing" on the word MVP. Converged on `multi-module prototype / 40%`.
- The real episode record carried the `## Term Clarifications` section (term + both readings + each agent's clarifying answer), sitting under the untouched `## Opening Positions`.

Product code is sound.

## Folded in: driver now drives the *real* `mediate()`

Addressed the one concern with this PR — that it grew the driver with a hand-mirrored copy of the stage-0 loop. Rather than mirror each new stage into the harness, `scripts/aligner_driver.py` is rewritten as a thin wrapper around the actual `AlignerEngine.mediate()`: in-repo fakes for the coordination stack, a real per-session `PiBrain`, and only the agent-reply seam swapped for an `InteractiveChannel` that surfaces each prompt and injects the human's reply into the transcript the mediator polls.

So stage 0, the opening-positions snapshot, issue discovery, the SAO loop, and the real episode-record write are all exercised **as shipped** — the validation above ran through the real `_clarify_terms`, not a reimplementation — and the harness stays a wrapper as the engine grows. `ruff` + `ty` clean; aligner/mediator/episode suites green.

PR #692, opened 2026-08-20
feat: page every list by keyset cursor; reverse-scroll room history (#654)

### LINKED ISSUE #654: Message history pagination: infinite reverse-scroll in the UI + walk-back in the CLI
## Summary

Neither the UI feed nor `mycelium room messages` can walk backward through a room's history. Both show a fixed newest-N window and stop there — once a room accumulates more than that window, older messages are effectively unreachable even though the durable transcript on disk holds them all. Add **infinite reverse-scroll** to the UI (scroll up → load older) and a **pagination / walk-back** mechanism to the CLI.

## Current state

**UI** (`mycelium-frontend/src/components/event-stream.tsx`)
- On mount it calls `fetchMessages(roomName)` with **no limit** → backend default of 50 (`src/lib/api.ts:fetchMessages`), reverses to oldest-first, and renders.
- Auto-scrolls to the bottom on new events. There is **no** load-older-on-scroll-up: no top sentinel, no prepend, no scroll-anchoring. Past ~50 messages, older history is invisible in the browser.

**CLI** (`mycelium-cli/src/mycelium/commands/room.py:messages`)
- Only `--limit` (default 20). No `--offset` / `--before` / paging. It's a point-in-time newest-N snapshot; the only way to see more is to bump `--limit` (capped by the backend at 500). No way to walk further back than that.

**Backend** (`fastapi-backend/app/routes/messages.py:list_messages`)
- Already accepts `limit` (default 50, max 500) and `offset`, sorted newest-first — so a naive offset pager is *technically* possible today.
- But it reads the **entire** conversation (`persister.room_conversation`), filters, sorts, then slices in memory (`filtered[offset:offset+limit]`). So every page is an O(n) full read, and `total` is the full count.
- The L9 wire feed (`GET .../messages/l9`) is a separate fixed replay (`limit` default 200, max 1000).

## Why offset alone isn't enough

Offset pagination is fragile against a **live-growing** transcript: new messages arriving at the head shift every offset, so reverse-scrolling a busy room duplicates or skips rows between pages. For reverse scroll we want a **stable keyset cursor**, not a numeric offset.

The room transcript (`log/transcript.jsonl`) is already an ordered, append-only log with a stable per-record index — a natural keyset anchor. This is the same transcript + per-handle cursor machinery #649 leans on (`DeliveryLog`), so the two are related: a `before=<transcript-index | message_id>` cursor reads only the requested page instead of the whole conversation, and is stable under concurrent appends.

## Proposed shape (for discussion)

**Backend**
- Add a keyset cursor to `GET /rooms/{room}/messages`: `before=<cursor>` (and/or `after=`) returning a page plus a `next_before` cursor, anchored on the transcript index or `message_id` rather than a raw offset.
- Ideally read just the page from the transcript rather than `room_conversation`-then-slice, so long rooms don't pay a full read per page. (Keep the existing `limit`/`offset` for back-compat, or migrate callers.)

**UI**
- Top sentinel / `IntersectionObserver`: when the user scrolls near the top, fetch the previous page via the cursor and **prepend**.
- Scroll-anchoring: preserve the scroll position on prepend (compute `scrollHeight` delta) so the viewport doesn't jump.
- Dedup prepended history against live SSE inserts (message_id keyed), and keep the existing "don't yank the user off a highlighted message" behavior.

**CLI**
- A walk-back mechanism: `--before <cursor>` returning the next cursor, and/or `--page`, and/or a `--all` that transparently pages and streams to a pager. Keep `--json` emitting the cursor so scripts can drive it.

## Acceptance
- UI: scrolling up in a room with thousands of messages transparently loads older pages without jumping the viewport or duplicating rows, and live messages still append at the bottom.
- CLI: a user/script can page backward through the full history deterministically, even while new messages arrive.
- Backend: paging does not re-read the entire conversation per page for large rooms.

Related: #649 (durable transcript per-handle delivery cursor — same transcript/cursor substrate), #497 (dual message store the read path ignores).

juliarvalenti:
## Summary

Neither the UI feed nor `mycelium room messages` could walk backward through a room's history. Both showed a fixed newest-N window and stopped there, so once a room grew past it the older messages were unreachable in the browser and in the CLI — even though the durable transcript on disk held them all.

Offset paging doesn't fix that on a live list. New messages arriving at the head shift every offset, so reverse-scrolling a busy room duplicates and skips rows between pages. So this pages by **keyset cursor**: a token carries the sort key of the row a page ended on, not a position, and the next page is "everything sorting strictly past this key". That boundary survives both rows arriving at the head and the anchor row being deleted mid-walk.

Per the follow-up ask, that's built as **one primitive every list shares** rather than a messages-only cursor — the same three names (`cursor` in, `next_cursor` / `has_more` out) on every list, so one client-side walker drives all of them. The issue proposed `before=`; a single direction-neutral `cursor` reads the same for a newest-first list and generalizes to the others, and `next_cursor` continues in each list's own traversal direction.

## Changes

**Backend**
- `app/services/pagination.py` — the shared primitive: opaque URL-safe cursor tokens, `paginate()` over any ordered list given a key function, and `timestamp_key()` (raw `isoformat` is not safe to compare as text — an offset suffix and optional microseconds order two equal instants arbitrarily).
- Applied to `GET /rooms/{room}/messages`, `GET /rooms/{room}/memory`, and `GET /rooms/{room}/episodes`. Existing `offset` keeps working where it already did; a malformed cursor is a 400, not a silent first page.
- Memory keeps its bare-list body (ETag sync unchanged), so it carries the same two values as `X-Next-Cursor` / `X-Has-More` headers.
- Episodes' "an in-progress one first" rule moved into the sort key, so a live episode can't drift onto page two mid-walk.
- **Paging no longer re-reads the whole conversation per page.** The transcript is append-only, so `conversational_messages` now resumes from where the last read stopped, sealed against the last consumed line so an in-place rewrite is re-read rather than stitched onto. A record caught mid-append waits for its line to finish instead of being parsed as a truncated row (that last one was a latent bug on the cold-read path).

**CLI** — one walker (`mycelium/pagination.py`), not per-command paging code
- `mycelium room messages`: `--all` walks history to the beginning, `--cursor` resumes a walk, and the two compose. `--json` carries `next_cursor` so a script can drive the walk itself; a partial page prints the command that continues it.
- `mycelium memory ls --all` walks past the first page (reading the cursor from the headers).
- The walk is bounded: a cursor that stops advancing costs one wasted page, never a hang.

**Frontend** — generic hooks, then wiring
- `use-cursor-pagination.ts` (cursor/loading/exhausted state for any paged list, with a guard so a page in flight when the reader switches rooms can't land in the new room), `use-reverse-scroll.ts` (top-sentinel loading + scroll anchoring + stick-to-bottom), `cursor-pagination.ts` (`mergeOlder`, pure).
- `event-stream.tsx`: an `IntersectionObserver` sentinel loads the page before the top and prepends it, anchored on distance from the bottom so the viewport doesn't jump; prepended history is deduped against live SSE inserts by message id; and the feed no longer yanks a reader who has scrolled up down to the newest message. A page whose rows are all filtered out of the channel view keeps walking instead of showing "No messages yet".

`openapi.json` refreshed; `docs/reference.html` updated for the two changed commands only (a full regen also pulls in ~400 lines of unrelated pre-existing drift, which didn't belong in this diff).

## Testing

All three gates run clean locally.

- Backend: 697 passed, 9 skipped. New: `test_pagination.py` (the primitive's properties — a walk covers every row once; rows arriving at the head neither duplicate nor skip; a deleted boundary row still resolves), `test_message_pagination.py` (the walk over HTTP, including under concurrent appends), `test_list_pagination.py` (memory + episodes speak the same contract), plus transcript-cache tests that **measure** incrementality by counting the lines handed to the parser.
- CLI: 500 passed. New `test_pagination.py` for the walker; `room messages` / `memory ls` walk, resume, and cursor-in-`--json` coverage.
- Frontend: 196 passed. New `event-stream.history.test.tsx` (loads the page before the top, prepends above, dedups against the live stream, doesn't yank a scrolled-up reader, keeps walking past a chat-less page, and doesn't spill a cross-room page) — the cross-room test was verified to fail with the guard removed.

- [x] `uv run ty check .`, `tsc --noEmit`, `pnpm test`

Not run here: the live-node SLIM slices and live-LLM tests (no node/keys in this environment); they skip by design.

## Related Issues

Closes #654

---
_Generated by [Claude Code](https://claude.ai/code/session_012xHEPCVM5CAAUgP6CjbvR5)_

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -159,6 +159,16 @@ is no litellm dependency.
   (version increments). Frontmatter the store doesn't manage is user data: it
   survives a rewrite rather than being dropped, and `MemoryCreate.meta` (CLI:
   `--meta k=v`, `--expandable`) writes it.
+- **Lists page by keyset cursor, not offset.** One primitive
+  (`app/services/pagination.py`) pages every list: ask with `cursor`, get back
+  `next_cursor` + `has_more`, and a null cursor means the end. The token carries the
+  sort key of the row a page ended on, so a page boundary survives rows arriving at
+  the head — the reason a reader can scroll back through a live room without
+  duplicating or skipping messages, which offset cannot do. Envelope responses carry
+  the two fields in the body; the one bare-list response (memory) carries them as
+  `X-Next-Cursor` / `X-Has-More`. `offset` still works where it already did.
+  Client-side the same contract has one walker each: `mycelium/pagination.py`
+  (`--all` / `--cursor`) and `use-cursor-pagination.ts` + `use-reverse-scroll.ts`.
 - **Memories interlink; the link index is derived.** `myc://key` (canonical) and
   `[[key]]` (shorthand) are the same edge, plus `![[key]]` transclusions and typed
   frontmatter relations (`supersedes`, `depends-on`, `part-of`, `relates-to`).
--- a/docs/reference.html
+++ b/docs/reference.html
@@ -455,9 +455,9 @@
 
       <div class="cmd-ref">
         <div class="cmd-ref-header">
-          <code><span class="bin">mycelium</span> <span class="cmd">room</span> <span class="cmd">messages</span> [<span class="arg">&lt;room&gt;</span>] [<span class="flag">--limit</span> N] [<span class="flag">--sender</span> <span class="arg">&lt;handle&gt;</span>] [<span class="flag">--type</span> <span class="arg">&lt;type&gt;</span>]</code>
+          <code><span class="bin">mycelium</span> <span class="cmd">room</span> <span class="cmd">messages</span> [<span class="arg">&lt;room&gt;</span>] [<span class="flag">--limit</span> N] [<span class="flag">--all</span>] [<span class="flag">--cursor</span> <span class="arg">&lt;cursor&gt;</span>] [<span class="flag">--sender</span> <span class="arg">&lt;handle&gt;</span>] [<span class="flag">--type</span> <span class="arg">&lt;type&gt;</span>]</code>
         </div>
-        <div class="cmd-ref-body">Read recent messages in a room (point-in-time, newest first). Filter with <code>--sender</code> / <code>--type</code>.</div>
+        <div class="cmd-ref-body">Read messages in a room, newest first. <code>--all</code> walks the full history; <code>--cursor</code> resumes from a previous page. Filter with <code>--sender</code> / <code>--type</code>.</div>
       </div>
 
       <div class="cmd-ref">
@@ -557,9 +557,9 @@
 
       <div class="cmd-ref">
         <div class="cmd-ref-header">
-          <code><span class="bin">mycelium</span> <span class="cmd">memory</span> <span class="cmd">ls</span> [prefix/]</code>
+          <code><span class="bin">mycelium</span> <span class="cmd">memory</span> <span class="cmd">ls</span> [prefix/] [<span class="flag">--all</span>]</code>
         </div>
-        <div class="cmd-ref-body">List memories from the hub. Optional prefix filters by namespace.</div>
+        <div class="cmd-ref-body">List memories from the hub. Optional prefix filters by namespace; <code>--all</code> walks past the first page.</div>
       </div>
 
       <div class="cmd-ref">
--- a/fastapi-backend/app/routes/episodes.py
+++ b/fastapi-backend/app/routes/episodes.py
@@ -14,10 +14,12 @@ GET /rooms/{room}/episodes/{short_id} — one episode + its envelope chain
 from __future__ import annotations
 
 import logging
+from typing import Any
 
-from fastapi import APIRouter, HTTPException
+from fastapi import APIRouter, HTTPException, Query
 
 from app.schemas import EpisodeDetailRead, EpisodeListResponse
+from app.services import pagination
 from app.services.episode_records import (
     EPISODES_PREFIX,
     episode_summary,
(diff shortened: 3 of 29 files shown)


juliarvalenti:
Heads up: this PR and #693 (threads) both touch `fastapi-backend/app/services/persister.py::conversational_messages()`, and not compatibly.

Test-merged the two branches in a scratch worktree to check (not merged anywhere real). Most of the overlap is fine:
- `fastapi-backend/app/routes/messages.py` — only conflicts on an import line; `list_messages`'s actual body auto-merges cleanly, and the composition is correct (the `thread` filter runs on the candidate list *before* `pagination.paginate()`, so a thread-scoped view still gets stable cursors).
- `mycelium-cli/src/mycelium/commands/room.py` — three conflicts, all mechanical: threading `--thread`/`thread=thread or UNSET` through this PR's cursor-walking `fetch()` closure, same shape as the route.
- `docs/reference.html` — generated, trivial.

The real problem is `persister.py`. Both PRs independently rewrote `conversational_messages()`'s caching:
- This PR: a proper incremental cache (`_ConversationCache` with `stamp`/`messages`/`consumed`/`seal`), `_extended()` only parses the bytes appended since the last read, via a new `_project()` helper — and that helper calls `stored_message_from_record(room, r)` with no `thread=` at all.
- #693: simple `(stamp, messages)` tuple cache, full re-parse on every change, computes `threads.thread_map(records)` over the whole transcript and attaches `thread=` per message.

These aren't reconcilable by keeping both blocks. `threads.thread_map()`'s rule 3 (a record inherits its **first parent's** thread) needs to walk the causal DAG back through however much of the transcript that takes — potentially past what's in the newly-appended slice `_extended()` looks at. And `_ConversationCache` only keeps already-projected `StoredMessage`s, not the raw `TranscriptRecord`s a full `thread_map` pass would need — so by the time you're looking at just the new lines, the parent records needed to resolve ancestry may already be gone from memory.

Whichever of these two merges first, the other will need to either:
1. Extend `_ConversationCache` with a growing `id → thread` map, so `_extended()` can resolve a new record's thread from its parent's already-known thread instead of needing the full record list — real design work, not a merge resolution, or
2. Keep this PR's incremental byte-parsing for building `StoredMessage`s, but always run `threads.thread_map()` over the full `load_transcript(room)` regardless of the cache — simpler, but gives up part of the performance win on rooms with active threads.

Flagging before either of you finishes so it's not a surprise at merge time. Cross-posted to #693.

PR #693, opened 2026-08-20
feat: threads — scope conversations inside a room (#631)

### LINKED ISSUE #631: Threads support: scope conversations within a room (not a SLIM primitive)
## Problem

A room today is a **flat broadcast transcript**. Every message is a `broadcast`
to the whole room, and `@`-mentions are the only scoping. As soon as more than
two participants are active — or agents talk to each other — the transcript
tangles: you can't tell which reply answers which prompt, parallel exchanges
interleave, and there's no unit smaller than "the whole room" to follow, mute, or
address.

This surfaced hard while testing the herdr wake integration (#540): multi-agent /
agent-to-agent chat is unreadable as a single stream, and reasoning about "who is
replying to what" is guesswork.

## Ask

First-class **threads**: a conversation scope *within* a room. A message can open
or belong to a thread; the read path and UI can group by thread; participants can
address/await a specific thread rather than the whole room.

## Is this something SLIM natively handles?

**No.** SLIM is the *transport*: a node + MLS-encrypted **group channels** +
presence. A room is one group channel (`slim_client.py`: "a channel is a Name
whose app segment is the topic"); SLIM delivers broadcasts to that channel's
members and has no notion of a sub-conversation. Threading is squarely a
**mycelium application-layer** concern.

The good news is the primitives largely already exist above SLIM:

- **L9 already carries causal parentage.** `message.parents`
  (`l9_models.py`, `l9.py:extract_parent_id` — *"Used to thread causality: a
  reply's envelope lists the tick's id in `message.parents`"*). Today it's used
  for negotiation causality (tick → reply → consensus), not surfaced as
  conversation threads — but it's the DAG we'd thread on.
- **Episodes are the precedent for scoping without a new channel.** An episode is
  *"one negotiation inside [a room]… a tag over the existing channel, not a
  separate one"* (`l9_slim.py`). A thread can follow the same pattern: a tag/scope
  over the room's single channel, **not** a second SLIM channel per thread.

So: don't reach for SLIM to deliver threads. Build them as (a) message metadata (a
`thread_id` / reuse `message.parents`), (b) read-path grouping in
`persister.room_conversation` / the messages API (a `thread` filter, mirroring the
existing `episode` filter), and (c) UI grouping in the frontend. Addressing/awaiting
a thread would extend the `await`/`respond` participation surface with an optional
thread scope.

## Sketch (non-binding)

- **Model:** `thread_id` on a message (root message id, or the first `parents`
  entry). Reuse the L9 causal DAG rather than inventing a parallel field.
- **Read path:** `GET /rooms/{room}/messages?thread=<id>` (parallels the existing
  `?episode=` filter); `room_conversation` groups/threads the merged view.
- **Participation:** `mycelium await --thread <id>` / `respond --thread <id>` so a
  turn-based agent can work one thread without the whole-room firehose.
- **UI:** collapse/expand threads in the room view (the inspector already renders
  episodes; threads are the general case).
- **Relation to episodes:** an episode is arguably a *typed* thread (a negotiation
  scope). Worth deciding whether threads generalize episodes or sit beside them.

## Why now

- Multi-agent and agent-to-agent conversations are first-class goals; without
  threads they don't scale past a toy.
- Directly unblocks the herdr multi-agent story (#540) — see that PR's "agent-to-
  agent mentions don't wake" gap; threaded addressing makes routing tractable.

_Type: design / feature. Not a SLIM change — a mycelium app-layer scope over the
room's existing channel._

juliarvalenti:
## Summary

A room is a flat broadcast transcript, so past two participants it tangles: which reply answers which prompt is guesswork, parallel exchanges interleave, and there's no unit smaller than "the whole room" to follow or address. This adds **threads**: a conversation scope *inside* a room.

Two decisions shape the implementation, both of which the issue called out as the things to settle first:

**Threads are derived, not stored.** A message already records what it answers — the L9 causal `parents` its envelope has always carried — and that edge *is* the thread. So there's no parallel `thread_id` field to keep in sync, nothing new on the wire, and a transcript recorded before threads existed reads as threaded on the first request. `respond --thread T` doesn't set a field; it parents the reply on `T`, and the derivation does the rest.

**An episode is a typed thread.** This is the fork the issue said gates the size of the work. An episode outranks causality when assigning a scope, so a negotiation's turns thread on their episode. Threads *generalize* episodes rather than sitting beside them as a second grouping — which is what keeps this a handful of additive parameters instead of a participation-cursor refactor.

The derivation rules, in order, for one transcript record:

1. On a **scoped episode** (any episode URN that isn't the room's catch-all `…:live`) → that episode's thread, keyed by its short id.
2. Declares itself a root (`payload.data.thread_root`, what `thread new` posts) → a thread rooted at itself. This is the one case causality can't express: a freshly-opened thread has neither a parent to hang off nor a child to root it.
3. Otherwise → inherit the first parent's thread, or open one *rooted at that parent*. A thread is born the moment someone replies, and its id is the root message's id.
4. Otherwise → room-level, no thread.

Everything is additive. With no thread named, `await`, `respond`, `room send`, and `room messages` behave exactly as before.

## Changes

**Backend**
- `app/services/threads.py` — the derivation: an incremental `ThreadIndex` (a long-poll keeps one warm rather than re-deriving the room per tick), thread summaries, and resolution by exact id, id prefix, or episode URN (ambiguous prefixes are reported, never guessed).
- `app/routes/threads.py` — `GET /rooms/{room}/threads`, `GET /rooms/{room}/threads/{id}`, `POST /rooms/{room}/threads` (opens one by posting its root message, so `@`-mentions in it wake agents into the thread they just opened).
- Messages: a `thread` field on every message, a `?thread=` filter mirroring the existing `?episode=`, and `thread`/`reply_to` on POST.
- `await`: optional `?thread=` narrowing. A scoped poll drains **a cursor of its own** (`handle#thread:<id>`, seeded from the handle's room-level position), so working one conversation never consumes the turns waiting in another. Every turn — scoped or not — now names its `thread`, so one resident loop can hold several conversations apart.
- `reply`: `thread`/`reply_to` place a reply in a scope. A named *chat* thread clears the woken tick's episode, or rule 1 would silently pull the reply back into the negotiation.
- The persister keeps a warm index alongside the delivery log, so bus frames and live list-store rows carry their thread in O(1).

**CLI** — `mycelium thread ls|show|new`, plus `--thread` on `await` / `respond` / `room send` / `room messages` and `--reply-to` on `respond` / `room send`. `--thread` is one query/body param on the same two stateless calls: no new client state.

**Frontend** — messages carry their thread; a chip on a message follows its conversation and a banner returns to the whole room.

**Docs** — a Threads concept page, a short section in the claude_code SKILL.md, README and CLAUDE.md notes, and the regenerated `openapi.json`.

## Testing

- [x] Unit tests pass (`uv run pytest tests/ -x -q`) — backend 692 passed / 9 skipped, CLI 487 passed, frontend 184 passed

`ty check` is clean on both backend and CLI (the two remaining CLI diagnostics are the pre-existing unresolved `negmas` imports — it's an optional extra that isn't installed here, which is also why `test_engine_mediator.py` / `test_engine_runtime.py` don't collect in this environment).

New coverage:
- `fastapi-backend/tests/test_threads.py` — derivation (reply chains, parallel exchanges staying apart, episodes as typed threads, `…:live` scoping nothing, an opener threading itself, presence traffic never opening one, incremental assignment), summaries, prefix/URN resolution, the read path stamping each message, a live row keeping the thread the transcript gives it, and the API's filter / 404 / 409 behaviour.
- `fastapi-backend/tests/test_participate_threads.py` — a scoped `await` serving only its thread, a thread drain leaving the room backlog alone, a room-level turn naming its thread, and the four reply-scoping cases (including that an unscoped reply still answers its tick, unchanged).
- `mycelium-cli/tests/test_thread_cli.py`, `mycelium-frontend/src/components/event-stream.threads.test.tsx`.

Not verified live against a running SLIM node — everything here is node-free by construction (threading is a projection of records the wire already carries), but the end-to-end aligner path with threads on hasn't been exercised on a real stack.

One note on the diff size: regenerating the docs site also picked up **pre-existing drift** between the committed HTML and its markdown sources (the hub-and-spoke guide and troubleshooting sections had moved on). That churn is in `docs/reference.html` / `docs/llms-full.txt` and isn't from this change.

## Related Issues

Closes #631

---
_Generated by [Claude Code](https://claude.ai/code/session_01C6nWpUAqc1AcjwMtvkpxff)_


juliarvalenti:
Heads up: this PR and #692 (keyset pagination + reverse-scroll) both touch `fastapi-backend/app/services/persister.py::conversational_messages()`, and not compatibly.

Test-merged the two branches in a scratch worktree to check (not merged anywhere real). Most of the overlap is fine:
- `fastapi-backend/app/routes/messages.py` — only conflicts on an import line; `list_messages`'s actual body auto-merges cleanly, and the composition is correct (the `thread` filter runs on the candidate list *before* `pagination.paginate()`, so a thread-scoped view still gets stable cursors).
- `mycelium-cli/src/mycelium/commands/room.py` — three conflicts, all mechanical: threading `--thread`/`thread=thread or UNSET` through #692's cursor-walking `fetch()` closure, same shape as the route.
- `docs/reference.html` — generated, trivial.

The real problem is `persister.py`. Both PRs independently rewrote `conversational_messages()`'s caching:
- This PR: simple `(stamp, messages)` tuple cache, full re-parse on every change, computes `threads.thread_map(records)` over the whole transcript and attaches `thread=` per message.
- #692: a proper incremental cache (`_ConversationCache` with `stamp`/`messages`/`consumed`/`seal`), `_extended()` only parses the bytes appended since the last read, via a new `_project()` helper — and that helper calls `stored_message_from_record(room, r)` with no `thread=` at all.

These aren't reconcilable by keeping both blocks. `threads.thread_map()`'s rule 3 (a record inherits its **first parent's** thread) needs to walk the causal DAG back through however much of the transcript that takes — potentially past what's in the newly-appended slice #692's `_extended()` looks at. And `_ConversationCache` only keeps already-projected `StoredMessage`s, not the raw `TranscriptRecord`s a full `thread_map` pass would need — so by the time you're looking at just the new lines, the parent records needed to resolve ancestry may already be gone from memory.

Whichever of these two merges first, the other will need to either:
1. Extend `_ConversationCache` with a growing `id → thread` map, so `_extended()` can resolve a new record's thread from its parent's already-known thread instead of needing the full record list — real design work, not a merge resolution, or
2. Keep #692's incremental byte-parsing for building `StoredMessage`s, but always run `threads.thread_map()` over the full `load_transcript(room)` regardless of the cache — simpler, but gives up part of #692's performance win on rooms with active threads.

Flagging before either of you finishes so it's not a surprise at merge time. Cross-posted to #692.

juliarvalenti:
Retiring this rather than merging it.

The problem #631 named is real — a flat broadcast transcript tangles past two participants, and there is no unit smaller than the room to follow or address. What changed is where the answer lives.

**A conversation attaches to the thing it is about.** An episode, an issue, a work item — the rows the board already projects. Those are the scopes the room actually has, each already carrying an id, a lifecycle, participants and provenance. A thread is a *second* grouping laid over the same transcript, so a room would end up with two answers to "what is this conversation about": the row it concerns, and the thread it happens to hang off. Rows win, because they are the ones with work attached.

That is the same argument as #825, which retires `plan/` for being a second store the board had to special-case. Threads would be the same shape on the conversation side.

Nothing in the implementation is wrong — the derivation off L9 `message.parents` is the right mechanism, "an episode is a typed thread" is the right observation, and the per-thread `await` cursor is a genuinely good piece of design. It is the framing that is being retired, not the engineering. If threading-by-row needs a derivation later, the rules in `threads.py` are where to start reading; the branch stays.

The PR is also un-mergeable against main now and would need a rebase across the board work regardless.

---

Issue #719, opened 2026-08-21
Epic: A2A bridge — let any agent that speaks A2A join a Mycelium room


juliarvalenti:
## The art of the possible

Today a Mycelium room coordinates the agents we ship an adapter for. A resident `claude_code` loop joins, the aligner brokers, consensus compiles into a plan. That is a closed set: to bring a new kind of agent into a room, we write an adapter.

A2A (Agent2Agent, now a Linux Foundation protocol with 150+ organizations behind it) is the open lingua franca for agent-to-agent work. If Mycelium speaks it, the closed set opens.

**Inbound to a room.** A human opens a room, types `@`, and among the local agents is a research agent that lives in another company's stack. They summon the aligner. The aligner @-addresses that remote agent over A2A exactly as it addresses any member, it takes a position, the negotiation converges, the plan compiles. Nobody in the room ever had to know one of the negotiators was across the internet.

**Outbound from a room.** Someone else's orchestrator discovers your room's Agent Card, sends it a task ("get your team to consensus on the rollout date"), the room runs its real negotiation internally, and returns the compiled plan as an A2A artifact. Your room *is* an agent on the fabric.

That is the coordination-tax story made concrete: Mycelium stops being an island of its own adapters and becomes a node in the open interop fabric (AGNTCY / IoC), able to coordinate across a heterogeneous field of agents it never had to integrate one at a time.

## Why this is cheap (the load-bearing insight)

Mycelium already has the exact turn primitive A2A needs. A resident agent runs `mycelium await --loop --exec <cmd>`: long-poll for a tick, reason, respond. A bridged A2A agent is that same loop where `--exec` is "forward the tick to a remote A2A endpoint and hand its reply back."

So the aligner never learns the member is remote. It @-addresses the handle over SLIM like any other member, and a backend-held seat answers. We are not adding a coordination path, we are adding a new runtime behind an existing seat. The mapping is near-isomorphic:

| A2A | Mycelium |
| --- | --- |
| task lifecycle, `input-required` | a tick awaiting a reply (`await` returns) |
| `message/send` reply | `respond` |
| Agent Card `skills` | the room's `skills/` namespace |
| message `parts` | L9 exchange payload |
| card `securitySchemes` | the `@handle` + OIDC/JWT identity work |

## Honest scope boundary

[...]

Issue #718, opened 2026-08-21
End to end: negotiate to consensus with a remote A2A member (e2e skill + docs)


juliarvalenti:
Prove the whole thing and document it.

- an e2e smoke test in the spirit of the `claude-code-e2e` skill: register a remote A2A member (echo for wire, a local a2a-samples agent for behavior), summon the aligner, run a negotiation to consensus with the remote member participating, confirm the plan compiles
- a docs page under the interop / engines area explaining the bridge, both directions, and the honest boundary; nest it where the adapter / engine docs live

Done when: the e2e passes against a live sample agent and the docs page ships (docs generation clean).

juliarvalenti:
Part of the A2A bridge epic #719.

Issue #713, opened 2026-08-21
Register an external A2A agent as a room member (agent create --kind a2a --card)


juliarvalenti:
Make an external A2A endpoint a first-class room member.

- `mycelium agent create <handle> --kind a2a --card <url>` (plus optional auth); store the card URL + auth material in the `agents/<handle>` memory frontmatter
- resolve + validate the card on registration, cache it, extract advertised skills, fail clearly on an unreachable or incompatible card
- the member shows in the room roster / members panel and in the composer `@` autocomplete like any agent

Done when: a registered A2A agent appears as a member in the CLI and GUI, backed by a resolved Agent Card, with a clean error path for a bad card.

juliarvalenti:
Part of the A2A bridge epic #719.

Issue #712, opened 2026-08-21
Spike: prove the a2a-sdk client against a live A2A agent, pin the protocol version


juliarvalenti:
De-risk the client leg before we build anything into the backend. Stand up a throwaway script (not wired into the app) using `a2a-sdk` that:

- fetches an Agent Card, probing both `/.well-known/agent-card.json` and the older `/.well-known/agent.json` (deployed agents lag the spec rename)
- runs `message/send` and `message/stream` and reads the returned parts
- runs against two targets: the live public echo `hello-world-gxfr.onrender.com` (wire only), and a locally-run `a2aproject/a2a-samples` agent (a real reasoning reply)

Decide and write down: the SDK, the pinned protocol/SDK version, how we detect version/path drift, and how streaming maps onto our long-poll turn.

Done when: a short spike note + script shows a successful card fetch, send, and stream against both targets, and we have a pinned dependency to `uv add`.

juliarvalenti:
Part of the A2A bridge epic #719.

juliarvalenti:
Done via #720 (spike PR, closed unmerged). Thesis confirmed: the a2a-sdk client leg is ~30 lines and maps onto a per-tick server seat. Three drift findings to bake into #713/#714: pin a2a-sdk==1.1.2, probe both well-known card paths (new then old), and always set `ClientConfig.accepted_output_modes` or older servers 400 the send. Public echo is wire-only; #718 needs a local a2a-samples partner.

PR #720, opened 2026-08-21
Spike: prove the a2a-sdk client leg against a live A2A agent (#712)

juliarvalenti:
Spike for #712 (epic #719). A throwaway `a2a-sdk` client script run against the live public echo agent, answering: is the SDK good enough to be the bridge's client leg, and does "the bridge is just an await/respond exec loop" survive contact with a real remote agent?

**Not for merge.** The deliverable is the findings below; this PR gets closed, not landed.

## Result: thesis holds

`fastapi-backend/spikes/a2a_client_spike.py` resolves a remote Agent Card, sends one message, and drains the reply stream in ~30 lines of real logic. No bespoke JSON-RPC. The client leg is: resolve card -> `send_message` -> read parts. That maps straight onto a server-side seat calling out on each tick, exactly the exec-loop shape.

```
[1] probing well-known card paths
  GET /.well-known/agent-card.json -> 404
  GET /.well-known/agent.json -> 200 OK
[2] resolving Agent Card
  name: Hello World Agent  version: 1.0.0  streaming: True  skills: hello_world
[3] send one message + drain the stream
  <- message: 'Hello World'
```

## Findings to carry into the build

1. **Pin `a2a-sdk==1.1.2`.** Current line. Its client is proto-typed (`SendMessageRequest`, `StreamResponse`, `a2a.types` are `a2a_pb2` messages), not the old dict API. `ClientFactory(ClientConfig(...)).create(card)` is the entry point.

2. **Probe both well-known paths.** The SDK defaults to the new `/.well-known/agent-card.json`; the live agent (and anything not freshly updated) serves the old `/.well-known/agent.json`. The new path 404s. The bridge's card resolver must try new then old.

3. **Version drift bites on the request body, not just the path.** Against a v1.0.0 server the 1.1.2 SDK fails with `Request payload validation error` unless `ClientConfig(accepted_output_modes=[...])` is set. Empty (the default) makes the SDK omit `configuration.acceptedOutputModes`, which older servers treat as required and reject with -32600. Always set `accepted_output_modes`. Once set, both streaming and blocking sends work; the earlier streaming failure was this same root cause, not a streaming-specific incompatibility.

## Implication for the next issues

- The public echo is good for the **wire** test only (it ignores input, returns "Hello World"). For a real negotiation partner (#718) we need a locally-run `a2a-samples` agent on the same SDK generation, which also sidesteps the v1.0.0 drift.
- #713/#714 should build the card resolver with dual-path probing and a pinned `ClientConfig` (accepted_output_modes set) as non-negotiable defaults.

### DIFF
--- a/fastapi-backend/spikes/a2a_client_spike.py
+++ b/fastapi-backend/spikes/a2a_client_spike.py
@@ -1,0 +1,112 @@
+#!/usr/bin/env -S uv run --quiet --script
+# /// script
+# requires-python = ">=3.11"
+# dependencies = ["a2a-sdk==1.1.2", "httpx"]
+# ///
+"""Throwaway spike for issue #712 (epic #719).
+
+Question it answers: is `a2a-sdk` good enough to be the client leg of the
+Mycelium A2A bridge, and does the "it's just an await/respond exec loop" thesis
+survive contact with a real remote agent?
+
+What it does, against a live A2A agent:
+  1. probes BOTH well-known card paths (new /.well-known/agent-card.json and the
+     older /.well-known/agent.json) so we know how bad the spec-rename drift is
+  2. resolves the Agent Card and prints its skills + capabilities
+  3. sends one message via the high-level client and drains the response stream,
+     printing every text part it gets back
+
+Default target is the public echo agent. Pass a base URL to hit another.
+
+    uv run fastapi-backend/spikes/a2a_client_spike.py
+    uv run fastapi-backend/spikes/a2a_client_spike.py http://localhost:9999
+"""
+
+from __future__ import annotations
+
+import asyncio
+import sys
+
+import httpx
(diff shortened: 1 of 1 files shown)


juliarvalenti:
Spike complete, closing unmerged (the value is the findings, not the code).

**Thesis holds:** the client leg is resolve-card -> send -> read-parts, which drops straight onto a server-side seat that calls out per tick. Proven end to end against the live echo agent.

**Carry into #713/#714:**
- pin `a2a-sdk==1.1.2` (proto-typed client)
- card resolver must probe new `/.well-known/agent-card.json` then old `/.well-known/agent.json`
- `ClientConfig` must set `accepted_output_modes` or older servers reject the send

**For #718:** the public echo is wire-only; a real negotiation partner needs a locally-run `a2a-samples` agent on the same SDK generation.

Full detail in the PR description above.

Issue #726, opened 2026-08-21
Decide: carry the A2A bridge's outbound hop over SLIM (SLIMRPC) vs HTTPS


juliarvalenti:
Spun out of the A2A bridge epic #719 (the #717 discussion). The bridge's outbound send currently uses the a2a-sdk **HTTP/JSON-RPC** client (plain HTTPS to the remote agent). AGNTCY ships `slima2a` (agntcy/slim-a2a-python), a **SLIMRPC transport** that carries A2A over SLIM with SLIM identity + encryption instead of plain HTTPS — the AGNTCY-native path.

**Decision:** should the bridge's outbound hop ride SLIM (SLIMRPC) instead of / in addition to HTTPS?

**Spike findings (dependency + API de-risk done; PR closed unmerged):**
- `slima2a 0.7.0` **coexists with our pinned `slim-bindings 2.1.0`** (its constraint is `~=2.0`).
- BUT it **pins `a2a-sdk==1.1.0` exactly**, while the bridge is on 1.1.2. Adopting SLIM means pinning the whole bridge to `a2a-sdk==1.1.0`. All our client/server/types symbols exist on 1.1.0 (verified), so it's viable — the cost is the fragile exact pin (blocks future a2a-sdk bumps while slima2a holds it).
- **SLIMRPC is point-to-point RPC to a SLIM identity, NOT room-group MLS membership.** It hardens the *hop* (SLIM identity + encryption vs plain HTTPS); it does **not** make the A2A agent a member of the room's MLS group channel. The honest boundary in #717 stands either way.
- **Not yet proven:** a live SLIMRPC round-trip over a running SLIM node (server `SRPCHandler` + client transport). That's the next spike step if we pursue this.

**Tradeoff:** SLIM-native hop (aligns with SLIM-native architecture + the identity tiers, no plain-HTTPS egress) vs. coupling the bridge to `a2a-sdk==1.1.0` via slima2a's exact pin, plus a second transport code path + config to choose.

Reproduce: `uv run fastapi-backend/spikes/a2a_over_slim_spike.py` (on the closed spike branch).

juliarvalenti:
[...]

Issue #717, opened 2026-08-21
Authorize bridged agents and document the not-an-MLS-member boundary


juliarvalenti:
Make the bridge gated and honest about what it is.

- outbound: honor the card's `securitySchemes` and carry credentials to the remote agent
- inbound: sit behind the existing HTTP-API JWT gate (`app/services/auth.py`), off by default like everything else
- slot `@handle` in as the identity spine across the bridge
- document, in code comments and on the docs site, that a bridged A2A agent is proxied by the backend seat and is NOT an MLS group member and NOT E2E from the hub, the same honesty as the custodial scope note

Done when: both legs authenticate correctly, the inbound server is gated by default, and the scope boundary is written where a reader will actually see it.

juliarvalenti:
Part of the A2A bridge epic #719.

PR #727, opened 2026-08-21
Spike: can A2A ride SLIM natively in our stack? (epic #719)

juliarvalenti:
Throwaway spike for the #717 discussion / new issue #726. **Not for merge** — the deliverable is the findings, mirrored to #726.

Question: AGNTCY's `slima2a` (agntcy/slim-a2a-python) is a SLIMRPC transport for the a2a-python SDK, so an A2A hop can ride SLIM instead of plain HTTPS. Does it coexist with our pinned `slim-bindings 2.1.x`?

**Findings:**
- `slima2a 0.7.0` installs alongside `slim-bindings 2.1.0` (constraint `~=2.0`).
- It pins `a2a-sdk==1.1.0` **exactly** (we're on 1.1.2). All our bridge symbols exist on 1.1.0, so it's viable but couples us to that exact pin.
- SLIMRPC is point-to-point RPC to a SLIM identity, **not** room-group MLS membership — hardens the hop, doesn't make the agent a room member. The #717 boundary holds.
- Not proven: a live SLIMRPC round-trip over a SLIM node.

Decision tracked in #726. Reproduce: `uv run fastapi-backend/spikes/a2a_over_slim_spike.py`.

### DIFF
--- a/fastapi-backend/app/main.py
+++ b/fastapi-backend/app/main.py
@@ -29,6 +29,7 @@ os.environ.setdefault("HF_HUB_OFFLINE", "1")
 from fastapi import Depends, FastAPI
 from fastapi.middleware.cors import CORSMiddleware
 
+from app.routes.a2a_agents import router as a2a_agents_router
 from app.routes.agents import router as agents_router
 from app.routes.engines import router as engines_router
 from app.routes.episodes import router as episodes_router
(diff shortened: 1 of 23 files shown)


juliarvalenti:
Closing unmerged — spike deliverable is the findings (in the PR body + issue #726). Dependency/API risk retired: slima2a coexists with slim-bindings 2.1.0 but forces a2a-sdk==1.1.0. HTTP-vs-SLIM decision + the live round-trip step tracked in #726.

PR #751, opened 2026-08-22
Survey agent harnesses and grade them against the participation contract

juliarvalenti:
## Summary

Mycelium's resident model asks one thing of an agent runtime: hold a reasoning turn. `mycelium await --loop --exec <cmd>` loops await → reason → respond, and `<cmd>` is the harness. Everything else an integration does — dropping coordination instructions, building a manifest, reporting health — is in service of that one call.

This surveys the runtimes that clear that bar (Kiro, Codex, Cursor, Claude, Perplexity, Copilot, Antigravity), and builds the harness that *proves* a binary clears it rather than asserting it in prose. Full writeup: [`docs/agent-harnesses.md`](https://github.com/mycelium-io/mycelium/blob/claude/agent-harnesses-mycelium-oj3xl3/docs/agent-harnesses.md).

**Two findings shape the design, and are asserted in tests rather than argued:**

1. **Every harness reads `AGENTS.md` and takes a prompt on argv.** The differences that remain are enumerable: the headless argv, one optional harness-native rules file, the auth probe, and workspace-vs-host install. That is a table, not six programs — so the right shape is **one parameterized AGENTS.md family**, not six hand-written adapters. Claude Code is the genuine outlier (per-host skill under `~/.claude/`) and already has its own integration.

2. **Perplexity is not a harness.** `pplx` is a single-binary client for the Search API (`pplx search web`, `pplx content fetch`, JSON out). It answers questions about the world; it does not hold a reasoning turn, call tools, or edit files. It's a *tool a harness calls* — it belongs behind MCP for the aligner and residents to reach, not in front of an adapter. Recorded in the table as `unsupported` so it isn't rediscovered.

**Recommended first build: Codex.** `AGENTS.md` is its native instruction surface (so it needs no harness-specific rules file at all), `codex exec --json` gives a JSONL event stream, and `codex exec resume` maps onto the durable per-agent session `await --loop` wants — which most of the field lacks. Then Kiro (`.kiro/steering/**/*.md` is a near-exact analogue of Cursor's project rules, so the cursor integration is the template), then Copilot gated on pinning a version. Antigravity waits on an unattended-auth story.

## Changes

- **`mycelium-cli/src/mycelium/integrations/harness_specs.py`** — the family table as data: binary, headless argv, auth probe, context-file conventions, and an honest capability grade (`proven` / `untested` / `candidate` / `unsupported`). Two consumers read it — an integration builds an adapter from a spec, the conformance runner points a binary at a room from the same spec — so a family can't be "supported" by code the runner has never exercised, or exercised by a runner reading stale flags.

- **`scripts/harness_conformance.py`** — grades a family on a ladder (`binary` → `auth` → `oneshot` → `participate`) and stops at the first rung it can't climb. Three deliberate properties:
  - **A SKIP is never a FAIL.** A runner with no Kiro binary has learned nothing about Kiro; reporting that red trains everyone to ignore the report.
  - **`auth` is advisory and doesn't gate the turn.** Harnesses authenticate in ways no probe can enumerate — this was found empirically: `claude` on the dev machine was fully authenticated with no `ANTHROPIC_API_KEY` and no `~/.claude/.credentials.json`. The turn is attempted anyway and a failure is annotated as probably-unauthenticated.
  - **`participate` needs a manifest, so it reports `needs-adapter`** for a family with no integration — precisely the gap an adapter closes — while `oneshot` still grades. That de-risks a candidate before a line of adapter code is written, which is the point of building the runner first.
  - It also ships the generic, harness-agnostic exec handler (turn in → argv rendered → stdout back via `mycelium respond`); an adapter would ship the same script and vary only `HARNESS_ARGV`.

- **`.github/workflows/harness-conformance.yml`** — weekly, on dispatch, and on PRs touching the harness surface. `oneshot` job: matrix over every family, installs each vendor CLI, no hub. `participate` job: adapter-backed families only, stands up a SLIM node + hub first. Nothing gates a PR — vendors ship breaking CLI changes on their own schedule (GitHub removed `copilot --headless --stdio` with no deprecation, github/copilot-cli#1606), so a red run is news, not a broken pull request.

- **`mycelium-cli/tests/test_harness_specs.py`** — guards drift between the table's claims and the code: a family graded `proven`/`untested` with no integration behind it, a `candidate` that has since gained one, an argv template that can't render, a prompt that doesn't survive as exactly one argv element, or a harness with no auth story written down.

- **`docs/agent-harnesses.md`** — the survey, the per-harness findings, the adapter proposal, and the build order.

## Testing

- [x] Unit tests pass — `525 passed, 3 skipped` (`mycelium-cli`), including 39 new
- [x] Linting passes (`uv run ruff check`)
- [x] Format check passes (`uv run ruff format --check`)
- [x] Type check passes (`uv run ty check .`)
- [x] Docs generator run — no drift
- [x] **Runner validated against a real harness binary**: graded `claude` 2.1.239 to the `oneshot` rung (binary ✓, oneshot ✓ — nonce echoed in 5.8s), and correctly reported skips-with-reasons for the six binaries not on the machine.
- [x] `participate` setup validated step-by-step against a live hub (`room create`, `agent create --adapter claude_code`, `room send`, `await --loop --exec`, `--json room messages`).

**Left as draft for one reason:** the `participate` rung has not been observed green end-to-end. The dev container had no Docker daemon, so no SLIM node could be started — and a room *is* a SLIM group channel, so with no node the hub can't provision one and `await` never serves a turn. That was found the slow way (a silent 7-minute timeout) and is now a pre-flight check that skips immediately with an actionable reason. The `participate` CI job stands up the node exactly as `ci.yml`'s `integration-slim` job does, but that job's first real run is the proof, and it needs `ANTHROPIC_API_KEY` in repo secrets to get past `oneshot`.

Secrets the workflow reads (all optional; an absent key skips that family with a reason rather than failing): `ANTHROPIC_API_KEY`, `CURSOR_API_KEY`, `OPENAI_API_KEY`, `COPILOT_GITHUB_TOKEN`, `KIRO_API_KEY`.

## Related Issues

Relates to the untested `cursor` adapter — this is the harness that would move it from `untested` to `proven`.

---
_Generated by [Claude Code](https://claude.ai/code/session_019gUfbJ3J8AzwKnWSUgS87w)_

### DIFF
--- a/.github/workflows/harness-conformance.yml
+++ b/.github/workflows/harness-conformance.yml
@@ -1,0 +1,250 @@
+# Point real agent-harness binaries at the participation contract and report what
+# happened. Mycelium claims a harness works when a resident loop has actually
+# answered a mention through it, and this is where that claim is earned — see
+# `scripts/harness_conformance.py` for the rungs and `mycelium/integrations/
+# harness_specs.py` for the family table both this workflow and the integrations
+# read.
+#
+# Nothing here gates a PR. Harness vendors ship breaking CLI changes on their own
+# schedule (GitHub removed `copilot --headless --stdio` with no deprecation), so a
+# red run means "a harness moved", which is news, not a broken pull request. The
+# job to watch is the report in the step summary.
+name: Harness conformance
+
+on:
+  schedule:
+    # Weekly, Monday 06:00 UTC: often enough to catch a vendor's breaking release
+    # within a sprint, rare enough not to burn API credit on every push.
+    - cron: "0 6 * * 1"
+  workflow_dispatch:
+    inputs:
+      families:
+        description: "Space-separated families to grade (default: all)"
+        required: false
+        default: ""
+  pull_request:
+    # Only when the harness surface itself changes: the spec table, the runner, or
+    # an integration. A frontend PR has nothing to learn from a Kiro binary.
+    paths:
+      - "mycelium-cli/src/mycelium/integrations/**"
+      - "scripts/harness_conformance.py"
+      - ".github/workflows/harness-conformance.yml"
+
+permissions:
+  contents: read
+
+jobs:
+  # ── rungs 1–3: does the binary reason? ──────────────────────────────────────
+  #
+  # No hub, no SLIM node, no adapter needed — which is the point. A candidate
+  # family with no integration still grades to `oneshot` here, so the harness is
+  # de-risked before any adapter code is written.
+  oneshot:
+    name: "oneshot: ${{ matrix.family }}"
+    runs-on: ubuntu-latest
+    # A missing binary or credential is a SKIP, not a failure; a genuine FAIL is
+    # reported without `--strict` so one broken vendor CLI doesn't mask the rest.
+    continue-on-error: true
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - family: claude_code
+            install: npm install -g @anthropic-ai/claude-code
+          - family: cursor
+            install: curl https://cursor.com/install -fsS | bash
+          - family: codex
+            install: npm install -g @openai/codex
+          - family: copilot
+            install: npm install -g @github/copilot
+          - family: kiro
+            install: curl -fsSL https://cli.kiro.dev/install | bash
+          # antigravity has no API-key env var: headless mode reuses a cached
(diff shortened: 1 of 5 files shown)


juliarvalenti:
Honestly the herdr integration just gives us this for free

Issue #600, opened 2026-08-16
Epic: interlinked memory & knowledge ontology (myc:// links)


juliarvalenti:
## Goal
Turn a room's memory into an **interlinked wiki / knowledge ontology** — leaning all the way into memory-as-markdown. A first-class link scheme (`myc://`), a **curator engine** that maintains it, **user-configurable frontmatter schemas** with **Notion-style views**, and the retrieval layer that turns it all into a **cost lever** — offloading cognition to cheaper / specialized models via curated context.

Mycelium already has the grain: rooms are folders, memories are markdown + YAML frontmatter (`.mycelium/rooms/{room}/{key}.md`), a per-room fastembed index, cognitive engines (aligner/synthesizer), and a `[[wikilink]]` convention in the auto-memory system. This epic names and hardens it.

## Tracks
- [ ] #594 — myc:// memory linking (backlinks + transclusion)
- [ ] #595 — Curator engine
- [ ] #596 — Configurable frontmatter schemas per directory
- [ ] #597 — Notion-style typed views over a namespace
- [ ] #598 — Curated-context retrieval (offload to cheaper/specialized models — the *why*)
- [ ] #599 — Frontend wiki/graph UI
- [ ] #614 — Full-page memory view (dedicated route)
- [ ] #618 — Chat: reference memories with autocomplete (myc://)

## Principles
- **DRY for knowledge** — one canonical leaf node per fact; everything else links. Update the leaf → the graph stays true. Backlinks make the blast radius visible.
- **Markdown is the data source** — no DB. Frontmatter is soft, typed, user-configurable data.
- **The engine is the curator** — maintenance is a first-party cognitive engine, faithful/anti-theatre like the aligner.
- **Cost lever** — curated, interlinked context lets small/specialized models do more.
- **Backward compatible** — unschema'd, unlinked memories behave exactly as today.

## Grounding
- Storage: `.mycelium/rooms/{room}/{key}.md`; fastembed JSONL index; `memory set` upserts; `mycelium reindex`.
- Engines: `engine create --kind {aligner|synthesizer}` → add `curator`.
- Frontend: `memory-panel.tsx`, `memory-detail.tsx`, `markdown-content.tsx`, `room-inspector.tsx` (Memory tab).

Issue #597, opened 2026-08-16
Notion-style typed views over a namespace (table/board/list)


juliarvalenti:
## Framing
Render a directory as a **typed view** — table / board / list — from its frontmatter schema. The "Notion database from nothing": configure a schema (sibling issue), then see + edit the namespace as structured data (e.g. an issue tracker). Part of the interlinked-memory epic.

## Scope
- **Views over a namespace**: table (columns = schema fields), board/kanban (group by an enum field), list. Filter / sort / group by frontmatter fields.
- **Inline edit** frontmatter from the view (writes back to the markdown).
- Backed entirely by markdown + frontmatter (**no DB**) — the view is a projection.
- Ships in the room's Memory area (`memory-panel.tsx` / room-inspector Memory tab).
- Example: a custom `issues/` namespace with `status`, `assignee` (a `@handle` link), `priority` → a kanban.

## Open questions
- Which view types for v1 (table first?).
- Saved-view config location + sharing across the room.

juliarvalenti:
Part of the interlinked-memory epic #600.

Issue #504, opened 2026-08-14
Frontend has no favicon


juliarvalenti:
The Next.js app ships **no favicon / app icon** — browsers fall back to the default blank/globe. Confirmed:

- No `src/app/favicon.ico`, `src/app/icon.*`, `src/app/apple-icon.*`, or `public/*.ico`.
- No `icons` entry in the `metadata` export in `src/app/layout.tsx`.

## Fix
Add an app icon via the App Router convention:
- `src/app/icon.svg` (or `icon.png`) + `src/app/apple-icon.png`, and/or a `favicon.ico`. Next.js auto-wires these into `<head>` — no manual `<link>` needed. Optionally set `metadata.icons` explicitly.

## Design
Match the brand: the mycelium mark, dark design tokens, works at 16–32px. A simple monogram/glyph (the fungal/network motif) reads better than text at favicon size. Provide light + dark if we want it to hold up in both browser themes.

PR #752, opened 2026-08-22
The coordination board: what needs you, who is holding it, and what the tools say (#493 + #597)

juliarvalenti:
A room's plan screen showed a checklist. This makes it the surface where work is found, taken, held and handed off — and where the state of that work in GitHub, Jira or wherever it actually lives shows up on the row without anyone copying it across.

#493 (a "needs you" cockpit) and #597 (typed views over a namespace) turned out to be one mechanism, so they are built as one.

## A row is a title plus its frontmatter

The schema is *inferred* from that frontmatter across the collection, and every view — cockpit, kanban, table, timeline, daily — is a different config over one filter/sort/group pipeline. The coordination surface **is** the typed view, pivoted differently. Which also makes any namespace a tracker for free: write memories under `issues/` with `status` / `assignee` / `priority` and you can group them into a kanban, because the fields are in the markdown.

Nothing is stored twice. Rows are projected from the room's plan tasks, its episodes, the `decisions/` `status/` `work/` `failed/` namespaces, and who is actually resident. Every row carries its provenance, and the plan's prose files are still there as the Docs view.

### The cockpit — what needs a person

Cockpit

### The table is where #597 lands

Columns and their types are read out of the frontmatter. `status`, `custody`, `upstream` and `writer` are four different questions and four separate columns, and the `select` / `handle` chips are inferred, not declared. Group by any of them.

Typed table

### Holding work is a lease

An agent is resident rather than one-shot, but every session eventually ends and none of them get to announce it. So **every assertion an ephemeral actor makes about the future is a lease.** A claim held as a fact leaves the board asserting "@someone is on this" forever; held as a lease it drains, and the row returns to the pool.

The test the model exists to pass: **a board whose agents all died an hour ago reads empty.** There is a test named that.

In flight

Custody turned out to be the freshness model already built for external values, pointed at claims: `claimed_at` + `ttl` is `fetched_at` + `ttl` + `swr`, fresh/stale/expired is fresh/stale/missing, renewed by the agent's loop instead of by a refresh.

### The row shows what the tool says, in the tool's own words

Write a pull request into a plan task and the row carries its state: `CI failing +1 1m ago`, `changes requested 1h ago`. The hub reads the room's own text, asks each provider what it recognises, and answers from cache — a board opening never becomes a burst of calls to GitHub. Nothing in the app knows what `#504` means; a provider does.

A reference with no answer yet draws a skeleton rather than a state nobody reported. A stale answer keeps its value and says it is stale.

### Kanban, timeline, daily

<table>
<tr>
<td>Kanban</td>
<td>Timeline</td>
</tr>
<tr>
<td>Dragging a card writes that one field. Empty columns still render, so a board is somewhere to move work rather than a summary of work that exists.</td>
<td>What moved while you were away.</td>
</tr>
</table>

### It is a terminal surface too

```
mycelium board                          # what needs you
mycelium board --lens in-flight         # claimed work, who holds it, CI
mycelium board --lens all --view table  # the room as structured data
mycelium board --group custody          # group by any field it found
mycelium board claim t3                 # take it, as a lease
mycelium board release t3 --note "..."  # hand it back, with a reason
mycelium board log --last-week          # what the room did, by day and by who
mycelium board credential set GITHUB_TOKEN
```

## What is in here

- **Frontend** — `src/lib/board/` (item, fields, schema, projection, view, capture, custody, upstream, activity, earcons) and `src/components/board/`. The plan pane renders the board; `room-plan-header.tsx` is untouched and reachable as the Docs view.
- **CLI** — `mycelium board` mirroring the same projection rules, plus `claim` / `release` / `resolve`, `board log`, and `board credential set/ls/rm`.
- **Backend** — `app/services/status/` (the provider contract, auth schemes, credential store, cache, runtime, registry, discovery, a GitHub provider) with `GET /rooms/{room}/status`, and `app/services/leases.py` + `/leases/*` for custody.
- **Contract** — `contracts/board-vocabulary.json` freezes the statuses, kinds, priorities, lens map, verb keys, schema inference table, custody block and upstream vocabulary. Asserted from all three implementations, so none can drift silently.
- **Docs** — a Board concept page, a status-providers section in the architecture reference, and the overview positioning.
- **Sound** — `audio-ping.ts` grows a `motif()` primitive; `ping()` is written in terms of it.

## Testing

```
backend    899 passed, 13 skipped
CLI        670 passed,  2 skipped
frontend   663 passed
ruff check / format / ty / tsc / eslint   clean
docs regenerated, no drift; openapi.json refreshed
```

## What is still rough

- **`owner/repo#123` is always read as a pull request** (#816). When it is an issue, the row says "not visible to this token" — a wrong answer pointing at the wrong cause. The one thing here that is confidently incorrect rather than merely absent, and worth fixing before anyone relies on it.
- **The live GitHub hop has never executed.** Every layer around it is proven and real pull-request state maps correctly through the real provider, but the sandbox this was built in cannot reach `api.github.com`. One token on one hub closes it.
- **`block` prints rather than writes.** The lease endpoints cover custody; a general field write is a different seam.
- **Refresh is not bounded by attention.** The design says ambient refresh should cover rows in somebody's lens; the implementation refreshes every reference in the room.
- **Interest is not a lease yet.** `await --lease` polls rather than registering, so there is nothing to expire.
- **Not designed with anyone.** Density, the sounds and the lens split are still first guesses.

## Related

Closes nothing on its own. Implements #493 and #597 (part of the interlinked-memory epic #600), and lands #768, #778, #782, #783, #785, #789, #794, #804, #812. Follow-ups: #816, #807, #808, #811.


juliarvalenti:
## Responsiveness

Checked at 1440 / 1280 / 1100 / 1024 / 768. The board sits between two rails, so the pane is a lot narrower than the window: 1440 → 864px, 1280 → 704px, 1100 → 524px.

Two things were mine and are fixed in `5bc1b34`:

- **Viewport breakpoints answered the wrong question.** The mode switcher was hiding its labels on `sm:`, which is about the window, not the pane. The board root is now a container query, so the mode buttons drop to icons and "views & fields" drops to a glyph based on the space actually available. That keeps the header on one row from 1024 up, instead of wrapping to four rows.
- **The keyboard legend clipped** its last verbs rather than wrapping. Now wraps.

Board at 1024

No layout escapes at any width: the page never scrolls horizontally, and the table and kanban scroll inside their own containers (verified 1024/1280 across all five views).

**Below ~1024 it degrades, and that part isn't the board.** The workspace shell keeps both rails at full width and crushes the centre pane; the Channel view does exactly the same thing at 768, so it's pre-existing and I left it alone rather than widening this PR. If narrow screens matter, the fix belongs in the shell (collapse a rail under a breakpoint) and would benefit every pane at once.

---

juliarvalenti:
## Merged main — #753 fixes the part that wasn't mine

`2d1be3e` ("Fold the workspace rails away on a narrow window") is exactly the shell fix I said this PR shouldn't take on. Merged it in and re-measured.

**Before, the pane got crushed. Now the rails fold and it doesn't:**

| viewport | pane before | pane after |
|---|---|---|
| 1440 | 864px | 862px |
| 1280 | 704px | 702px |
| 1024 | 524px | 446px (rails folding) |
| 768 | ~380px, clipped | 483px |
| 480 | unusable | 384px |

Wide layouts are untouched; below ~1024 the board now gets a real pane instead of a squeezed one, down to phone width.

**Two of my controls needed to meet it there** (`b83321f`): the search field and its options toggle now wrap as a pair, instead of the toggle orphaning onto its own line at 480; and the keystroke legend hides on a pane too narrow to spare two lines for it (below ~34rem of pane, where there's no room and usually no keyboard).

Thresholds now, all container-driven rather than viewport-driven:

| pane width | mode labels | keystroke legend |
|---|---|---|
| ≥ 46rem (1440+) | shown | shown |
| 34–46rem (1280) | icons | shown |
| < 34rem (1024 and below) | icons | hidden |

Board at 1024

Board at 480

Page never scrolls horizontally at any width, table and kanban still scroll inside their own containers, and the suite is green at 396 (was 375 before the merge brought main's new tests).

---

juliarvalenti:
## Daily log

A sixth view, and a CLI half. The board answers *what needs me now*; this answers *what got done*.

It's the same projection idea, bucketed by calendar day and attributed to whoever moved it. Nothing is written to it: it reads the facts that already carry a time and a name — messages, memory writes and revisions, finished plan tasks, negotiations.

Daily log

**Agents, engines and people each get a lane**, with a fortnight sparkline of their own rhythm. An agent that spent Tuesday on a migration is as legible as the person who reviewed it.

**"What did we work on last week"** is a chip, not a reconstruction — This day / This week / Last week / 30 days, over whichever day you click in the calendar:

Last week

**Timezone is the reader's**, since a day boundary means nothing without one — remembered per browser in the GUI, `--tz` (defaulting to `$TZ`) on the command line. Day arithmetic is done on calendar dates rather than millisecond offsets, so a zone changing offset doesn't shift a day; there are tests for the DST week and for one instant landing on three different days in three zones.

**Gamified, carefully.** A ring shows how full today is against a modest target (6), with the current and longest streak beside a ten-week heat calendar. It counts what actually moved, it belongs to the room rather than to a person, and nothing reads it as a score. That felt like the honest version of "fill up the daily log" — the nudge is to keep the log worth reading, not to farm it.

### The same log in the terminal

This is the half an agent needs when it rejoins a room after a week away — rather than replaying the whole channel.

mycelium board log

```bash
mycelium board log                    # the last week
mycelium board log --last-week        # the week before
mycelium board log --day 2026-08-19   # one day
mycelium board log --by @agent-y      # one worker's lines
mycelium board log --tz Asia/Tokyo    # someone else's clock
```

### Two things this surfaced

- **A double-count.** A memory write and the `memory_changed` message announcing it are the same work; so are an episode and its `coordination_consensus`. The log now reads each fact from one source and skips the wire copy.
- **Raw JSON in the lines.** Coordination payloads were being printed as their envelopes. The log reads them by type instead: `negotiated round 4 · @risk accept`.

Docs updated (a Daily section on the Board page), contract grown with the calendar conventions both surfaces assert, and green across the board: **519 CLI tests, 416 frontend** (46 on the board engines), lint/format/`ty`/`tsc` clean.

---

PR #769, opened 2026-08-22
Document status providers in the user-facing docs

juliarvalenti:
> Targets **#768** so the diff is docs only. Merge into that branch, not `main`.

## Summary

#768 adds the status provider contract; nothing in the docs told a reader how to use it. The board doc already promises that a row pointing at a pull request shows its live state — *"who has it, which branch, whether CI is green"* — with nothing behind it and no account of how a reader makes that happen, or how they'd point Mycelium at a tracker that isn't GitHub.

Two user-facing sections, written where a reader would look for each: the *what you do* on the board page, the *what you configure and extend* in the reference. Both are written in the future tense under a banner, since none of it is wired to a row yet.

## Changes

**`concepts/board.md`** — a **Live status** subsection under "GitHub, by reference", banner-first:

- **Mentioning the pull request is the whole of it.** No per-row setting: the link goes where the work is already described (a plan task, a memory, a message). Both `owner/repo#123` and the pasted URL count; two rows pointing at the same PR share one lookup.
- **What a row will show.** GitHub's own wording on top (`CI failing`, `changes requested`, `draft`), the six-state vocabulary underneath as the thing a surface can sort, filter and colour by — including the `ok` / `done` pair a board most needs to keep apart, and how GitHub maps onto the six more narrowly than it looks.
- **How current it will be.** The age on the row, why a render never waits on GitHub, last-good surviving a failed lookup, and a value too old to be evidence dropping rather than being shown as current. Cost is covered by the two properties the runtime actually has — single-flight and bulk fetch.

**`reference/architecture.md`** — a **Status providers** section, placed after Adapters as its counterpart: adapters connect *agents*, providers connect *the tools the work already lives in*.

- What the GitHub provider recognises and reports, and why providers are hub-side: that's where the credential is, one cache is shared by the room, and a spoke never holds a service token.
- The credential half that's real — a provider declares `credential` and `base_url` and is handed neither value nor host — and the half that isn't: nothing yet fills the runtime's credentials mapping, so there is nothing to configure. The one durable caveat stays: don't hand-edit `~/.mycelium/.env`, since `mycelium config apply` regenerates it and a token written in by hand vanishes without an error.
- **Teaching Mycelium another tracker**: the two methods (`claims`, `fetch`), the declared attributes, what the runtime does so a provider is request-and-parse, and the two rules it enforces (bulk only; failure is per reference).

Regenerated: `docs/index.html`, `docs/reference.html`, `docs/search-index.js`, `docs/llms-full.txt`.

## Review rounds

Two rounds, both checking the docs against #768's code rather than its description. Every finding verified independently before acting on it — the module has no caller outside itself; `contracts/board-vocabulary.json` and the provider states are two closed vocabularies that both contain `blocked`; `github.py` reaches `ok` only via an `APPROVED` review and has no branch emitting `unknown`. All correct, all addressed.

- **`15cd9b6`** — tense. One banner over the live-status material, the "the hub can resolve them" claim removed, the lens-scoped refresh paragraph cut (it was an idea in #768's description, not code), the `--group status` example cut rather than corrected, the state table retitled to the vocabulary it describes, and the invented `GITHUB_TOKEN`-from-the-environment mechanism removed.
- **`670f263`** — merged the base (`4eac2ee`) in and matched the docs to it: `ok` is healthy-and-in-flight rather than a good ending, the provider template declares `base_url`, and it's the runtime that refuses a reference whose provider has no credential, without calling the provider at all.

All nine review threads are resolved. Regenerated after the merge, so the generated HTML carries the state-table change.

## Still not documented

The agent-side bargain — pass a freshness bound, get `missing` rather than a stale "CI green" — because `mycelium board` has no `--json` / `--max-age` flag to describe. It's the part an agent reader most needs; worth adding the moment the flag exists.

## Testing

Docs-only; no Python changed, so the backend gates are untouched rather than re-run.

- [x] Docs regenerate clean on this head — reran the generator after merging the base and `git status` shows no diff, which is the CI drift gate (`ci.yml`) this change is subject to
- [x] `mycelium docs board` and `mycelium docs architecture` render the new sections in the terminal reader
- [ ] Unit tests / lint / format — not run: no code in this diff

## Related Issues

Targets #768, which targets the board PoC #752.


mycelium-cli/src/mycelium/docs/concepts/board.md
```diff
@@ -30,7 +30,9 @@ Review 1
 
 > **Experimental.** The board reads real room state today. The triage verbs act
 > on your own view: `resolve` writes through to a plan task, and the rest change
-> what you see without yet writing back to the room.
+> what you see without yet writing back to the room. Live status for linked pull
+> requests is landing: the hub can resolve them, and the rows are being wired to
+> show it.
```
juliarvalenti: "the hub can resolve them" is stronger than the code supports. Nothing in the hub imports `app.services.status` — I checked the whole of `fastapi-backend/app`:

```
grep -rn "services.status" fastapi-backend/app --include=*.py | grep -v "^fastapi-backend/app/services/status/"
# (no matches)
```

There's no registry, no route, no scheduler, and no `Context` implementation, so the module is unreachable code today: the hub *contains* a resolver, it can't currently *run* one.

The architecture page states this exactly right — "the shape below is what a provider implements today" — so it's just this callout that overreaches. Suggest matching that wording: the contract exists in the backend, nothing is attached to a row yet.

---

mycelium-cli/src/mycelium/docs/concepts/board.md
```diff
@@ -146,6 +148,58 @@ is a link, it's a link and not a copy:
...
+
+Every status carries the moment it was fetched, and the row shows its age
+(`CI green · 4m`). The board never waits on GitHub to draw: it shows what it
+last knew and refreshes behind you. If a lookup fails, the last good state stays
+on the row rather than the row going blank; if it gets old enough to stop being
+evidence, it drops off instead of being shown as if it were current.
+
+Only rows in your **needs you** and **in flight** lenses are kept live. A pull
+request nobody is currently looking at costs nothing until someone looks, which
+is what keeps a large room from turning into a polling loop against GitHub.
```
juliarvalenti: This one I'd cut rather than soften — it exists in no code anywhere.

Lens-scoped refresh was an idea in **my** PR description for #768 ("the board is the subscription list"), not something #768 implements. There's no sweeper, and the status module has no concept of a lens: `runtime.due()` takes whatever refs you hand it and knows nothing about where they came from.

Documenting it now would make it the first thing a reader believes about the cost model, and it's the claim most likely to be wrong when the wiring actually lands — the sweeper might key off room membership, or recency, or something else entirely. Worth leaving unwritten until something decides it.

---

mycelium-cli/src/mycelium/docs/concepts/board.md
```diff
@@ -146,6 +148,58 @@ is a link, it's a link and not a copy:
...
+|---|---|
+| `ok` | approved, checks green |
+| `pending` | draft, checks running, or waiting on a first review |
+| `blocked` | changes requested — a person is the fix |
+| `failed` | checks failing — a machine is the fix |
+| `done` | merged or closed |
+| `unknown` | the tool said something Mycelium can't map |
+
+So `mycelium board --group status` puts everything blocked together whether it's
+a room decision nobody has answered or a pull request somebody has to revise.
```
juliarvalenti: This is wrong twice, and finding it is genuinely useful because the second half is a hole in **my** sketch rather than in your writing.

**There are two closed vocabularies that both contain `blocked`.** The board's own row status, frozen in `contracts/board-vocabulary.json`:

```
open · claimed · in_progress · in_review · blocked · resolved · dismissed
```

and the provider state from #768:

```
ok · pending · blocked · failed · done · unknown
```

So: a room decision nobody has answered is `open`, not `blocked` — the example inverts the board's own vocabulary. And `--group status` groups the row's `status` field, which is the first list; provider state isn't a row field at all, so it would never appear in that grouping.

The underlying gap is mine: #768 never specifies how a provider's state lands on a row without colliding with the row's own `status`. It probably wants its own field name (`pr_state`, or namespaced per provider) precisely so both can coexist and be grouped separately. I'll take that back to #768 — please don't try to document around it.

---

mycelium-cli/src/mycelium/docs/concepts/board.md
```diff
@@ -146,6 +148,58 @@ is a link, it's a link and not a copy:
...
+colour work from:
+
+| State | What it is for a pull request |
+|---|---|
+| `ok` | approved, checks green |
+| `pending` | draft, checks running, or waiting on a first review |
+| `blocked` | changes requested — a person is the fix |
+| `failed` | checks failing — a machine is the fix |
+| `done` | merged or closed |
+| `unknown` | the tool said something Mycelium can't map |
```
juliarvalenti: Two rows in this table don't match `providers/github.py`, and the table's heading ("What it is **for a pull request**") is what makes them wrong — they'd both be fine as descriptions of the vocabulary in general.

**`ok` — "approved, checks green"** reads as either/or. In the code, `ok` requires `reviewDecision == APPROVED`; green checks on their own fall through to the final branch and come back `pending` / "awaiting review":

```python
elif node.get("reviewDecision") == "APPROVED":
    state, label = "ok", "approved"
else:
    state, label = "pending", "awaiting review"
```

**`unknown`** — the GitHub provider can never emit it. All seven branches of `_status()` yield `ok`, `pending`, `blocked`, `failed` or `done`; there's no path to `unknown`. It's a real member of the vocabulary (and the right answer for a *future* provider that can't map something), but in this table it describes behaviour that can't occur.

Either retitle to make it the general vocabulary, or drop `unknown` and tighten `ok` to "approved".

---

mycelium-cli/src/mycelium/docs/reference/architecture.md
```diff
@@ -189,3 +189,75 @@ mycelium agent create design-agent --adapter cursor \
...
+cache is shared by everyone in the room rather than each client polling GitHub
+for itself. A spoke never holds a service token.
+
+### Giving the hub a credential
+
+The GitHub provider reads `GITHUB_TOKEN` from the backend's own environment; a
+read-only token is enough (add `repo` scope for private repositories). Set it
+wherever your deployment supplies environment to the backend service — note that
+`~/.mycelium/.env` is regenerated by `mycelium config apply`, so hand-edits there
+don't survive.
```
juliarvalenti: The mechanism here is invented, though not by you — you had nothing better to go on.

`providers/github.py` calls `ctx.secret("GITHUB_TOKEN")`, and `Context` in #768 is a **Protocol with no implementation**. Nothing anywhere reads the process environment, so "reads `GITHUB_TOKEN` from the backend's own environment" describes a path that doesn't exist yet; where the value comes from is genuinely undecided.

The `~/.mycelium/.env` warning is a good catch and worth keeping whatever we land on — `mycelium config apply` does regenerate that file, and someone hand-editing a token into it would lose it without explanation.

I'm fixing the underlying inconsistency on #768: the provider will declare `credential = "GITHUB_TOKEN"` and the runtime will resolve it and hand back a pre-authenticated transport, so the provider never sees the value. Once that lands this section can describe something real — I'll comment here when it does.

---

mycelium-cli/src/mycelium/docs/reference/architecture.md
```diff
@@ -189,3 +189,75 @@ mycelium agent create design-agent --adapter cursor \
...
+        of its own: PROJ-14 means a ticket because you said so."""
+
+    async def fetch(self, refs: list[Ref], ctx: Context) -> list[Outcome]:
+        """Resolve a batch. One Ok or Err per reference, in any order."""
+```
+
+What you don't write is as important as what you do. `ctx.http` arrives with the
+credential, timeout and retry policy already applied, so a provider is
+request-and-parse. Batching, de-duplication, caching, single-flight and
+rate-limit backoff belong to the runtime — a provider that reimplements them is
```
juliarvalenti: This paragraph is right, and it caught a contradiction in #768 that I'd missed.

You took "`ctx.http` arrives with the credential… already applied" from my `Context` docstring — but my own worked example doesn't do that. `providers/github.py` pulls the token out with `ctx.secret()` and sets the header by hand:

```python
token = ctx.secret("GITHUB_TOKEN")
...
headers={"Authorization": f"Bearer {token}"},
```

So the contract says one thing and the file people are told to copy does another — and the copy is the version that spreads. Since the example is the specification in practice, I'm changing the code to match your sentence rather than the sentence to match the code: `credential` becomes a declared attribute, the runtime builds the authenticated transport, `ctx.secret()` goes away, and a provider never handles the token at all. It also buys a property worth having — a provider can't exfiltrate a credential it's never handed.

Keep this paragraph as written; it'll be true shortly.

---

juliarvalenti (commented):
Reviewed against the code in #768 rather than against its description, since that's where the risk is when docs land ahead of wiring.

**Faithful on mechanism, not on tense.** Every account of *how the contract works* is accurate. Most accounts of *what happens today* describe code paths that don't exist, and one describes something that exists nowhere but my own PR text.

### Accurate, and worth keeping as-is

Bulk-only and why; `max_batch = 1` still being a legitimate provider; failure-per-reference; single-flight; last-good surviving a failed lookup; values dropping once past the stale window; hub-side-only and the reasoning; the split of what a provider writes vs what the runtime owns; "a hundred rows in two calls". The `~/.mycelium/.env` regeneration warning is a real catch — that file *is* rewritten by `mycelium config apply`, and a reader hand-editing a token in would lose it silently.

I also reran the generator on this branch: `git status` comes back clean, so the drift-gate claim holds.

### Not faithful

Six inline comments, in severity order:

1. **"the hub can resolve them"** — nothing imports `app.services.status`; there's no registry, route, scheduler or `Context` implementation. The module is unreachable code.
2. **Lens-scoped refresh** — exists in no code. It was an idea in my #768 description; it should not become the first thing a reader believes about the cost model.
3. **`--group status` mixing decisions and pull requests** — wrong in both halves, and the second half is a hole in my sketch: there are two closed vocabularies that both contain `blocked`, and I never said how provider state lands on a row without colliding with the row's own `status`.
4. **`unknown` and `ok` in the pull-request table** — `unknown` is unreachable in `github.py`; `ok` requires an approval, not just green checks.
5. **`GITHUB_TOKEN` from the environment** — an invented mechanism, because there was nothing real to point at. My fault, not yours; I'm fixing the code so there is.
6. **`ctx.http` carrying the credential** — the one place the docs are right and my code is wrong. Keep it; I'm changing `github.py` to match.

### Two things you flagged, answered

- **Landing ahead of the wiring** — I'd land it, but rewrite the usage sections in the future tense under a single honest banner. Right now they're present-indicative with runnable-looking commands that won't do anything if typed, which is the failure mode worth avoiding: a reader who tries it concludes it's broken rather than unbuilt.
- **Not inventing `--json` / `--max-age`** — correct call, and the right instinct. The agent-side bargain is the part I most want documented, and the honest reason it isn't is that the flag doesn't exist.

The asks are about tense, not substance. The mechanical explanations here are better than the ones in my own PR description, and I'd like them to survive to whenever this is true.

---

juliarvalenti:
The `ctx.http` inconsistency is fixed on the base branch (`9823d91`), so the paragraph you wrote is now true of the code as well as the contract.

**What changed in #768:** a provider names its credential and never sees the value.

```python
class GitHubProvider:
    name = "github"
    base_url = "https://api.github.com"
    credential = "GITHUB_TOKEN"   # named, never read
```

`ctx.secret()` is gone. The runtime resolves the credential and hands back a transport already bound to the host, so `fetch` is one call with no headers and no auth branch — the file people copy no longer teaches them to handle a token at all. Two properties fall out: a provider that cannot read a credential cannot leak one, and it cannot reach a host it never declared.

Refusal also moved up a level. A provider whose credential is missing is **never called** — the runtime marks every ref in the chunk with what's absent, rather than each provider spending a request to discover it has no token. Three new tests cover it, including that one unconfigured provider doesn't stop the others.

**For this PR**, that makes the "Giving the hub a credential" section describable for the first time, with one correction: the token is resolved *by the runtime*, not read by the provider from the environment. Where the runtime gets it from is still undecided — that's the wiring — so the honest version is something like "the hub is configured with a `GITHUB_TOKEN`; the provider declares that it needs one and never handles it". Your `~/.mycelium/.env` warning stands either way.

The other five comments still apply as written.

---

mycelium-cli/src/mycelium/docs/concepts/board.md
juliarvalenti: Confirmed and fixed in 15cd9b6. I ran the same grep and got nothing back, so "the hub can resolve them" was claiming a code path that has no caller.

The clause is gone from the top callout, which is back to describing only the board's own scope. The live-status material now sits under its own banner: the backend has the resolver, nothing attaches its answers to a row, so no row shows a pull request's state today.

---

mycelium-cli/src/mycelium/docs/concepts/board.md
juliarvalenti: Cut, not softened — you're right that it was yours rather than the code's, and I lifted it from the #768 description without checking it against `runtime.due()`.

The paragraph is replaced by the two properties that *are* in the runtime and make the same point about cost: identical references answered once (single-flight), and a tool asked about many references in one call rather than one at a time. Neither commits us to what the sweeper eventually keys off.

---

mycelium-cli/src/mycelium/docs/concepts/board.md
juliarvalenti: Checked `contracts/board-vocabulary.json` and you're right on both halves — `statuses` is `open · claimed · in_progress · in_review · blocked · resolved · dismissed`, so an unanswered decision is `open`, and `--group status` groups that field, which provider state isn't.

The example is gone rather than corrected. Documenting around the collision would have picked the answer for you in a doc, which is the wrong place for that decision to get made — it's yours on #768.

---

mycelium-cli/src/mycelium/docs/concepts/board.md
juliarvalenti: Took the retitle, since the architecture page points at this table for the whole vocabulary and dropping `unknown` would have left that cross-reference describing five states.

The column is now "What it means" and reads generally (`ok` = the thing it was waiting for happened; `unknown` = the provider couldn't map what the tool said). Both of your corrections moved into a sentence under it: `ok` means **approved**, so green checks with no review are `pending` / `awaiting review`, and the GitHub provider never emits `unknown`. The person-vs-machine distinction between `blocked` and `failed` survives there too.

---

mycelium-cli/src/mycelium/docs/reference/architecture.md
juliarvalenti: Invented mechanism removed. The section no longer says where the value is read from — it says a provider declares the credential it needs and never handles it, and that how the hub is given that value is still being settled, so there's nothing to configure yet.

The `~/.mycelium/.env` warning stays, reworded so it doesn't depend on the mechanism: whatever it lands on, a token hand-edited into that file disappears the next time `mycelium config apply` runs, with no error to explain where it went.

Leaving this open for your `credential` change — ping me here when it lands and I'll write the real path.

---

mycelium-cli/src/mycelium/docs/reference/architecture.md
juliarvalenti: Kept as written. Worth flagging one knock-on: since the prose now says a provider declares its credential and never handles the value, the `JiraProvider` example would have contradicted it while still showing `ctx.secret()`-era shape — so it gained `credential = "JIRA_TOKEN"` with a note that the runtime resolves it. That's ahead of the code by the same margin as the paragraph you asked me to keep, and the section's banner covers both.

If the attribute ends up named something else, this example is the other place to change.

---

mycelium-cli/src/mycelium/docs/concepts/board.md
```diff
@@ -146,6 +146,64 @@ is a link, it's a link and not a copy:
...
+costs no more than referencing it once.
+
+The row will show GitHub's own words — `CI failing`, `changes requested`,
+`draft`, `merged` — because that's the phrasing you already recognise.
+Underneath, each is filed as one of six states, which is what a surface can
+sort, filter and colour by without knowing what a pull request is:
+
+| State | What it means |
+|---|---|
+| `ok` | done with, in the affirmative sense — the thing it was waiting for happened |
```
juliarvalenti: This is the one row I'd still change, and the reason it's wrong is that I never wrote the semantics down — `State` was a bare `Literal` with a one-line comment, so you had to infer them. Fair inference; wrong answer.

`ok` isn't a kind of completion. It's **healthy and still in flight**: an approved pull request is `ok` right up until it merges, at which point it becomes `done`. Reading `ok` as "done with, in the affirmative sense" and `done` as "closed out, however it ended" makes them sound like *good ending* vs *any ending*, which would put an approved-but-unmerged PR and a merged one in the same bucket — and those are exactly the two a board most needs to separate.

I've fixed the root cause in #768 (`4eac2ee`) — the six states are now defined on the type, with `ok` vs `done` called out as the pair that gets conflated, and five tests holding the GitHub mapping to it:

```
ok       nothing is wrong and nobody is needed. Healthy, not finished.
pending  in motion, nobody is required.
blocked  waiting on a person: a decision, a revision, an approval.
failed   waiting on a fix, and a machine is what said no.
done     terminal, however it ended — the label carries how.
unknown  the provider met a state it cannot place. Never a default.
```

Your other five rows match that already. Suggest just taking `ok` as "nothing is wrong and nobody is needed — healthy, not finished".

---

mycelium-cli/src/mycelium/docs/reference/architecture.md
```diff
@@ -189,3 +189,82 @@ mycelium agent create design-agent --adapter cursor \
...
+### Teaching Mycelium another tracker
+
+A provider is one small class in `app/services/status/providers/`;
+`providers/github.py` is written to be copied. It declares its batching and
+freshness, then implements two methods:
+
+```python
+class JiraProvider:
+    name = "jira"
+    credential = "JIRA_TOKEN"       # resolved by the runtime; never seen by the provider
```
juliarvalenti, line 242: Missing `base_url`, which landed in the same commit as `credential` and is now part of what a provider declares:

```python
class JiraProvider:
    name = "jira"
    base_url = "https://your-org.atlassian.net"   # ctx.http is bound to this
    credential = "JIRA_TOKEN"
```

It matters more than it looks: `ctx.http` is bound to that host, which is what makes "a provider cannot reach a host it never declared" true rather than aspirational. Someone copying this template as-is would leave `fetch` with a transport pointed nowhere.

Not your miss — `base_url` didn't exist when you wrote this and your branch predates the commit.

---

mycelium-cli/src/mycelium/docs/reference/architecture.md
```diff
@@ -189,3 +189,82 @@ mycelium agent create design-agent --adapter cursor \
...
+
+How the hub is given that value is still being settled, so there is nothing to
+configure yet. One thing is worth knowing whatever it lands on: don't hand-edit
+`~/.mycelium/.env`. That file is regenerated by `mycelium config apply`, and a
+token written in by hand disappears the next time it runs, with no error to
+explain where it went.
+
+A provider whose credential is missing refuses each reference and says why,
+rather than answering with a blank — a blank on a row reads as *this pull
+request has no CI*, which is worse than an honest gap.
```
juliarvalenti: Small attribution correction: it's the **runtime** that refuses, not the provider. A provider whose credential is missing is never called at all — the runtime marks every reference in the chunk with what's absent, so no request is spent discovering there's no token.

The user-visible outcome is exactly what you describe, so only the subject of the sentence changes: "*A reference whose provider has no credential is refused with the reason, rather than answered with a blank*" or similar.

Worth keeping the distinction because it's load-bearing for the sentence just above it — a provider never handles credentials, and that includes never having to check whether it has one.

---

juliarvalenti (commented):
Re-reviewed against the code. **All six findings addressed**, and two of the fixes are better than what I asked for.

### Fixed

| | |
|---|---|
| "the hub can resolve them" | Gone from the top callout; replaced by a scoped **Not built yet** banner that says plainly no row shows a pull request's state today |
| Lens-scoped refresh | Removed, and replaced with something *true*: identical references answered once, many references per call. That's dedupe + bulk, both of which the code actually does |
| `--group status` | Removed |
| `unknown` / `ok` in the PR table | Table is now the general vocabulary, with a separate paragraph on how GitHub maps onto it more narrowly — including that the GitHub provider never emits `unknown` |
| `GITHUB_TOKEN` from the environment | Now "a provider declares which credential it needs and never handles the value", plus an explicit "how the hub is given that value is still being settled, so there is nothing to configure yet" |
| `ctx.http` carrying the credential | Unchanged, and now true of the code too |

Verified the drift gate myself on this head: regenerating produces no diff.

Two things I'd single out. **"How the hub is given that value is still being settled, so there is nothing to configure yet"** is the sentence I'd want in every doc that lands ahead of its wiring — it tells a reader the difference between *you haven't set this up* and *this isn't finished*, which is the distinction the first revision lost. And the `~/.mycelium/.env` warning survived as standalone advice rather than being cut with the mechanism it was attached to; it's still true and still saves someone an hour.

You also picked up the credential redesign from a comment, without the commit being on your branch. That's why the two remaining nits exist — see below.

### Three left, two of them mine

1. **`ok` in the state table** — the only substantive one, and it's my fault: `State` was a bare `Literal` with no semantics, so you inferred them. `ok` is *healthy and still in flight*, not a kind of completion; an approved PR is `ok` until it merges. Fixed at the root in `4eac2ee` — the six states are now defined on the type with `ok` vs `done` called out, and five tests hold the GitHub mapping to it. Your other five rows already match.
2. **The Jira template is missing `base_url`** — new in the same commit as `credential`, and load-bearing: `ctx.http` binds to it, which is what makes "cannot reach a host it never declared" real.
3. **"A provider whose credential is missing refuses each reference"** — the *runtime* refuses; the provider is never called. Same outcome, different subject.

### One mechanical note

This branch is based on `1c0bdfc`; the base has since moved to `4eac2ee` (credential redesign, then state semantics). Worth merging the base in before this lands, and regenerating afterwards — the generated HTML will pick up the state-table change.

Nothing here blocks. The prose is in better shape than my PR description was.

---

mycelium-cli/src/mycelium/docs/concepts/board.md
juliarvalenti: Taken, and the distinction is worth more than the row — I had `ok` and `done` reading as *good ending* vs *any ending*, which collapses exactly the two things a board exists to keep apart.

`ok` is now "nothing is wrong and nobody is needed — healthy, not finished", and the other five rows moved to your wording too. Added a line under the table making the pair explicit: an approved pull request is `ok` right up until it merges, and `done` the moment it does.

---

mycelium-cli/src/mycelium/docs/reference/architecture.md
juliarvalenti: Added, with your comment on it. Merged the base in first so the template matches `github.py` rather than my guess at it.

The `ctx.http` paragraph moved with it: it now says the transport arrives bound to the declared `base_url` with the credential applied, so a provider "names a host and a secret and is handed neither". That's the sentence the host binding makes true, and it wasn't sayable before this commit.

---

mycelium-cli/src/mycelium/docs/reference/architecture.md
juliarvalenti: Fixed — the subject is the reference now: *"A reference whose provider has no credential is refused with the reason, rather than answered with a blank."*

Kept your reason for the distinction rather than just the correction, since `_run_chunk` refuses before the provider is called: a misconfigured provider never spends a request discovering it has no token. That's the concrete form of "never handles credentials", so it earns the sentence.

---

mycelium-cli/src/mycelium/docs/reference/architecture.md
juliarvalenti: `9823d91` is merged in, so this can close.

The section still doesn't name where the hub gets the value — the runtime takes a `credentials` mapping and nothing yet decides what fills it — but the provider's half is now real rather than promised: it declares `credential` (and `base_url`), and there's no way to read the value. The "nothing to configure yet" sentence stays until something fills that mapping.

---

mycelium-cli/src/mycelium/docs/reference/architecture.md
juliarvalenti: True of the code now — `ctx.secret()` is gone and `github.py` sets no auth header. Closing this one.

The paragraph gained the host half in the merge: bound to the declared `base_url`, credential applied, so a provider names a host and a secret and is handed neither.

---

Issue #770, opened 2026-08-22
Frontend SSE streams exhaust the HTTP/1.1 6-connection cap → "Reconnecting…" + stalled renders


juliarvalenti:
## Symptom

Opening a room in the GUI intermittently gets stuck on **"Reconnecting…"** with the room view failing to render (blank / partial), especially over plain HTTP (local dev `next dev`, the standalone Docker image, or an appliance deployment with no HTTP/2 termination). It clears on its own after a while, then recurs. Very frustrating to work against.

## Root cause

The browser caps concurrent connections **per origin at ~6 over HTTP/1.1**, and a held `EventSource` occupies one for its entire lifetime. The app opens up to **four** long-lived SSE streams at once, plus SWR is polling several REST resources on top:

| Stream | Opened in | Scope |
| --- | --- | --- |
| room messages | `src/components/event-stream.tsx:394` | per room view |
| L9 wire feed | `src/components/l9-inspector.tsx:418` | per room view (inspector) |
| app events | `src/components/rooms-sidebar.tsx:77` | global, always on |
| notifications | `src/components/notifications-provider.tsx:155` | global, always on |

Four held SSE connections leave only ~2 of the 6 for every REST call the page makes (roster, plan, memories, episodes, search, presence polling). Under load — or with a second tab open — the pool is exhausted: new requests queue indefinitely and the `EventSource`s that drop can't re-establish, so the connection badge sits on "Reconnecting…" (`src/app/room/[name]/page.tsx:213`) and data-dependent panels never fill.

This does **not** bite in production behind an HTTP/2-terminating proxy, because HTTP/2 multiplexes all of these over a single TCP connection. So it's specifically a **plain-HTTP / HTTP/1.1** problem: `next dev`, the standalone image run directly, and bare-IP appliance deploys.

## Options (in rough order of leverage)

1. **Multiplex the SSE into one stream.** Serve a single `EventSource` (e.g. `/api/stream`) that fans out event *types* (room messages, L9, app events, notifications) as named SSE events, and have each consumer subscribe to the type it cares about. Removes the dependency on transport entirely and drops 4 held connections to 1. This is the robust app-level fix.
2. **Share one connection across tabs/components** via a `SharedWorker` or `BroadcastChannel`, so N tabs and N components don't each open their own set.
[...]

Issue #755, opened 2026-08-22
bug: `pnpm dev:mock` isn't self-contained — /skills falls through to a real backend, and presence is always empty


juliarvalenti:
## What

`src/mocks/README.md` and `src/mocks/index.ts` state the promise plainly:

> With `MYCELIUM_UI_MOCK=1`, the Next route handlers serve fixtures instead of proxying to a real backend — so the whole UI renders populated / in-progress / empty states with no SLIM node, no LLM, and no backend server.

Two endpoints break that. Opening any room under `pnpm dev:mock` produces a console error and lights the Next dev-tools issue badge:

```
ERROR Failed to load resource: the server responded with a status of 502 (Bad Gateway)
ERROR [mycelium] fetch failed: /api/rooms/atlas-migration/skills
      Error: Backend unreachable at http://localhost:8000
    at apiFetch (…)
    at async fetchSkills (…)
```

## Why

1. **`/api/rooms/{room}/skills` has no mock route.** `src/mocks/handlers.ts` handles `search`, `title`, `tasks`, `l9`, `graph`, `integrity`, `expand`, `members`, and episodes, but nothing matches `skills`, so the request falls through to the proxy and out to `http://localhost:8000`. `useRoomSkills` runs on every room page (it backs the composer's `/` autocomplete), so this fires every time.

2. **`/api/rooms/{room}/sessions/members` is stubbed empty** — `handlers.ts:352` returns `{ members: [] }` unconditionally. Nothing is *broken* by that, but it means presence can never be exercised in mock mode: the Members rail's live/awaiting badges, `useRoomRoster`'s presence map, and anything keyed on a resident agent all render as if the room were empty. The fixtures otherwise go to some length to make a populated room look real.

## Why it matters beyond the console noise

- Screenshot and demo runs go through `pnpm dev:mock` (`screenshots/capture.ts`), so every captured state is one where a backend call is failing in the background.
- #484 wants the mock GUI hosted on GitHub Pages as an interactive demo. As it stands that build would ship a page making a doomed request to `localhost:8000` for every visitor.
- Presence is a core part of the product story (server-held membership, resident agents). It is currently the one populated-room behaviour the mock can't show.

## Repro

```bash
cd mycelium-frontend && pnpm dev:mock
# open http://localhost:3000/room/atlas-migration, check the console
```

## Suggested fix

- Add a `skills` branch to `handlers.ts` returning a small fixture (a couple of `skills/…` memories already fit the shape).
[...]

PR #771, opened 2026-08-22
Move the board's demo layer out of the GUI into dev:mock

juliarvalenti:
Stacked on the coordination-board PoC (#752).

The board shipped a synthetic **demo layer** baked into the app itself — rows for the sources a live board would fill from (presence, CI, PRs) that the PoC has no reader for yet — defaulting **on**, with a `d` toggle, a header checkbox, and dashed styling. That put illustrative data inside the real GUI.

This moves it where the rest of the project's fake data already lives: the **`dev:mock`** fixtures. The app GUI now renders only real room state, and `pnpm dev:mock` renders a populated board through the **same real projection** — no in-app demo path.

## Remove (in-app demo layer)

- Delete `src/lib/board/demo.ts`.
- Strip all plumbing: the `demo` toggle state, the `d` keyboard shortcut, the "demo layer" header checkbox (+ its `FlaskConical` icon), the footer "demo layer on" text, the `demo?` inputs on `projectItems`/`projectActivity`, the `demo?` field on `LiveItem`/`ActivityEvent`, and the `DemoTag` + dashed styling across cockpit/kanban/table/timeline/daily.

## Seed (`src/mocks`)

- `MockMemory.value` now accepts an object, so a board row projects its typed fields straight from it. This is the store's **structured value** shape (a memory whose `value:` frontmatter key holds a mapping — what `MemoryCreate.value` as an object round-trips to), a real, client-reachable path. It is NOT arbitrary markdown frontmatter, which the read path drops — that gap is filed separately as #772.
- Add coordination-state memories to `atlas-migration` under `decisions/ work/ failed/` carrying `status / kind / owner / priority / ci / pr / branch / blocks / choices`, covering every lens (needs-you, in-flight, resolved) and every inferred column.
- Add per-room `presence` to the fixture (`growth` on a SLIM socket, `risk` on a lease), served from `/sessions/members`, so the board's resident rows appear with no live node.

## Testing

- `next build` (runs `tsc`) — clean.
- Board suite: **46 passed** (`activity.test.ts` 19, `board.test.ts` 27 incl. the vocabulary contract).
- Mock + room-data suite: **15 passed** (`handlers.test.ts` 11, `room-data.test.tsx` 4) — graph/integrity invariants still hold with the new rows as orphan nodes.
- Verified visually in `MYCELIUM_UI_MOCK=1`: `/room/atlas-migration` shows a populated board across cockpit/kanban/table/timeline with no demo toggle in the UI.

## Notes

- Still part of the DO-NOT-MERGE PoC (#752) — this is a cleanup against that branch, not a merge to main.
- Unrelated find while running this: the frontend's held SSE streams exhaust the HTTP/1.1 6-connection cap on plain HTTP → filed #770.


mycelium-frontend/src/mocks/fixtures.ts
```diff
@@ -246,6 +255,159 @@ const atlasEpisodeSummary: EpisodeSummary = {
...
+// priority, ci, pr, branch, blocks, choices) straight through — the same path a
+// real room's `decisions/ status/ work/ failed/` memories take. Together they
+// give the board something in every lens (needs-you, in-flight, resolved) and a
+// column for every inferred field, without any in-app demo layer.
+const atlasBoardRows: MockMemory[] = [
+  {
+    key: "decisions/token-ttl",
+    value: {
+      title: "JWT access-token TTL: 15m or 60m?",
+      status: "open",
```
juliarvalenti, line 277: Seeding `value` as an object is the right call and it does match a real API shape — but not the one the PR description names, and the difference matters.

The description says this is "the same path the real store takes serving parsed frontmatter". The real store doesn't serve parsed frontmatter. I ran the round trip through the backend's own functions — write the way `--meta` writes, read the way the route reads:

```
on disk           status: open, owner: '@julia', priority: high
parser sees       {'status': 'open', 'owner': '@julia', 'priority': 'high'}
API returns       {'text': 'Blocked behind the custody seam'}
```

`_reconstruct_value` returns `meta["value"]` or `{"text": content}`, and `MemoryRead` has no generic passthrough, so arbitrary frontmatter never reaches a client. What *does* reach one is a structured `value` — a memory created with `MemoryCreate.value` as a dict, round-tripped through the `value:` frontmatter key. That's exactly what these fixtures encode, so they're faithful to a reachable shape.

The catch is that it's not the shape the board docs promise ("a row is a title plus whatever its markdown frontmatter carries"). Someone following those docs writes `--meta status=open`, gets a blank row and no error — while the mock shows a populated board. That's my gap, not yours, and this PR only inherits it; filed as #772.

Nothing to change here. Worth fixing the one sentence in the description so the next person doesn't take "frontmatter" literally and go looking for a code path that doesn't exist.

---

juliarvalenti (commented):
**Would approve.** (GitHub won't let the same account approve its own PR, so: comment.) I ran it rather than read it — 61 tests pass across the board, mock and room-data suites, and the mock board renders populated with no demo affordance anywhere in the UI.

This corrects a mistake I made. `dev:mock` is the repo's stated home for fake data — *"the whole UI renders populated / in-progress / empty states with no backend"* — and I built a second mechanism for the same job inside the shipped GUI, defaulting on. One of them had to go, and it wasn't the one with a README.

It's also **a better proof than what it replaces.** My demo layer constructed `LiveItem`s directly, skipping the projection and the schema inference — so it exercised the renderer and nothing else. These fixtures go in as memories and come out through the real path, which means the mock board is now evidence that projection and inference work rather than evidence that React does. The tell is in the rows themselves: provenance reads `decisions/token-ttl` and `failed/thin-spoke` instead of "demo layer".

Seeding presence is a bonus — it closes half of #755 (skills is still unmocked, so leave that one open).

### Three notes, none blocking

1. **The fidelity claim in the description is wrong** — inline. "The same path the real store takes serving parsed frontmatter" isn't what happens; the store drops arbitrary frontmatter on read. These fixtures match the *structured value* path instead, which is real and reachable, so the code is fine and only the sentence needs correcting. It did surface a genuine product bug underneath, now filed as **#772**: `--meta` frontmatter is written, persisted, documented as user data, and unreadable by any client.

2. **The demo room looks abandoned.** Every row reads `10d`, `@risk` shows "seen 243h ago", TTL bars are drained red — while the header says "6 resolved today". This is the room every screenshot and demo comes from, so the timestamps are worth pulling forward; a board that looks neglected undersells the thing it exists to show.

3. **Rebase before merge.** The branch point predates the main-merge on the PoC branch, so a tip-to-tip diff reads as though this deletes shotkit's video capture (~1,700 lines). It doesn't — three-way merge is fine — but I checked twice before believing it, and the next reader may not.

The direction is right and the execution is cleaner than what it replaces. Fix the one sentence, freshen the timestamps, rebase, and it's good to go in.

---

PR #785, opened 2026-08-22
Status credentials: produce the mapping the runtime resolves against

juliarvalenti:
Draft. Targets `claude/status-integrations-design-k3m9x2` (the status-provider
contract branch, PR #768), not `main` and not the board PoC branch.

## What this closes

`StatusRuntime` takes `credentials: Mapping[str, str]` and renders those values
onto the wire for each provider's declared `auth` scheme. Nothing anywhere
produced that mapping. This does, on both sides of the one thing the thin CLI
and the backend can share: a file.

## The credential source (backend)

`app/services/status/credentials.py` resolves names to values, env over file,
mirroring `mycelium/agent_credentials.py`:

- **Flat name -> value.** The store knows nothing about schemes (#782): it
  answers "what is name X" and "is X present". How many names a provider needs is
  the scheme's business.
- **Not `config.toml` / `config.json`.** Both are rewritten wholesale by
  `save()` / `_write_json_snapshot`, and `config apply` regenerates `.env` from
  `config.toml`, so a value written there disappears. The store is a dedicated
  `0600` `status-credentials.json`, exactly as `agent-credentials.json` is.
- **Backend reads its own file.** No import of the CLI module. Confirmed the
  compose bind-mount of `${MYCELIUM_DATA_DIR:-~/.mycelium}` ->
  `/home/mycelium/.mycelium` makes a host-written store readable in the
  container with no new plumbing (`compose.yml:78-80`).
- **Env wins over the file.** A status credential name *is* an env var name, so a
  container with one `GITHUB_TOKEN` and no store file works.
- **Absent is not empty.** A configured-empty name is present in the resolved
  mapping with value `""`; an unset name is absent. That is what lets the
  runtime's error say *set but empty* versus *not configured*.

## The CLI (`mycelium board credential set/ls/rm`)

Modelled on `agent credential set`:

- value from a prompt or stdin, **never argv** (shell history, `ps`);
- `ls` shows names and set/empty, **never a value** (a test asserts the value
  never appears in output, the equivalent of `github.py`'s no-`Authorization`
  test);
- `0600`, and the command says so.

Home: `board`, because status providers are what keep board rows live, so
`mycelium board credential set GITHUB_TOKEN` is where an operator managing the
board's live data will look.

## The conformance check (registration time)

`StatusProvider` is a structural `Protocol`, so a provider that **forgets**
`auth` reads as no-auth and goes unauthenticated, and one that **misdeclares** it
(the pre-scheme `auth = "GITHUB_TOKEN"` string) raises `AttributeError` inside
`_missing_credential` — which runs *outside* `_run_chunk`'s `try`, so it
propagates through `refresh()`'s `gather` and sinks the whole sweep, not just the
offender. Reproduced.

Put the check in `StatusRuntime.__init__` (registration), raising
`ProviderConformanceError`. It is the one choke point every provider passes
through, so the invariant "every provider has a well-formed `auth`" holds for the
runtime's whole life and the per-sweep hot path never re-validates. `AuthScheme`
is `runtime_checkable`, so the check is a cheap `isinstance`. A forgotten `auth`
is refused too: no-auth must be the explicit `auth = None`, not an oversight.

## Docs

`reference/architecture.md`: the "Giving the hub a credential" section now
documents the real command and the store, drops the "nothing to configure yet"
line and the bare `.env` warning, and the Experimental callout stays honest —
you can configure a credential now, but nothing *spends* it yet (no route, no
schedule, no board-row field constructs the runtime).

## Still not this handoff

No route, no scheduler/sweeper, no board-row field, no UI. Nothing constructs a
`StatusRuntime` in the app; that stays true.

## Gate

- `fastapi-backend`: `pytest` 772 passed, 9 skipped (was 758; +14 new); `ruff
  check`, `ruff format`, `ty check` all clean.
- `mycelium-cli`: `pytest` 530 passed, 2 skipped; `ruff check`, `ruff format
  --check`, `ty check` all clean.
- `generate_docs.py` runs clean and is idempotent (no drift).

## What fought me (a real design point)

The handoff's "absent is not empty" is not just a store nicety — it forced a
change in the **runtime**, which the handoff framed as already done. `Mapping[str,
str]` can carry the distinction (a key present with `""` vs absent), but
`_missing_credential` collapsed it with `not self._credentials.get(name)`, and
the caller hard-coded `"... not configured"`. So producing the mapping correctly
is necessary but not sufficient: the runtime had to be taught to read presence,
not truthiness, and to phrase the two cases differently. Both the resolver and
`_missing_credential` now turn on `name in mapping`, and that is the seam where
the two halves of the requirement meet.


juliarvalenti:
## Review fixes (289486f6)

Thanks for the reproduction. Three findings addressed; the env one was the real hazard.

**1. Env precedence — the footgun.** Fixed by making resolution explicit-over-ambient:

1. `MYCELIUM_STATUS_<NAME>` (namespaced env override) — highest
2. the store — the operator's stated intent
3. bare `<NAME>` env var — the convenience, now **lowest**

So an ambient `GITHUB_TOKEN` (a proxy token, unrelated to status providers) can no longer beat a value the operator explicitly stored: the store outranks it, and only the namespaced form can override the store. The bare-name convenience the handoff asked for still works when nothing else is set. The divergence from `agent_credentials` (which namespaces everything and reads no bare name) is now called out in the module docstring, since I reintroduced an ambient-collision surface and had to defuse it.

**2. Test isolation.** An autouse fixture now `monkeypatch.delenv`s every name under test, bare and namespaced. Verified green with `GITHUB_TOKEN=ambient uv run pytest` — the exact condition that failed on your box. Added a regression test (`test_the_store_beats_an_ambient_bare_env_var`) that pins the hazard directly.

**3. Corrupt store ≠ absent.** `_read` (both backend and CLI) caught `OSError`/`ValueError` identically and returned `{}`. Now a **missing** file is empty (legitimate), but a present-but-unparseable one raises `CredentialStoreError` — so a damaged store surfaces instead of making every provider report "not configured". `ls` exits non-zero and says so; `set`/`rm` refuse to merge into a file they couldn't parse.

**Gate:** backend 774 passed / 9 skipped (+2 net new tests), CLI 532 passed / 2 skipped, ruff + format + ty clean on both, docs regenerate idempotently (md5-stable), zero em-dashes.

Issue #825, opened 2026-08-23
One substrate: retire `plan/`, and make a row the only unit of work


juliarvalenti:
The board was built on a thesis, stated in #752:

> A row is a title plus its frontmatter. The schema is *inferred* from that frontmatter across the collection... you never define a schema.

`plan/` cannot satisfy that thesis, and never will. A `- [ ]` line in a markdown checklist **has nowhere to put frontmatter**. Everything the board does — infer fields, group by any of them, hold a lease, expire — needs a place to write, and a checklist line has none.

So plan is not a screen inside the board. It is a second store, and the board special-cases it everywhere to hide that.

#823 retired the plan *screen* and explicitly kept the concept ("plan tasks still project as rows, consensus still compiles into `plan/tasks.md`"). That is the wrong half. This issue is for the other one.

## The special-casing, as it stands

| Every other row | `plan/` |
|---|---|
| a memory + its frontmatter | its own API types (`PlanResponse`, `PlanTask`) and endpoints |
| fields read from frontmatter | `projection.ts:planItem` hardcodes `kind: "action"`; status derived from `- [x]` |
| carries a TTL | `ttl_minutes: null`, with a comment explaining why it must be |
| claimable — custody is a lease | **refused**: `"a plan task is the room's commitment, not a lease: assign it with @handle"` |
| no dedicated rail (the `skills/` principle in CLAUDE.md) | its own screen, holding the GUI's only editing affordances |
| — | `plan.py` (350) + `plan_compiler.py` (222) + `plan_sync.py` (189) |

Each of those is locally reasonable. Together they are a namespace that had to be exempted from the model in order to live inside it.

## Why it exists at all

`plan_compiler.py` says so in its own docstring: it consumes the aligner's `{issue: value}` agreement and compiles it into a markdown checklist. `plan/` is not a general work surface that the aligner happens to write to — it is **the aligner's private output format**, which the board then had to accommodate.

That is the actual finding here. One engine has a privileged output path, and the substrate grew a second store to serve it.

## The blocker: the board barely writes back

This has to be said before anything is deleted, because it is a real sequencing dependency and the obvious ordering is wrong.

[...]

Issue #823, opened 2026-08-23
Finish superseding the plan screen: delete the board's Docs view and RoomPlanHeader


juliarvalenti:
The board was meant to replace the plan screen. It did not, quite: `RoomPlanHeader` is still alive behind the board's **Docs** view, which renders the old screen whole inside a tab of the new one.

That was not a design decision, it was a shim. Deleting it would have dropped two capabilities with no replacement, so it stayed, and then it shipped. This is the issue for finishing the job.

## Why it is still there

`room-plan-header.tsx` is the only place in the GUI that can:

- set the room's display title (`setPlanTitle` → `plan/title.md`)
- add a plan task (`addPlanTask` → `plan/tasks.md`)

`memory-detail.tsx` has no editing at all, so nothing else covers either.

## The principle it violates

CLAUDE.md is explicit that a promoted namespace does not get its own rail:

> **In the GUI there is deliberately no dedicated skills rail** — a skill shows up in the Memory list like any `skills/…` memory, with a small "skill" tag in the detail view.

`plan/` is the same shape as `skills/`: a namespace of memories, promoted. It should not have a tab either.

## Why the board does not need a prose view

The board is a lens on what needs attention. A plan file's narrative does not need attention — it is reference material, and `plan/{slug}.md` is already a memory, already listed under Memory. Keeping a reading surface inside the board was the error; there is nothing to relocate, only something to stop doing.

## The three moves

1. **Title editing onto the board hero.** The italic room title is already rendered by the board (`plan/title.md`); it just is not editable there. `setPlanTitle` moves with it.
2. **Task creation into capture.** Capture already writes memories. A line shaped like a task, or a deliberate sigil, routes to `plan/tasks.md` rather than `work/` — one gesture instead of a second screen. `parseCapture` already extracts the owner, so `- [ ] flip reads behind a flag @growth` needs no new parsing.
3. **Delete the `docs` view mode and the board's use of `RoomPlanHeader`.** Drop it from `ViewMode`, from `VIEW_TABS`, from the `mode === "docs"` branch, from `SAVED_VIEWS` if present, and from the Board concept page ("Docs: the plan's prose, for when you want to read rather than triage").

Keep `RoomPlanHeader` itself if another surface still mounts it; delete it if nothing does.

## The gap this exposes, to be stated rather than hidden

**No memory is editable in the GUI.** Once Docs is gone you can read plan prose under Memory but not change it, where the tab let you. That is a real regression for that one file.

[...]

PR #827, opened 2026-08-23
Make a board verb write to the room, not to the browser (#825)

juliarvalenti:
## Summary

Only `claim` / `release` / `resolve` had a write behind them, and only on `work/` rows. Every other verb — `block`, `dismiss`, `promote`, a kanban drag, a table cell edit — set a React overlay that evaporated on reload. Meanwhile the table's inline editor already echoed **"· written to frontmatter"** while writing nothing, and the CLI's `_would_write` printed a patch it never sent.

A surface asserting a change the room was never told is the failure the board's own custody model exists to prevent — it is the same shape as a claim held as a fact rather than a lease.

The board's thesis is that **a row is a title plus its frontmatter**, so a verb is a frontmatter write. This makes that true.

It is the first of the three moves in #825, and the one the other two are blocked on: `plan/` cannot stop being a store until the board can write a row.

## Changes

**Backend**

- `app/services/fields.py` — writes a frontmatter patch through the canonical memory upsert, so a card moved in a browser is the same versioned, indexed, link-parsed, broadcast change a `memory set --meta` makes. `null` clears a key.
- `app/routes/fields.py` — `POST /rooms/{room}/fields`, `GET /rooms/{room}/fields/{key}`. A sibling router rather than routes under `/memory`, for the reason `leases` and `links` already sit beside it: the memory router ends in a `{key:path}` catch-all.
- `leases._write` now delegates its upsert to `fields.upsert_patch`, so a lease and a status change cannot drift in how they reach the store. One writer, two callers.

**Custody is refused rather than shared.** A lease owns `custody` / `owner` / `claimed_at` / `ttl_minutes` / `custody_note` / `custody_note_by` under rules a field write cannot check — a live claim is not stealable, an expired one is, and expiry is derived from a clock rather than written down. A plain write setting `owner` would forge a claim that passed none of that, so those keys are refused by name and pointed at `/leases`. A refused write is refused whole: no partial application.

**Only a memory has frontmatter.** `plan`, `episode`, `agent`, `capture` and `github` rows refuse in their own terms instead of accepting a write that goes nowhere.

**Frontend**

- `writeFields` in `lib/api.ts`; `fieldWriteRefusal` / `memoryKeyOf` in `lib/board/fields.ts`; `RESERVED_FIELDS` / `reservedIn` in `lib/board/custody.ts` (it sits above `fields.ts`, which the module header keeps below both `item` and `custody`).
- `room-board.tsx` — the overlay becomes the optimistic echo and `patch` does the write. Custody verbs keep the lease path; everything else goes through the field write, including the kanban drag, the table edit and the cockpit's answer buttons. `updated` is stripped (it is the board's name for the store's own `updated_at`, so sending it would write a second competing timestamp) and `undefined` becomes `null` (`JSON.stringify` drops the former, so a cleared field would silently survive).

**`promote` no longer invents a GitHub back-link.** It was writing `700 + (overlay.length % 90)` as an `issue:` field. Survivable as an overlay; written to the room it is a fabricated reference the upstream provider would then resolve against a real, unrelated issue.

**CLI**

- `_would_write` becomes `_write_fields`. `board block --on` writes, and `board resolve` on a memory row that cannot hold a lease writes `status: resolved` instead of describing it.
- `mycelium/board/fields.py` mirrors the GUI's refusals and reserved set.

**Contract**

`contracts/board-vocabulary.json` grows a `fields` block for the refusals and the writable source kinds. It deliberately carries **no copy of the reserved keys** — those are derived from `custody` (its field, its companion fields, and the holder), and all three implementations assert that derivation rather than a fourth list that could drift.

**Docs** — the Board concept page's "Experimental" note said the verbs act on your own view; it now says what they do. The `board.py` module docstring said the same. Docs site regenerated.

## Testing

Mutation-checked, not just green: removing the reserved-key guard reddens 7 tests, and making a missing row auto-create instead of 404 reddens the one that names it.

```
backend    930 passed, 13 skipped   (20 new)
CLI        698 passed,  2 skipped   (19 new)
frontend   704 passed              (24 new)
ruff check / ruff format --check / ty check      clean
tsc --noEmit / eslint                            clean (5 warnings, all pre-existing)
scripts/check_docs_links.py     47 documents, every link resolves
scripts/check_workflows.py      pinned, permissions declared, budgets aligned
```

`openapi.json` refreshed — the frontend's `api-contract` test caught the missing refresh before I did, which is that derivation gate doing its job.

## What is still rough

- **No memory prose editing.** This writes fields, not text. A `plan/{slug}.md` narrative is still readable and not editable in the GUI, which is the gap #823 named honestly and which move 3 of #825 makes more urgent, not less.
- **Not verified in the running app.** Every layer is unit-covered on all three sides and the API is exercised end to end in the backend tests, but a real drag in a real browser against a live hub has not been done.
- **`promote` still resolves the row without filing anything.** Dropping the fabricated issue number makes it honest, not complete: there is no GitHub write behind that verb yet.
- **Field writes broadcast.** Unlike a lease renewal, which is deliberately quiet, every verb is news. That is probably right for a status change and probably wrong for a rapid sequence of table edits.

## Related Issues

First of the three moves in #825. Does not close it — plan tasks becoming `work/` memories, and the deletion of `plan/`, follow.

---

https://claude.ai/code/session_01TiKo4bgUAxxURBvvdxAUV3

---
_Generated by [Claude Code](https://claude.ai/code/session_01TiKo4bgUAxxURBvvdxAUV3)_

### DIFF
--- a/contracts/board-vocabulary.json
+++ b/contracts/board-vocabulary.json
@@ -1,5 +1,5 @@
 {
-  "_comment": "Frozen vocabulary for the coordination board (PoC). The GUI (mycelium-frontend/src/lib/board/) and the CLI (mycelium-cli/src/mycelium/board/) each carry their own copy of the projection rules \u2014 the thin `uv tool` CLI can't import the frontend and vice versa \u2014 so the words they must agree on live here, the same way contracts/slim-l9-wire.json freezes the SLIM+L9 constants. A unit test on each side asserts against this file, so neither copy can drift without a red gate.",
+  "_comment": "Frozen vocabulary for the coordination board. The GUI (mycelium-frontend/src/lib/board/) and the CLI (mycelium-cli/src/mycelium/board/) each carry their own copy of the projection rules \u2014 the thin `uv tool` CLI can't import the frontend and vice versa \u2014 so the words they must agree on live here, the same way contracts/slim-l9-wire.json freezes the SLIM+L9 constants. A unit test on each side asserts against this file, so neither copy can drift without a red gate.",
   "version": 1,
   "statuses": [
     "open",
--- a/contracts/board-vocabulary.json
+++ b/contracts/board-vocabulary.json
@@ -138,6 +138,19 @@
     },
     "blocked_field": "blocked_by"
   },
+  "fields": {
+    "_comment": "What a board verb writes when it is not custody: frontmatter on the row's memory, through the same upsert `memory set --meta` goes through. Only a memory carries frontmatter, so every other projected row refuses in its own terms rather than accepting a write that goes nowhere. The reserved set is deliberately NOT listed here \u2014 it is derived from `custody` (its field, its companion fields, and the holder), and each implementation asserts that derivation rather than a second copy that could drift.",
+    "writable_source_kinds": [
+      "memory"
+    ],
+    "refusals": {
+      "plan": "a plan task is a checklist line; there is no frontmatter on it to write",
+      "episode": "an episode is a recorded negotiation, not a row the room can edit",
+      "agent": "presence is what the runtime reports, not a field you set",
+      "capture": "this row is not in the room yet; capture it first",
+      "github": "this value belongs to the tool it came from; change it there"
+    }
+  },
   "schema": {
     "_comment": "Field types the two surfaces must infer identically, and the order options come back in. `cases` is a table both suites run through their own classifier: a field name, the values it holds, and the type both must call it. These drifted once (the GUI grew number-from-string and link rules the CLI never got, and sorted out-of-vocabulary options first where the CLI sorted them last), which is what a table asserted from both sides prevents.",
     "types": [
--- a/docs/index.html
+++ b/docs/index.html
@@ -690,10 +690,6 @@ Blocked 1
 Review 1
  ◉ 7c2   @agent-z opened PR #504, wants eyes on the custody seam
          @agent-z   feat/custody-seam   CI green   #504       12m</code></pre>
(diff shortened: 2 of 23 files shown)


juliarvalenti:
Pushed `af2e504` — the board stops calling itself experimental.

Four places said so, and each was hedging something this PR removes:

- The **Board concept page**'s `> **Experimental.**` note. What a verb writes now sits under *One gesture each*, with the verbs, rather than in a caveat above the fold.
- The **status-provider note** in the architecture reference keeps its substance (a read reports what is known and starts a fetch for what is not) and drops the label.
- **`board-vocabulary.json`** described itself as the vocabulary for "the coordination board (PoC)". It is asserted by three implementations against a red gate; that is not a PoC's contract.
- **`VerbContext.issueNumber`** — the last PoC-shaped thing in the code, documented verbatim as *"Issue number a `promote` would file as, in a PoC with no GitHub write."* Nothing passes it after the fabricated back-link came out, so it was a dead field whose only remaining caller was its own test. Removed rather than reworded.

One correction that came with it: the docs said anything worth outliving the work gets `promote`d into a GitHub issue "with a link left behind". It never filed anything. `promote` marks a row as belonging somewhere more durable and resolves it; filing the issue is still yours to do, and the verb no longer invents a reference the status providers would then resolve against a real, unrelated one.

Re-verified after the change: backend 930, CLI 698, frontend 704, ruff/ty/tsc/eslint clean, docs regenerated and 47/47 links resolve.

---

Issue #503, opened 2026-08-14
Retire OpenClaw: ~320 stale references across metrics, UI, skills, and a leftover worktree


juliarvalenti:
`openclaw` is deprecated (per CLAUDE.md: rode the removed SSE/coordination-tick model, never migrated to SLIM) and **already gone as a live adapter** — `mycelium-cli/src/mycelium/integrations/` is only `claude_code`, `cursor`, `engine`. But the name still appears **~320 times** in the working tree. This tracks retiring the residue.

The count is misleading, so bucketing by what it actually is:

## 1. Live subsystem, not "stale refs" — the metrics stack (~140)
`commands/metrics.py` (64), `collector.py` (49), `commands/traces.py` (27). The whole observability path is an OTLP receiver for **OpenClaw's `diagnostics-otel` plugin** — `openclaw.channel` / `openclaw.agent` attributes, `openclaw status --json`, "Collect and display OpenClaw agent metrics."

**This is a real decision, not a find/replace:** is OpenClaw telemetry staying? If OpenClaw is fully dead, either retire the metrics stack or decouple it to generic OTLP (drop the `openclaw.*` attribute coupling). Needs its own scope.

## 2. Genuinely wrong / user-facing — fix now (3)
`mycelium-frontend/src/components/agents-panel.tsx` — the "Add agent" dialog still tells users `--adapter openclaw` and "discovers your OpenClaw agents," but `openclaw` isn't in `AGENT_ADAPTERS` anymore (claude_code, cursor, engine). Misleading dead-end for anyone reading it.

## 3. Prose cleanup — low risk (~89)
`.claude/skills/persona-before-and-after/SKILL.md` (46), `e2e/SKILL.md` (29), `troubleshooting/SKILL.md` (14). Update examples/flows to the live adapters.

## 4. Generated — regenerate, don't hand-edit (~18)
`mycelium-client/…/backend_client/**`, `openapi.json`, `mycelium-cli/openapi.json`. Fall out of a backend schema regen once the backend drops any `openclaw` enum/example.

## 5. Keep (intentional)
`CHANGELOG.md` (9, history), `CLAUDE.md` (1, documents the deprecation), `docs/design/*` (historical design). Leave or lightly annotate.

## 6. Not the codebase — delete
[...]

PR #829, opened 2026-08-23
Rebuild the metrics page on what the stack actually reports

juliarvalenti:
## Summary

The metrics page was reading from things the stack no longer has. `CFN HTTP`, `CFN LLM`, `KNOWLEDGE GRAPH` and `COORDINATION` all pulled from counter namespaces `app/services/metrics.py` stopped filing; `SPEND`, `AGENT ACTIVITY`, `SPEND · BY MODEL` and `SPOKE SITES` all pulled from the OTLP collector, which only parses `openclaw.*` metric names and only records spans named `openclaw.model.usage` — and OpenClaw was removed from the stack (#503). On a running hub every one of those rendered as a dash, a `0`, or an empty-state plate, so the page said the least about the hub that was busiest. Two of the sparkline fields it typed (`spend_5m`, `err_5m`) were never in any payload at all.

This rebuilds the page from the ground up on the three surfaces the current stack does produce, and removes the frontend's dependence on the collector rather than leaving a disabled-mode plate for a feature nothing can populate.

| Source | What it gives the page |
|---|---|
| `GET /api/observability` | memory, embeddings, indexer and Pi cognition counters + latency histograms |
| `GET /health` | the coordination fabric, the SLIM identity tier, the HTTP API gate |
| per-room episode records | every convening of the aligner, rolled up by outcome and L9 quality |

Two properties carry the redesign: **nothing is drawn that the stack cannot produce**, and **an unmeasured counter reads `-`, never `0`** — a panel with no data says what would put data in it instead of drawing a grid of dashes. Where a figure would be misleading, the panel says so rather than omitting it: cognition notes that `pi` reports no per-turn token usage, so calls, failures and latency are counted and spend is not claimed, where the old page implied a spend it had no way to measure.

The negotiations panel is new capability, not just a repair — it is the first fleet-wide view of whether rooms are actually reaching agreement (episodes, converged/rejected/open, compiled work rows, and mean MPC/GAR/SCR/provenance over the episodes that committed with metrics).

## Changes

- **`components/metrics-screen.tsx`** — rewritten. Header band (uptime, cadence, pause) → 3 KPIs (memory · fabric · cognition) → coordination-fabric panel with one row per live channel (channel/persister state, members present, episode, invites, durable inbox, errors) → negotiations panel → four counter panels (memory & retrieval, embeddings, indexer, cognition). Semantic colour tokens throughout; the old page hard-coded an RGB lerp toward `#f87171` and a `text-red-400`.
- **`lib/metrics-data.ts`** (new) — the page's reads and rollups. SWR-keyed like every other read in the app, so it shares the `/health` entry with the Network rail and the room list with the sidebar; the screen owns only the cadence. `useFleetEpisodes` fans out one request per room and is therefore floored at 15s however fast the page is set to refresh — pausing still stops it outright. A room whose episode read fails contributes nothing rather than blanking the panel.
- **`lib/metrics-format.ts`** — dropped `statusKind`/`errorRate` (CFN HTTP-status helpers with no remaining caller), added `fmtPct` (a rate over zero is `-`, not `0%`) and `subCounters` for reading the backend's flattened dimension keys. `subCounters` drops a dimension's nested totals (`by_operation.task_compile.input_tokens` is that operation's token count, not another operation) but keeps a dotted value whole where one is legitimate (`openai/gpt-4.1`).
- **`lib/api.ts`** — removed `fetchCollectorMetrics`, `fetchHosts` and `HostInfo`; nothing reads them now. `/api/observability/collector`, `/traces/recent` and `/hosts` are untouched on the backend and still serve the CLI.
- **`lib/use-status.ts` + `components/status-items.tsx`** — the status bar's model cell read the collector's top-token model and its spend cell read the collector's cost total; both were permanently empty for the same reason. The model now comes off `/health`'s `llm.model` (the hub's configured model) and the spend cell is gone. Both come off the one shared `health` entry, so the bar still costs one request across every page.
- **`mocks/`** — `BACKEND_METRICS` reshaped to what `/api/observability` actually returns (four namespaces, flattened dimension keys, zero token/cost counters, one histogram per measured latency); collector fixtures and their mock routes removed.
- **`docs/reference/metrics.md`** — the "what you get today" paragraph listed coordination counters that no longer exist; corrected, with the `pi` token-usage caveat stated and a pointer to the app's Metrics page. Not published to the docs site (no `SECTION_CONFIG` entry), so `generate_docs.py` is a no-op diff — verified.

## Testing

- `npx vitest run` — 57 files, 711 tests pass, including 15 new ones.
  - `components/metrics-screen.test.tsx` — draws the counters the backend reports; names each Pi operation and not a spend it cannot measure; reports the fabric room by room; rolls every room's episodes up into how they ended; says what would fill a panel rather than drawing zeros; reports an unreachable hub instead of an empty dashboard.
  - `lib/metrics-data.test.ts` — outcome counted by commit subkind not the projected outcome text; quality averaged over the episodes that carry metrics, not all of them; quality left unclaimed when nothing has committed; recent list ordered newest-first across rooms; `subCounters` nesting and dotted-value cases; `fmtPct` over a zero denominator.
- `npx tsc --noEmit`, `npx eslint .` (0 errors), `npx next build` — all clean.
- Visually verified via `shotkit` at desktop, 1440px, 560px and phone widths, in both themes, with the mock hub populated and with every panel empty. Fixes found and applied that way: the header band now wraps instead of clipping at narrow widths, the two-column panel grid is `items-start` so a panel without a footnote does not stretch, and `Panel` carries `min-w-0` so the fabric table scrolls inside its card rather than widening it.

Backend, CLI and Python packages are untouched, so the `uv run pytest` / `ruff` gates are unaffected by this diff.

- [x] Unit tests pass (frontend: `npx vitest run`; no Python changed)
- [x] Linting passes (frontend: `npx eslint .`; no Python changed)
- [x] Format check passes (frontend: `npx tsc --noEmit` + `npx next build`; no Python changed)

### Follow-ups this deliberately leaves alone

The collector itself (`mycelium-cli/src/mycelium/collector.py`, the `metrics` compose profile, `mycelium metrics collect`) is still OpenClaw-shaped end to end — it discards any span not named `openclaw.model.usage` and ignores any metric not named `openclaw.*` — and its module docstring still says "OpenClaw telemetry". Nothing in the current stack exports OTLP, so it cannot populate. Deciding its fate is a separate call from redoing this page; this PR just stops the UI claiming to show its output.

---
_Generated by [Claude Code](https://claude.ai/code/session_0182dfxxUWPcTxDYrUyi4f95)_


sfph (commented):
## Code review — metrics page rebuild

Tests pass (711/711), `tsc --noEmit` clean. Three findings:

---

### 1. Fleet episodes panel stays stale after SSE pushes — `metrics-data.ts:150`

`useFleetEpisodes` registers under SWR key `["episodes-fleet", names.join(",")]`. The app-wide revalidation hook in `room-data.ts` filters for `key[0] === "room"`:

```ts
void mutate((key) => Array.isArray(key) && key[0] === "room" && key[1] === room);
```

That predicate never matches `"episodes-fleet"`, so when an SSE event fires (episode closes, agent replies), the fleet panel isn't revalidated. It stays stale until the `EPISODE_FLOOR_MS` polling interval fires (≥15 s). The fix is to either change `key[0]` to `"episodes-fleet"` in the revalidate filter, or adopt the `["room", name, "episodes"]` key convention the rest of the app uses.

---

### 2. Two independent polls to `/api/health` — `room-data.ts:290` + `use-status.ts:68`

`useNetworkStatus` (key `"network-status"`) and `useHubHealth` (key `"health"`) both `fetch("/api/health")` but under different SWR keys, so deduplication never fires. The metrics page is the one place both are mounted — `<GlobalStatusItems />` in the layout brings in `useGlobalStatus → useHubHealth`, and `MetricsScreen` calls `useNetworkStatus` directly — so every cadence tick fires two concurrent health requests.

Root cause: the two hooks were written for different consumers (status bar vs. fabric panel) and each reached for `/api/health` independently. The fix is one shared `useSWR("hub-health", fetchFullHealth)` that returns the complete response, with each hook projecting its own slice rather than fetching separately.

---

### 3. Mock `/health` missing `llm` block — `handlers.ts:141`

The MSW mock returns `status / coordination / identity / auth` but no `llm` key. `useGlobalStatus` reads `data?.llm?.model ?? null`, so the model chip is permanently null in mock/dev mode. Any regression in the model-display path can't be caught by tests that mount the status bar against mock handlers.

```ts
// handlers.ts — add to the health response:
llm: { model: "anthropic/claude-sonnet-4-6" },
```

---

**Refuted (not a bug):** `FabricPanel` reads `r.pending_invites` — this is correct; the backend emits `pending_invites` in `room_channels.py:252` and the TypeScript type agrees.

PR #856, opened 2026-08-24
Fix spoke config apply, track dev build mode, clean up lint

sfph:
## Summary

- **Spoke `config apply` no longer generates a SLIM PSK** — `ensure_slim_master_secret` is now gated on `_is_hub_config(config)` (same hostname-in-localhost check doctor already uses). Spokes only use the HTTP API and have no SLIM node, so the PSK generation and the "Generated hub SLIM PSK" message were both wrong on a spoke.
- **Doctor auth warning gives actionable instructions** — the hint now says to run `mycelium config apply --restart` (not just `config set`) so it's clear the backend container needs to restart before the change takes effect.
- **`config apply --restart` respects locally-built images** — adds `MYCELIUM_BUILD_MODE` to `_OPERATOR_MANAGED_KEYS` (same pattern as `MYCELIUM_IMAGE_TAG`). `mycelium up --build` stamps `MYCELIUM_BUILD_MODE=dev` into `.env`; `config apply --restart` reads it and re-injects `compose-dev.yml` so `pull_policy: never` stays in effect and the restart uses the local `:dev` images instead of pulling from GHCR. All paths that transition back to pulled images (`mycelium up`, `mycelium pull`, `mycelium install`) clear the flag.
- **Regenerated `mycelium-client` from `openapi.json`** — fixes 9 pre-existing broken tests (`amend_message` endpoint, `edited_at` on `MessageRead`, `created_by` on `UserCreate`, `meta` on `MemoryRead`).
- **Frontend lint clean** — `aria-description` → `aria-label` on two button elements (`l9-inspector.tsx`, `markdown-content.tsx`); suppressed unused drain variable in `sse.test.ts`.

## Test plan

- [ ] On a spoke (`server.api_url` pointing at a remote host): `mycelium config apply` should not print "Generated hub SLIM PSK" and should not modify `slim.master_secret` in `config.toml`
- [ ] On the hub with auth disabled: `mycelium doctor` warning now reads "Enable auth on the hub: … mycelium config apply --restart"
- [ ] `mycelium up --build` → `~/.mycelium/.env` contains `MYCELIUM_BUILD_MODE=dev`; subsequent `mycelium config apply --restart` includes `-f compose-dev.yml` in the docker compose command and does not pull GHCR images
- [ ] `mycelium up` (no `--build`) → `MYCELIUM_BUILD_MODE` is absent from `.env`
- [ ] `mycelium pull` → `MYCELIUM_BUILD_MODE` is cleared
- [ ] `uv run pytest tests/ -q` in `mycelium-cli` — 687 passed, 0 failed
- [ ] `pnpm run lint` in `mycelium-frontend` — 0 errors, 2 pre-existing `set-state-in-effect` warnings only

### DIFF
--- a/mycelium-cli/src/mycelium/commands/config.py
+++ b/mycelium-cli/src/mycelium/commands/config.py
@@ -236,22 +236,41 @@ def apply_config(
                 typer.secho("  ✗ Could not find compose.yml", fg=typer.colors.RED)
                 raise typer.Exit(1)
 
-            result = subprocess.run(
-                [
-                    "docker",
-                    "compose",
-                    "-p",
-                    "mycelium",
-                    "-f",
-                    str(compose_path),
-                    "--env-file",
-                    str(env_path),
-                    "up",
-                    "--force-recreate",
-                    "-d",
-                ],
-                text=True,
-            )
+            from mycelium.docker_utils import read_build_mode
+
+            cmd = [
+                "docker",
+                "compose",
+                "-p",
+                "mycelium",
+                "-f",
+                str(compose_path),
+            ]
+
+            # When the stack was started with `mycelium up --build`, inject
+            # compose-dev.yml so `pull_policy: never` stays in effect and the
+            # restart uses the locally-built images instead of pulling from GHCR.
+            restart_env: dict[str, str] | None = None
+            if read_build_mode(env_path) == "dev":
+                dev_compose = Path(__file__).parent.parent / "docker" / "compose-dev.yml"
+                if dev_compose.exists():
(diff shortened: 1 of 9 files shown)


juliarvalenti:
Thanks for the catch — confirmed and fixed. The import path now runs unconditionally via `allow_generate=False` on spokes, so an existing secret in `.env` is promoted into `config.toml` and rendered correctly. Generation of a new secret remains gated on `_is_hub_config`. Three regression tests added covering: spoke preserves existing secret, spoke does not generate, and `allow_generate=False` no-op.

Issue #738, opened 2026-08-21
A2A bridge: consolidate duplicated room-post + _norm + manifest-write


juliarvalenti:
Non-blocking cleanup surfaced by the A2A bridge review (#719). Not bugs, but real duplication worth consolidating so the persistence contract doesn't drift per-caller.

- `_norm(handle)` is copy-pasted in ~6 places (a2a_bridge, a2a_agents, aligner, agent_registry, engines route, ...). Consolidate to one helper.
- `create_a2a_agent` closely clones `create_engine` (manifest write + AgentRead projection). Factor the shared manifest-write path.
- The `channel.send(...)` + `persister.ingest_local(...)` broadcast primitive is re-implemented in the a2a responder, the inbound a2a server, the synthesizer, and the aligner. A shared `ManagedRoomChannel.post(...)` would keep the persistence contract in one place.

PR #869, opened 2026-08-25
refactor(a2a): consolidate _norm + manifest write path (#738 PR A)

sfph:
## Summary

- **`norm_handle()` in `agent_registry.py`**: promotes the private `_norm` helper to a public canonical function (lowercase slug, no `@`, returns `str|None`). Removes the four previously duplicated private copies in `a2a_bridge`, `a2a_activity`, `a2a_agents`, and `engines`.
- **`write_agent_manifest()` in `agent_registry.py`**: factors the `yaml.safe_dump → MemoryBatchCreate(embed=False, tags=["agent-manifest"]) → upsert_memories` sequence that was duplicated identically in `create_a2a_agent` and `create_engine`. The manifest contract (tag name, embed flag, key shape) now lives in one place. Uses the established lazy-import pattern from `task_sync.py` / `synthesizer.py`.

No behaviour change.

## Test plan

- [ ] Full backend gate: `938 passed, 6 skipped`

### DIFF
--- a/fastapi-backend/app/routes/a2a_agents.py
+++ b/fastapi-backend/app/routes/a2a_agents.py
@@ -15,14 +15,13 @@ driver (#714) needs to call the remote agent.
 import logging
 import re
 
-import yaml
 from fastapi import APIRouter, HTTPException, Request
 from pydantic import BaseModel, Field
 
-from app.routes.memory import upsert_memories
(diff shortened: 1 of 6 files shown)


sfph:
Consolidated into #912

PR #871, opened 2026-08-25
feat(a2a): settle transport decision + wire A2A_ALLOW_PRIVATE_HOSTS (#726)

sfph:
## Summary

**Transport decision (#726 → closed):**
- `a2a_bridge.py` module docstring updated: HTTPS is the explicit decided default. `slima2a` was evaluated and deferred — it requires `a2a-sdk==1.1.0` exactly while the project tracks `1.1.2`, and adopting it would pull SQLAlchemy/aiosqlite/OpenTelemetry into the image for a purchase that doesn't change the honest boundary.
- `CLAUDE.md` fixed: the stale reference implied the project's `a2a-sdk` version was `1.1.0`; it is `1.1.2` and `slima2a`'s requirement is the constrained one.

**`A2A_ALLOW_PRIVATE_HOSTS` config wiring:**
- `A2aConfig` class added to `MyceliumConfig` (`allow_private_hosts: bool = False`).
- `generate_env_file()` now renders `A2A_ALLOW_PRIVATE_HOSTS` so `mycelium config set a2a.allow_private_hosts true && mycelium config apply` reaches the backend container. Previously this was only reachable by hand-editing the container's environment.
- `"a2a"` added to `_global_sections` so the field persists correctly across saves.
- `a2a-bridge.md` updated to replace the raw `A2A_ALLOW_PRIVATE_HOSTS=1` instruction with the `mycelium config set` command.
- Docs regenerated: `[a2a] allow_private_hosts = false` now appears in the Configuration reference.

## Test plan

- [ ] `test_config_env_coverage.py`: all 5 tests pass — `A2A_ALLOW_PRIVATE_HOSTS` is automatically picked up by the AST walk
- [ ] `docs/` drift check: `git diff --exit-code docs/` is clean after `generate_docs.py`
- [ ] Backend + CLI quality gates: ruff + ty + pytest all pass

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -419,13 +419,15 @@ is no litellm dependency.
   scoped), and `a2a_server.py` exposes a room *as* an A2A agent via the a2a-sdk
   server (card + JSON-RPC). **Honest boundary:** a bridged A2A agent is **not** a
   member of the room's MLS group — it's proxied by a backend seat that reads
-  plaintext and calls the remote out-of-band. The hop is plain HTTPS today; it
-  can ride SLIM (SLIMRPC) via `agntcy/slim-a2a-python`, but that's point-to-point
-  RPC to a SLIM identity, still not room-group membership, and slima2a pins
-  `a2a-sdk==1.1.0` (HTTP-vs-SLIM decision: #726). Either way the hub sees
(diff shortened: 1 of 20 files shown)


sfph:
Consolidated into #912

PR #893, opened 2026-08-26
Name a task the same way everywhere the channel mentions it (#889)

### LINKED ISSUE #889: bug: channel ping/notice rendering is inconsistent — activity shows the slug, claim shows the title; mixed font sizes
Two inconsistencies in how the channel renders board activity:

## 1. Slug vs. title

- A **claim notice** renders the task's **name/title**: `event-stream.tsx:251` → `content = notice.title ?? notice.key` (the notice payload carries the title).
- A **thread-activity ping** renders the **slug/short-id**: `event-stream.tsx:703` → `shortId = threadShortId(thread) ?? "thread"`, drawn as `Activity in <shortId>`. A ping only carries the episode + sender + message id, so there's no title on it and it falls back to the URN tail.

So the same task reads as its name in one line and a slug in the next. Activity should resolve the ping's episode to the task's **title** (look it up from the board rows / roster the panel already has) and render the name, matching the notice.

## 2. Inconsistent font sizes

The ping/notice lines don't share a type scale — `SystemNotice` and the notice/consensus rows use different sizes, so the activity feed looks ragged. Normalize the ping/notice rows to one consistent size/treatment.

Surface: `src/components/event-stream.tsx` (the `PING_TYPE` render around L701-714, the `NOTICE_TYPE` render, and `SystemNotice`), with the episode→title lookup from the room's rows.

juliarvalenti:
## Summary

A thread-activity ping printed a `work/…` slug where the board notice one line above printed the row's name. The channel already looked the title up — what it got back *was* the key.

**The root cause is not where the issue points.** `event-stream.tsx` already rendered `owner?.title ?? shortId`; the title it was handed was the slug. `memoryTitle` (`room-data.ts`) read a title off a **string** `value`, but the store keeps prose in the markdown body and rebuilds the API value from it (`routes/memory.py:_reconstruct_value`), so every prose memory arrives as `{text: …}` and fell straight through to `memory.key`. The board's projection had its own, separate derivation that happened to fall back to `content_text` — which is exactly why one surface read right and the other didn't.

**Why it shipped, and why it would have come back.** The mock served a shape the API never returns: a fixture's prose `value` went out as a bare string, so the bug was invisible in `dev:mock` and in every test that used a fixture-shaped memory. The existing channel tests all used `value: { title: … }` — a structured value — and so never touched the prose path at all.

## Changes

- **One `memoryTitle`**, in `lib/memory-preview.ts`, reading the body whichever way the store hands it over (`{text: …}`, a bare string, or `content_text`). `room-data.ts` and `lib/board/projection.ts` both call it and their two private derivations are gone — a task is a board row *and* a thread, so the two must not disagree about its name.
- **The mock serves the real shape.** `memoryRead` reconstructs `{text: …}` from a prose fixture the way the backend does. Fixtures stay readable as prose; every consumer now sees the shape it will actually be handed.
- **Knowledge notices resolve too.** A `knowledge` envelope carries a key and no prose (`memory_sync.build_knowledge_envelope` has no content field), so that line was *always* the slug — this half of the reporter's screenshot comment was a different miss from the ping's. It now resolves through the same names via `useRoomTitles`, off the same shared memories cache (no extra request), and opens the memory.
- **One treatment for the rows.** A single `NoticeChip` behind the ping, the board notice and the knowledge push, replacing three near-identical buttons at two different max-widths, with a per-row `aria-label` (`Open thread …` / `Open task …` / `Open memory …`).
- **One size.** `.mono-inline` for an identifier set inline in copy (the episode tag, GAR, the plan file). Geist Mono runs optically larger than IBM Plex Sans at the same nominal size, which is what made the feed look ragged — `e4f1a2` and `GAR 0.79` visibly outsized the sans on either side of them. Relative (`0.92em`), so it holds at every step of the scale.
- Dropped the ping row's trailing `· click to open`. Every chip in these rows is clickable and only that one said so.

## Testing

- `pnpm test` — **766 passed**, 59 files.
- `pnpm lint` (`tsc --noEmit && eslint .`) — 0 errors (2 pre-existing warnings in `use-collapsible-rail.ts`, untouched).
- New regression coverage, aimed at the gap that let this ship:
  - `memory-preview.test.ts` — `memoryTitle` over the `{text: …}` shape, a bare string, a structured `title`, a heading marker, the `content_text` fallback, and the key as last resort.
  - `event-stream.thread.test.tsx` — a ping whose bound row is prose *in the shape the store hands it over* renders the first line and **not** the `work/…` key; a knowledge push renders the memory's name and not its key.
  - Two existing knowledge assertions updated to the new render; one of them now pins the honest case — a key no row in the room carries still reads as the key.
- Visually checked in `dev:mock` before/after via shotkit (`shot app /room/atlas-migration --mock`): titles still resolve under the new value shape on both the channel and the Board pane, and the mono runs now sit at the same optical size as the copy around them.

Backend untouched, so the Python gates in this template don't apply.

## Related Issues

Closes #889

## Notes for the reviewer

- **The one thing worth arguing with** is the mock change. It is the largest blast radius here (every frontend read of a prose memory in mock mode now gets `{text: …}`), and I could have fixed only `memoryTitle` and left the fixtures alone. I did not, because the divergence *is* the bug's cause: without it, `dev:mock` and most of the suite go on asserting against a shape the hub never sends, and the next reader that understands only the string ships the same way. It surfaced no other break — the full suite is green — which is itself the evidence the rest of the tree was already reading `value` defensively.
- **Not done:** the CLI's `_ping_line` (`commands/room.py:535`) prints the thread's short id, not a title. That is not this bug — a terminal tail has no rows loaded to resolve against, and six characters is the terse read a tail wants. Left alone deliberately rather than overlooked.
- `.mono-inline` is applied only to the mono runs inside these notice rows. The `→ recipient` chip in a chat header keeps `font-mono`: it has its own background and reads as a chip, not as copy.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XUaSrgbGig4s2BJH4DA3sS)_

### DIFF
--- a/mycelium-frontend/src/app/globals.css
+++ b/mycelium-frontend/src/app/globals.css
@@ -226,6 +226,16 @@ body {
   letter-spacing: 0.18em;
   text-transform: uppercase;
 }
+/* An identifier set inline in sans copy — an episode's short id, a metric, a
+ * file name. Geist Mono runs optically larger than IBM Plex Sans at the same
+ * nominal size, so a bare `font-mono` inside a line of copy reads as a size
+ * change rather than a change of register; stepping it down is what makes the
+ * line one size. Relative, so it holds at every step of the scale. */
+.mono-inline {
+  font-family: var(--font-mono);
+  font-size: 0.92em;
+}
+
 .square-dot {
   display: inline-block;
   width: 8px;
--- a/mycelium-frontend/src/components/episode-tag.tsx
+++ b/mycelium-frontend/src/components/episode-tag.tsx
@@ -23,7 +23,7 @@ interface Props {
  * clickable-reference look across chat, whichever kind of record it points at.
  */
 export function EpisodeTag({ urn, shortId, onOpen }: Props) {
-  const base = "rounded font-mono transition-colors";
+  const base = "rounded mono-inline transition-colors";
   if (!onOpen) {
     return (
       <Tooltip content={urn}>
--- a/mycelium-frontend/src/components/event-stream.render.test.tsx
+++ b/mycelium-frontend/src/components/event-stream.render.test.tsx
@@ -178,7 +178,9 @@ describe("<EventStream /> live message rendering", () => {
       });
     });
 
-    expect(await screen.findByText("plan updated → plan/tasks.md")).toBeInTheDocument();
+    // No row in this room carries that key, so the chip reads as the key —
+    // the honest read, and the only case in which a slug reaches the channel.
+    expect(await screen.findByText("plan/tasks.md")).toBeInTheDocument();
     expect(screen.getByText("Knowledge")).toBeInTheDocument();
     expect(screen.getByText("by @aligner", { exact: false })).toBeInTheDocument();
     expect(warn).not.toHaveBeenCalled();
--- a/mycelium-frontend/src/components/event-stream.thread.test.tsx
+++ b/mycelium-frontend/src/components/event-stream.thread.test.tsx
@@ -168,6 +168,70 @@ describe("<EventStream /> and the threads inside the room", () => {
     expect(await screen.findByText("flip reads behind a flag")).toBeInTheDocument();
   });
 
+  it("names a prose row by its first line, in the shape the store hands it over", async () => {
+    // The store keeps prose in the markdown body and rebuilds `value` from it,
+    // so a task row arrives as `{text: …}` rather than as a bare string. Reading
+    // only the string is what put a `work/…` slug in the channel where the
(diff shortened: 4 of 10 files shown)


juliarvalenti:
Closing: the rendering half here was superseded by #892 (useRoomRowNames/RowNaming rework), so this can't cleanly merge. But the real #889 bug it targets is still live on main — memoryTitle returns the key for prose memories ({text:…} shape) because it doesn't use memoryValueText. Reapplying just that fix cleanly on current main in a fresh PR.

PR #922, opened 2026-08-27
Add a synthetic backing track to the promo video

juliarvalenti:
## Summary

The promo is a code-defined video with no sound. This gives it a score written the same way: an ambient piece generated from source under `mycelium-promo/audio/`, rendered by `npm run audio` to a committed MP3 the composition loads as an `<audio>` clip on track 11.

Generating it rather than licensing a loop is what makes it **cueable**. The score's cue times are read straight out of `index.html`'s own GSAP calls, so the music lands on the frames the animation lands on — a terminal line, a service coming up healthy, an agent registering, a NEGMAS round advancing — and a beat that moves can be followed. That is not something you can do to a stock loop, and it is why the score lives in this repo under the same review as the composition it scores.

### The piece

One continuous organism rather than a cue per scene: a drone bed and a filtered-noise substrate run unbroken for all 107 seconds and everything else grows out of them. Two tables drive it — `SECTIONS` gives each scene a chord, a brightness and a pitch pool; `CUES` places motifs on individual frames.

**The argument is carried in the third of the chord:**

| | harmony | what is on screen |
| --- | --- | --- |
| 0–13.8s | D and A, no third | the hero card; a bell, and a long way down |
| 13.8–52.2s | the third arrives as **F** — D minor | install, room, agents |
| 61.6–81.4s | the chord drops its third entirely; `@rowan`'s figure lands on **F**, `@avery`'s on **F#** | the five NEGMAS rounds |
| **81.4s** | rowan takes F# and a sustained voice glides the semitone under it | the offer table locks green |
| 81.4–100.4s | D–F#–A–E, major | the plan compiles, the room is distilled |
| 100.4–107s | back to bare D and A, an octave wider | the outro |

The disagreement and its resolution are the same note. The generative layer carries the other half of the name: a pluck that fires may **fork** — a child a fifth or an octave away, which may fork again — so the texture spreads outward from each strike rather than repeating a pattern. Forks snap to the section's own pitch set, so a branch is always in the mode it lands in.

## Changes

- `mycelium-promo/audio/dsp.mjs` — oscillators, filters, envelopes (including a vactrol-ish lowpass-gate response), a modulated 8-line FDN reverb, WAV encoding
- `mycelium-promo/audio/master.mjs` — glue compressor, look-ahead limiter, and a BS.1770-4 loudness meter
- `mycelium-promo/audio/score.mjs` — the section table, the cue table, and the voices
- `mycelium-promo/audio/build.mjs` — render → WAV → MP3, printing what it measured
- `mycelium-promo/audio/README.md` — the piece, and how to verify a change to it
- `mycelium-promo/assets/mycelium-score.mp3` — the rendered asset (2.0 MB, 160 kbps)
- `mycelium-promo/index.html` — the `<audio>` clip
- `npm run audio` script, `ffmpeg-static` as the only devDependency, the intermediate WAV gitignored
- `CLAUDE.md` / `AGENTS.md` in the promo and the note in the root `CLAUDE.md`: **re-render after moving a beat the score is cued to**

No runtime dependencies — the DSP is hand-rolled. `ffmpeg-static` is used only to encode the WAV, and the system `ffmpeg` is preferred when one is on PATH.

## Testing

Nothing here can be checked by reading it, so the mix was tuned against measurement rather than by ear.

- **`hyperframes render` run end to end** with the track muxed in. Extracting the audio back out of the MP4 and comparing it to the source WAV: within **0.5 dB at every cue**.
- **The harmonic argument verifies.** Goertzel probes on the render: at t=79s **F4 is the loudest partial and F#4 sits 32 dB under it**; at t=83s that inverts — F#4 at −6.8 dB, F4 down at −29 dB, and no F natural anywhere in the D major sections.
- **Loudness:** `-17.0 LUFS · LRA 12.3 LU · true peak -1.9 dBFS`, from the meter in `master.mjs` — it agrees with `ffmpeg -af ebur128` to a tenth of a decibel. The loudest second of the piece is 82s, so the climax is the climax.
- **Deterministic:** two consecutive renders produce a byte-identical MP3.
- `npx hyperframes lint` — **0 errors**, 13 warnings, all pre-existing and all about the GSAP animation.

Two bugs the measurements caught that reading would not have:

1. `pluckFM` broke out of its render loop on `env < 1e-5`, which is true on the envelope's own first sample — **every FM bell in the piece was rendering silence**.
2. The fork intervals included a minor third, quietly reintroducing F natural into the D major sections right after the consensus had spent twenty seconds removing it.

**Not verified here:** `hyperframes validate` fails with a headless-Chrome navigation timeout in this environment — it fails identically on `main` without this change, so it is the sandbox, not the diff.

**One measured quirk, documented rather than patched:** the render pipeline's audio stage puts the track 1024 samples (21.3 ms, 0.64 of a frame at 30 fps) later than the source. It is not the MP3 — ffmpeg decodes the asset sample-exact. It is inside broadcast sync tolerance and inaudible on a bed with no transient sharper than a woodblock, so it is written down in `audio/README.md` rather than papered over with an offset that would rot the first time the pinned `hyperframes` version moves.

## Related Issues

None.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CGQoXqwLVMhzBChP56dza7)_


juliarvalenti:
superceded

PR #934, opened 2026-08-31
docs(tome): reground .tome pages in task-first architecture

juliarvalenti:
## Summary
- \`.tome/wiki.md\` and \`.tome/pages/{overview,architecture}.md\` still described the pre-task-first system (plan/tasks.md, "two engine kinds", no board/A2A/identity-ladder mentions), even though #916 had already patched the encryption over-claim inside them.
- Rewrote all three against current \`main\`: board/task/thread as the primary coordination model, negotiation as an optional in-thread activity (not the container), the A2A bridge (bidirectional, out-of-band HTTPS), the two-rung identity ladder now that SPIRE is removed, and the corrected "node forwards ciphertext only, hub holds plaintext" framing throughout instead of "end-to-end encrypted".
- \`team.md\` is unaffected (people, not architecture) and untouched.

These feed \`repos/mycelium/{overview,architecture}.md\` in Tome (mirror: true, tracked by source_sha) — merging is what propagates the content.

## Testing
Docs-only change, no code touched.


juliarvalenti:
The content reframe is a solid improvement and accurate on the substance (task-first board/task/thread, `plan/tasks.md` retired, A2A bidirectional/out-of-band, two-rung identity ladder + SPIRE gone, and especially the node-forwards-ciphertext / hub-holds-plaintext / not-E2E correction). Verified against `main`: `tasks.py`, `mediator.py`, `pi_session.py`, `a2a_bridge.py`, CLI `board.py` all exist, `plan_compiler.py` is indeed gone, and the tag really is `v3.0.4`.

One blocker before merge: three file pointers in `wiki.md`'s **"Where the actual logic lives"** section don't resolve, and that section's whole job is to be ground truth ("ground every claim against one of these"):

1. **`fastapi-backend/app/services/board/`** — does not exist. There's no `board/` dir or board/projection service under `app/services/`. Suggest pointing at `services/tasks.py` + `routes/tasks.py` instead (the board is projected from tasks, not its own module).
2. **`fastapi-backend/app/services/a2a_server.py`** — wrong dir; it's at **`app/routes/a2a_server.py`**.
3. **`pi_brain.py`** (in "pi_brain.py / pi_session.py") — does not exist. Only `pi_session.py`.

All three are ~30-second fixes. Everything else looks good to merge once those paths are corrected.

Minor, non-blocking: the "resident, never cold-spawned" bullet still references the removed `claude -p` spawner, and a few `PR #` breadcrumbs remain — but that's historical-narration cleanup separate from this PR's scope, and this file is a deliberate "what's easy to misdescribe" guardrail doc, so I wouldn't gate on it.

juliarvalenti:
Fixed all three: `services/tasks.py` + `routes/tasks.py` (board is a projection over tasks, not its own module), `routes/a2a_server.py` (was pointing at services/), and dropped the nonexistent `pi_brain.py` reference (only `pi_session.py` exists). Verified against main. Left the cold-spawn/PR-breadcrumb cleanup out of scope per your note.

Issue #932, opened 2026-08-31
A lease-only participant joining (or expiring) mid-negotiation doesn't abort it


sfph:
## Context

While fixing #924 (freezing a negotiation's roster on `members(room)`'s SLIM∪lease union instead of the SLIM-only `managed.members`), a follow-up review surfaced an asymmetry `_enforce_membership_change` still has, on top of that fix.

## The gap

`_enforce_membership_change` — the function that aborts an active negotiation when its frozen roster changes — is only ever called from two places:
- `_register_member` (a real SLIM join, via `invite()`)
- `remove` (an explicit leave)

Now that `open_episode`/`_enforce_membership_change` freeze/compare on `members(room)`'s union (SLIM presence *and* server-held HTTP leases, per #924), a genuine membership change on the *lease* side of that union has no matching enforcement hook:

1. **A brand-new lease-only participant appears mid-negotiation.** `refresh_lease` (`room_channels.py`) already detects the not-present→present edge (that's what gates `announce_join`), but it never calls `_enforce_membership_change`. The frozen negotiation keeps running as if the new participant never joined — asymmetric with the SLIM-join case, which correctly aborts (see `test_a_real_slim_join_still_aborts_a_frozen_negotiation` added alongside the #924-adjacent fix).
2. **A frozen lease-only participant's lease silently expires mid-negotiation** (its process died, network partition) with no other join/leave happening afterward. `_live_leases()` correctly stops counting it the next time anything calls `members(room)`, but nothing proactively re-checks the frozen negotiation against that drop — there's no periodic sweep, only the two invite/remove-triggered checks above.

Both are the same class of problem: enforcement only watches the SLIM half of the roster it's now supposed to be comparing as a whole.

## Why not fixed inline

This needs its own design pass rather than a quick hook:
- `refresh_lease` is a plain sync method called on every single `await` poll (~every 0.4s per active participant); wiring an async `_enforce_membership_change` call off it needs a fire-and-forget task (mirroring the existing `invite_in_background` pattern) fired only on the not-present→present edge, not every poll.
- Lease expiry has no event to hook at all today — it's discovered lazily, only when something happens to call `members(room)`/`_live_leases()`. Detecting it proactively would need an actual periodic sweep (or checking on every mediator round-timeout tick, which already treats a silent participant as reject/hold as a separate safety net — worth deciding whether that's actually sufficient in practice before adding a second mechanism).

## Suggested fix (not yet implemented)

[...]

PR #945, opened 2026-09-01
fix(await): deliver thread ticks to a bare await, and fix the reverse double-delivery gap

sfph:
## Summary

Two related fixes to `GET /rooms/{room}/await`, found while chasing why every stub-driven negotiation in mycelium-e2e-test's `nightly_suite` timed out or got silently rejected.

- **`participate.py`**: an unscoped (bare) `await` call skipped any record whose episode wasn't the room's own live channel — added in a prior fix (`b9d20314`) for a real cursor-hang bug, but with no test requiring it to also skip *delivery*. Since every aligner-mediated negotiation (task-bound via `board coordinate`, or bare via `engine invoke`) runs inside its own freshly-minted thread episode — never the live channel — this made every negotiation tick permanently undeliverable to any HTTP long-poll participant (the exact audience `await`'s own docstring describes: "a Claude Code session, a subagent"). Fixed to deliver thread records too, while keeping the independent-cursor behavior that avoided the original hang.
- A follow-up review caught the reverse case: a `--task`-scoped read (which never advances the room-wide cursor, by design) followed by a bare read could still hand out the same record twice. Fixed by having the bare path check the thread's own cursor before delivering, with no double-delivery either direction — regression tests for both orders included.
- **`room_channels.py`**: `_enforce_membership_change`'s abort-envelope `recipients` still used the SLIM-only `managed.members` (the same class of bug #924 fixed for the freeze/compare calls, but that PR didn't touch this line) — a lease-only participant frozen into a negotiation wouldn't be listed as a recipient if it aborted. Also dropped a redundant second `members()` recompute in the same function, and a redundant cursor-persist that turned out to be structurally unnecessary (`EpisodeCursors.position()` already defaults an un-forked thread cursor to the room-wide one).

## Verification

- Full backend suite: `uv run pytest tests/` — 1043 passed, 6 skipped, 0 failed.
- Live repro (side-by-side, same room, same instant): `GET /await?handle=X&episode=<negotiation-episode>` delivered the tick immediately; `GET /await?handle=X` (bare) returned `{"message": null}` before this fix, delivers the same tick after.
- Full `mycelium-e2e-test` `nightly_suite` (8 stub-coordination + hub-and-spoke tests, previously failing on silent timeouts) — 8/8 pass locally and on the actual GitHub Actions runner, both before and after merging in `main`/#924.
- Two rounds of independent code review (including a second-opinion agent) on this diff; findings applied are reflected in the commits above.

## Related

- #924 fixed the same class of bug for `open_episode`/`_enforce_membership_change`'s freeze/compare calls — this PR converged cleanly with it via merge and only carries the delta it didn't cover (the abort-envelope recipients line, plus regression tests for both).
- #932 tracks a related, broader gap this PR does **not** fix: `_enforce_membership_change` still isn't triggered by a lease-only join mid-negotiation, a lease expiring, or a genuine SLIM disconnect (`_drop_member`) — only an explicit `remove()` call reaches it today.


juliarvalenti:
This LGTM and is mergable.  TY!!

